PR #81, opened 2019-07-23
Corrected Julia section title

evelynmitchell:


jeapostrophe:
We actually do want to reference about Hacket. I just think that there was a typo in the first version.

evelynmitchell:
Yes, this is a typo correction. The Julia section was mis-titled.

jeapostrophe:
Ah, okay. Thanks!

PR #88, opened 2019-07-26
RFC: C-expressions

jeapostrophe:
This RFC attempts to define some concrete syntax for Racket2 in response to issue #3 

It is draft and I expect there are some big things that need to be updated, as well as medium things (like better explanations) and small things (like typos).

Please check it out and comment!

The easiest way to read the RFC is at [this link](https://github.com/jeapostrophe/racket2-rfcs/blob/cexpr/text/0001-cexpr.md).

### DIFF
--- a/text/0001-cexpr.md
+++ b/text/0001-cexpr.md
@@ -69,12 +69,12 @@ fun timed_thunk(thunk) {
      (stx-stmt answer))) */
      
 mac timed {
- case { s ... } => timed_thunk(() => { s ... });
- case e:expr => { timed e; } }
+ case { _ e; } => timed_thunk(() => e);
+ case _(e:expr) => { timed e; } }
 /* (stx-stmt mac timed
     (stx-stmts
-     (stx-stmt case (stx-op => (stx-stmts s ...) (stx-app PAREN timed_thunk (stx-op => null (stx-stmts s ...)))))
-     (stx-stmt case (stx-op => (stx-op : e expr) (stx-stmts (stx-stmt timed e)))))) */
+     (stx-stmt case (stx-op => (stx-stmts (stx-stmt _ e)) (stx-app PAREN timed_thunk (stx-op => null e))))
+     (stx-stmt case (stx-op => (stx-app PAREN e (stx-op : e expr)) (stx-stmts (stx-stmt timed e)))))) */
 
 fun g(x, y, (sign : s = +1)) {
  x + (s * y); }
--- a/text/0001-cexpr.md
+++ b/text/0001-cexpr.md
@@ -47,7 +47,7 @@ var x = 3;
 // (stx-stmt var (stx-op = x 3))
 
 fun f(y, z) { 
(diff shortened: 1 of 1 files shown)


gus-massa:
In [line 199](https://github.com/racket/racket2-rfcs/pull/88/commits/b6236f014e30ca28aee409d09b51c549400c8f02#diff-9a4cba2e630f52209b7db98b0ba06ff4L191) you are using ' instead of \`. 

Also it needs something like "and so on".

jeapostrophe:
Thanks @gus-massa

jackfirth:
Overall impression: I like this approach, but I've been assuming there would be an RFC discussing what we _want_ out of an alternative syntax before any proposals for one. @jeapostrophe, were you intending to write something like that as well?

jeapostrophe:
I don't intend to write an RFC about wants. I see some value in that, but I don't know how write it. I think these conversations about wants on the list and issues are valuably fluid. I think that everyone's reactions to some concrete proposal will expose more understanding of wants than just thinking about wants themselves. I'd really like to see someone else come up with an alternative or explain some want that this doesn't fulfill and we can look for ways to improve both and reach a new plateau.

mflatt:
The function and macro examples don't have a `;` after the body. Was there meant to be one, or have I misunderstood the grammar/intent?

jeapostrophe:
@mflatt That was a typo, I just pushed a fixed.

vityou:
This is slightly related, but one thing I haven't really understood is why the macro name has to be at the front of an expression.  For example, why couldn't you do something like tell the expander to look for the pattern `(a and b)` and replace it with `(and a b)`.  Could this restriction be lifted in something like c expressions?  Sorry if it's a stupid question.

jeapostrophe:
@vityou You need to have a consistent rule to know that the macro is `and` rather than `a` in your example.

vityou:
@jeapostrophe wouldn't the expander be able to figure that out similarly to how it knows if an something is a function rather than a macro?  Why wouldn't it work so that once the expander encounters a macro, like `and` that it gives the entire expression `(a and b)` to `and`?

jeapostrophe:
@vityou You're saying that when the expander sees a `(` it gives control to the first macro that it sees, wherever it is? So `(+ 1 a)` is a macro invocation if `a` happens to be bound to a macro in that block. I think that is possible but I don't like it :)

vityou:
@jeapostrophe I thought it may be helpful if we're moving to a more infix friendly syntax, however I probably haven't figured out all the drawbacks.  What problems do you see with it?

jeapostrophe:
Something I think might be beautiful would be a rule like that for situations where you have sequence of tokens but NOT when you have a infix operator (`1 + 1`) or a function call (`f(x, y)` --- doesn't depend on whether `x` or `y` is bound to a macro.)

jeapostrophe:
I'm opening this to make it more visible for history pre RacketCon

PR #95, opened 2019-07-29
Syntax considerations file

bzyeung:


AlexKnauth:
These considerations should include macros / language-extensions and language-oriented programming. When a user wants to add a new form to the language, does the syntax allow it to feel natural next to the existing forms in the language?

jackfirth:
In the "familiarity" section, I think we should emphasize that the syntax should feel familiar to people who have been exposed to the math notation of (US) high-school algebra specifically (functions, comma-separated arguments, infix math, grouping parens). That makes it clearer why some syntax features are in that list and other aren't. The notation of high-school algebra is very ubiquitous.

jeapostrophe:
@AlexKnauth @jackfirth Can you propose concrete things to add? Do you think we should add a file like this as-is and then start pushing new content into it?

AlexKnauth:
Concretely, in the `Considerations List` itemization, there should be a bullet point saying `macros / language-extensions feel natural next to existing language forms`

bzyeung:
@jackfirth That's a great idea.  I was shooting for a checklist item that was a bit less specific, with the notion that familiarity is a consideration that goes beyond just high school algebra, but I agree it would be helpful to identify those items with HS algebra.  Maybe adding another layer of nesting would accomplish both well enough?

@jeapostrophe I do wonder myself whether we should start it as this nebulous blob or whether we want to at least put a little bit of structure/guidance in there.

I tried to be kind of minimal with exposition, but if it would be helpful to have more of a proposed usage to react against, I've pushed a version that has a bit more of that language, along with the proposed list changes noted so far.  What do people think?

jackfirth:
@bzyeung I think that's a great idea! Another layer of nesting seems perfect.

@jeapostrophe Pushing a file like this as-is and adding new content via PRs sounds like a good workflow to me. We should probably have some way to easily tell which files are for "draft" documents that are active works in progress and which files are for finished products. Maybe by changing the file name to `syntax-considerations-draft.md`?

bzyeung:
I agree that high-school algebra notation is a trade-off.  Without diving too deeply into this, even given that we decide we specifically want to be familiar to those coming from HS algebra, it's not clear that comma-delimited arguments is a requirement.  A case could be made that people familiar with that notation wouldn't really have trouble with removing the commas, as long as the function name preceded a parenthesized list of arguments.  But that's a new set of trade-offs.

That line of thinking did partially motivate my choice of that particular nesting.  "Familiarity" feels like a consideration that we should have in mind as we weigh options.  I agree that "resembling HS algebra" feels like a constraint that was proposed along the way before we were thinking on that level.  I do think there's value to leaving a "trail" for people who might not be thinking on that level to make it easier for them to see the value/utility of this exercise.  Maybe the way to go is something like the top bullet being "Familiar to (not an exhaustive list):" with a bunch of other bullets, like HS algebra, s-expressions, C, natural languages, etc.  Maybe also additional Unresolved questions and/or explanatory language.  Thoughts?

jeapostrophe:
@jackfirth Everything is a draft :)

@bzyeung I personally think "familiarity" is fine to say. The idea of this list is to have proposals for syntax judged along these lines. So, I propose C-exprs, I'll talk about how they are familiar to C and kind of familiar to algebra; while if someone proposes something else, they'll say it is more familiar to algebra and familiar to Python. Etc

PR #104, opened 2019-08-12
 RFC: #lang inclusion processes

tgbugs:
This is a meta RFC about how new #langs will be considered for inclusion in the core implementation of Racket. I had to fudge the sections a bit because this isn't a technical RFC.

This RFC is still quite incomplete with regards to the dimensions and criteria for evaluating languages, and is entirely silent on the process that the community might use to determine them.
It is also missing a section about what it means for a language to have a community that relies on a core implementation as well as a discussion on what 'different' really means in the context of a language.

Despite the very rough state I wanted to get it out for preliminary comment and feedback from the community.

### DIFF
--- a/0002-lang-process.md
+++ b/0002-lang-process.md
@@ -343,39 +343,9 @@ doccument.
 
 19. Extensibility
 
-20. Extensibility of syntax vs extensibility of semantics
-
-    I'm worried that the failings of lisp (if there are any) will be
-    multiplies a billion times if you can allow inline definition of
-    completely new syntactic forms. The ability to create new syntax on
-    the fly which someone has never seen before and where they can't
-    infer the basic "oh, if I go look for (define-syntax (name stx) …)
-    then I can figure out what this is doing" then it becomes
-    monstrously difficult to use language to communicate, because the
-    basic conventions by which we communicate are themselves contextual.
-    Imagine if there was a lisp that allowed either head or tail
-    position or some intermediate position to be where the form that
-    triggers a macro is placed. The kind of calvinball chaos that would
-    ensue is not a pretty sight, and the way out would end up being
-    something like having operator precedence for every new form that
-    was not in head position. Don't get me wrong, it would be totally
-    cool from a technical perspective if I could magically define the C
-    ternary operator and was able to use it in arbitrary places in a
-    Racket program, but the technology that would let me do that seems
-    like something that could be used to destroy the world and wreak
-    havoc on any sensible persons mind because it is literally a tool
-    for painting bikesheds more fearsome than ever seen before – at
-    least with lisp you know that the first thing in the list is the
-    name of the thing you need to look up to figure out what it does –
-    how could you even look up foo ? bar : baz ?!?!? "c questionmark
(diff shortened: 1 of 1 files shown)


0002-lang-process.md
```diff
@@ -0,0 +1,493 @@
...
+2.  Regularity of visual form
+
+    This is a tough one – the primary issues is that in Racket you can
+    use \[\] and {} as a way to visually group blocks to make them stand
+    out and easier for people who read code visually (rather than
+    verbally / mathematically) to literally see what is going on.
+
+3.  Confusability (possibly fits under robustness)
+
+    How easy is it to change a single visually similar charachter and
```
AlexKnauth: typo

tgbugs: Thanks! Now I know not to ever trust default `flyspell-mode` behavior.

jeapostrophe:
I think this RFC mixes very different ideas. The first is the criteria for a language being part of the main distribution. The second is the criteria for submitting RFCs. The third is some criteria for thinking about syntax. I think these should all be totally separate.

On the first point (`#lang` inclusion): A major component of my vision for the package system was to winnow away the "main distribution" as much as possible so that there was a tiny tiny core and people installed all of the packages relevant to their work afterwards. We have the ability to do that now (Minimal Racket is very different than the old Textual Racket), but we haven't really pushed it hard. I have always thought and still think that the distribution is too big. The main reason we can't pursue this minimality is that educators must have a single thing to point their students to and we have a large variety of different kinds of classes that use Racket (intro courses, PL survey courses, PL theory courses, interpreter courses, compiler courses, etc.) So in summary, I feel like we should think about the main distribution purely as an educational thing and encourage all other users to think about Racket as a kernel + relevant packages. In other words, I don't think the idea of "including a new `#lang`" should be a topic of discussion. As a slightly broader matter, when should we talk about a `#lang` on the home page or in "advertising"? I think we do it when it is an attractive thing about Racket. Great examples of this, IMHO, would be `brag`, `hackett`, and `rosette`: these tools are great reasons to use Racket. We don't currently have them on the home page because the home page is very "the Racket management team" rather than "the Racket community" and we don't want to take credit for others' work.

On the second point (criteria for RFCing): You are essentially proposing that people cannot participate in the RFC process, except as commentators, if they cannot produce an actual complete implementation. I think that is the opposite direction from what we want; we want to lower the barriers to participate and discuss ideas at an early stage, rather than after a huge amount of investment in implementation has been made and people are emotionally attached to their sunk costs.

On the third point (syntactic analysis): These are great points.

mflatt:
@tgbugs I generally agree with @jeapostrophe. The lang-inclusion part of your PR seems to me in a different direction and/or maybe too early in the overall process, and I'm not sure we need a meta-RFC about how to use RFCs. Other parts of your PR seems to belong in "syntax-considerations.md", although maybe it's a mixture of things that belong in "syntax-considerations.md" and further commentary on that content.

I suggest splitting your PR into two: one to update "syntax-considerations.md", and (assuming that you want to keep it) this one that's just about language inclusion, perhaps with less of the meta-RFC detail.

PR #109, opened 2019-08-17
RFC: Subtle-expressions

jeapostrophe:


rocketnia:
This is great! I like the indent expressions and the text expressions.

jeapostrophe:
I just pushed an experimental parser for a language that is ALMOST subtle-expressions. It is LALR(1) and uses a patched version of `parser-tools` to get "error messages by example" (https://github.com/jeapostrophe/parser-tools/commit/3fb7834b578f2a888d8bb7a54adb23a8e913354d) although I don't actually have any examples yet.

It does a few things differently that felt nicer when I was playing with the code and examples. The big things that it doesn't do is (1) a post-processing fix-up pass to deal with precedence and (2) dealing with the `@` notation. I think that `@` requires an LL parser so that the lexer can be influenced by being in the `@` state. Maybe it's for the best because an LL parser would make it easier to do a top-down precedence algorithm.

Please take a look!

jeapostrophe:
@dstorrs 

The multiple options are in other pull requests, other peoples' ideas, and the [related work file](https://github.com/racket/racket2-rfcs/blob/master/prior-art.md).

Process 2 and Disadvantage 3: In this proposal, whitespace has to match in adjacent lines, it's not that it sometimes matters and sometimes doesn't. As for tabs, when I write whitespace, I always mean a space and nothing else. Tabs are for monsters and I eradicated them from my brain 25 years ago.

Process 3: There's a proposed table on line 208.

Process 4 and Disadvantage 3: All syntaxes are powered up by useful tools. S-exprs are a huge pain without highlighting of matching parens, regions, auto-insert of both parens, and missing `forward-sexp`/`backward-sexp`. Beginners are destroyed by trying to go without these. The proposed editor affordances are really simple to implement and not necessary.

Disadvantage 1. Clearly many programmers can survive with a small precedence table to memorize, given how many languages have them. I think at the very least, having the table be identical to normal math would be reasonable and very useful. I don't think parens around infix expressions are just like S-exprs. They feel very different to people.

--

I am actually working on a new proposal that I find much more beautiful than this one that I think may make you even more annoyed, so stay tuned. :)

dstorrs:
On Wed, Aug 28, 2019, 8:35 PM Jay McCarthy <notifications@github.com> wrote:

> @dstorrs <https://github.com/dstorrs>
>
> The multiple options are in other pull requests, other peoples' ideas, and
> the related work file
> <https://github.com/racket/racket2-rfcs/blob/master/prior-art.md>.
>
> Process 2 and Disadvantage 3: In this proposal, whitespace has to match in
> adjacent lines, it's not that it sometimes matters and sometimes doesn't.
>
There is an example in there where it explicitly states that it doesn't
matter, because there are two lines with different literal indentation but
the same relative indentation and that difference is lost in the parse tree.




As for tabs, when I write whitespace, I always mean a space and nothing
> else. Tabs are for monsters and I eradicated them from my brain 25 years
> ago.
>
That's a relief. Significant whitespace is which is a hassle when it's just
spaces and newlines, adding take is a disaster. Might be with mentioning,
though. You may consider "whitespace" and "space" to be eq?, but ASCII
disagrees.

> Process 3: There's a proposed table on line 208.
>
Yes, which includes the line "and all other unmentioned operators" and then
suggests that the last should be defined based on a rule applied to the
Unicode set.

> Process 4 and Disadvantage 3: All syntaxes are powered up by useful tools.
> S-exprs are a huge pain without highlighting of matching parens, regions,
> auto-insert of both parens, and missing forward-sexp/backward-sexp.
> Beginners are destroyed by trying to go without these. The proposed editor
> affordances are really simple to implement and not necessary.
>
> Disadvantage 1. Clearly many programmers can survive with a small
> precedence table to memorize, given how many languages have them. I think
> at the very least, having the table be identical to normal math would be
> reasonable and very useful. I don't think parens around infix expressions
> are just like S-exprs. They feel very different to people.
>
You are trivializing my point. Of course people can survive with operator
precedence, but it is a surge of bugs and it is not necessary when using
S-expressions. You are proposing a change that you claim will have
advantages; I am telling you that your RFC does not make clear what those
advantages are and it *does* make clear what some of the disadvantages of
your proposal are. You're welcome to ignore me -- I'm some random guy on
the internet and I recognize and am okay with the fact that I don't get a
vote when the core team makes decisions. Still, the point of an RFC
(Request For *Comments*) is to sell your proposal, and you're not doing
that.

--
>
> I am actually working on a new proposal that I find much more beautiful
> than this one that I think may make you even more annoyed, so stay tuned. :)
>
Ah, brilliant. Should be interesting ; you're a scary smart guy, so if you
think it's beautiful then it's definitely worth reading.

gus-massa:
> Racket2 could use the following hierarchy order: `::`, [...], `×` (syn with `*`), `÷` (syn with `/`), ...

Does it mean that `(2 * 3 / 4 * 5)` is `((2 * 3) / (4 * 5))` or `(((2 * 3) / 4) * 5)`?

I never remember the order of `and` and `or`. Where is `xor`?

jackfirth:
> Does it mean that `(2 * 3 / 4 * 5)` is `((2 * 3) / (4 * 5))` or `(((2 * 3) / 4) * 5)`?

This is an interesting question because of how math notation usually deals with this, which is one of a few ways:

- You're actually using polynomials so multiplication is implicit via concatenation, e.g. 2x / 5y
- You're writing division with a large horizontal bar which implicitly suggests the top and bottom are each grouped in parenthses
- You're using the ÷ sign instead of the / sign, which I think more clearly suggests that you're intending multiplication and division to have the same associativity and precedence behavior as `+` and `-` do

Using `/` in a programming-language seems like a way for code to _suggest_ the presence of a horizontal bar, as in 2. But using `÷` doesn't suggest this at all. So I bet if you ask someone how `2 * 3 / 4 * 5` and `2 * 3 ÷ 4 * 5` should parse, you'll often get different answers for those two forms.

dstorrs:
On Thu, Aug 29, 2019, 12:12 AM Jack Firth <notifications@github.com> wrote:

> Does it mean that (2 * 3 / 4 * 5) is ((2 * 3) / (4 * 5)) or (((2 * 3) /
> 4) * 5)?
>
> This is an interesting question because of how math notation usually deals
> with this, which is one of a few ways:
>
>    - You're actually using polynomials so multiplication is implicit via
>    concatenation, e.g. 2x / 5y
>    - You're writing division with a large horizontal bar which implicitly
>    suggests the top and bottom are each grouped in parenthses
>    - You're using the ÷ sign instead of the / sign, which I think more
>    clearly suggests that you're intending multiplication and division to have
>    the same associativity and precedence behavior as + and - do
>
> Using / in a programming-language seems like a way for code to *suggest*
> the presence of a horizontal bar, as in 2. But using ÷ doesn't suggest
> this at all. So I bet if you ask someone how 2 * 3 / 4 * 5 and 2 * 3 ÷ 4
> * 5 should parse, you'll often get different answers for those two forms.
>
If only there were some way to structure code such that rules of precedence
weren't necessary....

jackfirth:
To clarify, here's what I think:
- Mixing addition, subtraction, and multiplication is unambiguous
- Division is the devil's operator, and its presence means you should start putting parens in places instead of relying on operator precedence

jeapostrophe:
This proposal is complete trash. #114 is the new hotness.

PR #117, opened 2019-09-04
Sapling notation

mflatt:
Line-sensitive, indentation-insensitive, partially grouping (i.e., probably needs enforestation to arrive at mature tree).

### DIFF
--- a/text/0005-sapling.md
+++ b/text/0005-sapling.md
@@ -1,6 +1,6 @@
 - Feature Name: Sapling notation
 - Start Date: 2019-09-04
-- RFC PR: [racket/racket2-rfcs#115](https://github.com/racket/racket2-rfcs/pull/1145)
+- RFC PR: [racket/racket2-rfcs#117](https://github.com/racket/racket2-rfcs/pull/117)
 
 # Summary
 [summary]: #summary
--- a/text/0005-sapling.md
+++ b/text/0005-sapling.md
@@ -274,7 +274,7 @@ See [demo.sap](demo.sap), [interp.sap](interp.sap), and
 
 Sapling notation is not a good choice where precise and complete
 grouping is needed, both because its grouping is coarse-grained and
-the grpuing rules generate lots of extra `#%grp` layers.
+the grouping rules generate lots of extra `#%grp` layers.
 
 Saplings do not resolve the question of how infix expressions parse.
 There is no precedence at the sapling level, for example, other than
@@ -301,7 +301,7 @@ range of code to reindent if editing has made it unreadable.
 
 There is a lot of room to change the definition of _arrow_, _equal_,
 etc., to adjust the indentation rules, and possibly to tweak the
-grouping rules to avoid unecessary `#%grp`s.
+grouping rules to avoid unnecessary `#%grp`s.
 
 # Prior art
 [prior-art]: #prior-art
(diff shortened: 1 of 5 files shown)


sorawee:
Typos:

- grpuing (in Drawbacks)
- unecessary (in Rationale and alternatives)
- some parsing can be deferred a later arser has many precents in the (in Prior art)

samth:
There's one `salping`. 

A few questions I had:

Is the `\` in the `match` examples needed? Could they be on one line, or omit the `\`?

Is using `&` for `in` in `let` necessary? Could `let` use `in`, or would that require `in` to be part of the reader?

Why include the `\|` forms in the result instead of dropping them?

Is there a reason to parse juxtaposition as a 2-place "call" relation instead of an n-place "juxtaposition" relation (this is how the Fortress parser worked, for example)?

mflatt:
@samth:

I don't see `\` in the `match` examples. Can you point me to the specific place? (You're not thinking of Lexprs?)

Yes, `let` needs the `&` to continue to the `let` group. If you leave it out, then the body is a new group instead.

The parsed form keeps `|` because the intent is for the next layer of the parser to see them. It may need to distinguish `|` from `&` (as in `let`) or complain if the wrong number of hings are between `|`s. Ditto for `,`s. The intent is drop things only sparingly, but `;` and `\` are dropped.

I would read `f(x)(y)` as calling the function that results from `f(x)`, giving it the argument `y`. Is that the case that you had in mind?

samth:
1. Sorry, I got confused between thinking of my question, reading the Lexpr RFC, and writing down my question. Is the newline after  `=>` in the `match` examples requires?

2. In Fortress, we parsed `f(x)(y)` as (effectively) `(#%juxt f x y)` and `f x y` as `(#%loose-juxt f x y)`. That seems like it would be in keeping with the rest of the design here, since enforestation could turn one or both of those into applications.

mflatt:
(Not actually merged. How I didn't make things worse with a force-push to unwind that.)

mflatt:
@samth 

No, a newline after `=>` is not required, and in general adding a newline after `=>` will at most change the way it indens. Some `match` examples don't have the newline. See also "demo.sap" for more.

I see what you means about the Fortress parsing. I don't have a strong opinion right now, and probably we'd have to try out both if saplings ends up going anywhere.

samth:
Unfortunately I don't think there's a way to "unmark" it as merged, so if you want it to show up as an open PR you'd have to open a new one. :(

sorawee:
The parentheses in the LHS of `=>` in

```
define fib =
  lambda (n) =>
    cond | (n = 0) => 0
         | (n = 1) => 1
         | else => fib(n-1) + fib(n-2)
```

are not mandatory, correct? Otherwise, it will look very weird that the LHS requires parens but RHS does not.

mflatt:
@sowaree Correct.

At the sapling-notation level, there's no change to indentation or grouping outside the `(n = 0)` if you change it to `n= 0`.

And at the language-implemented-with-saplings layer, the `cond` that I have in mind would not require parentheses. Enforestation while parsing the first expression in a `cond` clause would turn `n = 0` into `(= n 0)`.

gus-massa:
Is it necessary to use the `=` after a `define`? The nice thing of Python is that to open each block you use `:` so it's not necessary to memorize which construction uses which symbol.


I think think it's better to allow hyphens and other symbols inside identifiers, like `string-append`, `quotient/remainder`. This is important to help with migration and interoperability, and to avoid an split in the community. Most people put spaces around operators anyway. (But the case `1+2` is tricky, because it's to similar to `1+2i`. Number are weird. )

mflatt:
Using `:` would behave the same as `=` at the sapling level. I'm uncertain about the exact set of operators that should behave like `:` and `=`, and I don't so far have strong opinions about where `:` or `=` (or other alternatives) should be used.

Meanwhile, the initial proposal's set of arrow identifiers like `=>` is almost certainly too large. Maybe, for example, it should be just operators that start `=` and end `>`.

So far, I really want `x-y` to mean `x` minus `y`. I'm not convinced most people reliably put spaces around operators, even if we only look at numerical operators. (FWIW, I tried a web search "example X program" for several infix-language Xs, and I find a mix of spaces and no-spaces.) But there are certainly different choices of syntax for identifiers that would work with the rest of the notation design.

mflatt:
Continued as #119

PR #122, opened 2019-10-01
Shrubbery notation

mflatt:
An indentation-sensitive notation that remixes elements of `#lang something`, Lexprs, and saplings.

[Rendered](https://github.com/mflatt/racket2-rfcs/blob/shrubbery/shrubbery/0000-shrubbery.md)


samth:
I think the link to the RFC discussion is wrong.

Can you add something comparing it to sapling notation?

mflatt:
Fixed and added. The short answer in comparison to sapling notation is that shrubbery notation is indentation-sensitive.

jeapostrophe:
See `|` used for `if` was enlightening to me! Beautiful. I think if I had thought of that, I wouldn't feel like the line continuing rule after the indented block was as necessary.

In `print_sexp`, the `for` has indentation but no suffix... I assume this is "visible", might be nice to add a comment in the first example about the indentation grouping rule 

I think this a lot and am willing to move forward with this. For next steps:
- I think that the `@` notation is hopefully obvious
- I think the `#{}` idea is fine, although I think it would fine to tweak C names a little bit to accommodate `?` and maybe a few others
- I am anxious about delaying precedence/etc to another level, but am willing to say that shruberies 
can do whatever, but the next language may have some particular system.

gus-massa:
I think forbiding `.` in identifiers is a good idea in spite the incompatibilities that it may cause, but nobody is using it. Probably forbiding `:` is also good, it may cause more problems like in `exn:fail?`, but help to avoid some bugs using `syntax-parse`. I'm not sure about  `!`, I like it but it's a nice shorthand of `not`. And I really like `?`.

About delaying the precedence to another level, I think it is necessary to support user-defined operator (my examples in a recent thread was the `@` in python and the `.*` in matlab).

mflatt:
@jeapostrophe I think it was confusing to have the "a colon could go here" comment in the middle of the first example, since that's not the first place where a colon is optional. I changed the text before the example to explain `:` as indentation up-front and ignore the detail that extra `:`s are allowed.

@gus-massa I think the C and the Lisp ends of the identifier spectrum make sense, but I'm skeptical of in-between points. With Lisp-style identifiers, you have to put space around any operator (including something like `:`), so that's why I lean toward C-style identifiers in an infix-friendly notation. I also lead toward C-style because I am tired of having to explain that the `-` in `x-y` or the `!` in `set!` is just part of the name; it means the syntax has already confused my audience, and I'm explaining to try to undo that confusion. I can see how allowing `?` in identifiers just might work ok with C-style syntax, because `?` is not commonly used in operators (and we don't need the C-style `?`...`:` since we're not going to make the mistake of `if` being a statement and not an expression). Still, `?` won't look like part of a name to many programmers, and I would prefer not to spend time in the years ahead explaining that `?` is just a part of an identifier's name.

samth:
Also, Rust and JavaScript now both use `?` as an operator (for related things, but not related to the C `?`).

sorawee:
> @jeapostrophe I think it was confusing to have the "a colon could go here" comment in the middle of the first example, since that's not the first place where a colon is optional. I changed the text before the example to explain `:` as indentation up-front and ignore the detail that extra `:`s are allowed.
> 
> @gus-massa I think the C and the Lisp ends of the identifier spectrum make sense, but I'm skeptical of in-between points. With Lisp-style identifiers, you have to put space around any operator (including something like `:`), so that's why I lean toward C-style identifiers in an infix-friendly notation. I also lead toward C-style because I am tired of having to explain that the `-` in `x-y` or the `!` in `set!` is just part of the name; it means the syntax has already confused my audience, and I'm explaining to try to undo that confusion. I can see how allowing `?` in identifiers just might work ok with C-style syntax, because `?` is not commonly used in operators (and we don't need the C-style `?`...`:` since we're not going to make the mistake of `if` being a statement and not an expression). Still, `?` won't look like part of a name to many programmers, and I would prefer not to spend time in the years ahead explaining that `?` is just a part of an identifier's name.

I think Ruby allows `!` and `?`, and lots of programmers really like Ruby. Is it really the case that people are surprised by this issue?

(This is not even to talk about how people who never program before wouldn't feel `!` and `?` particularly weird... the whole programming is weird to them!)

mflatt:
A clarification on Ruby: Ruby allows `?` and `!` specifically at the end of method names (not in identifiers in general, and not in other places within the identifier). It also has `!` as an operator. It seems that the expression `x!(1)` or even `x!1` is a call to the `x!` method with the argument `1`, while `x ! 1` is a call to `x` with the argument `false` (since `!1` is false).

I am not aware of any formal study that would confirm whether `-` or `?` as part of an identifier really surprises people. Maybe there has been one. Currently, I can only report my perception from my experience (working with programmers at various levels) that it does surprise them, and it requires specific explanation in a way that `a` or `x` as part of an identifier does not.

jeapostrophe:
As much as I don't like this, I agree 100% that people are constantly asking if `?`, `-`, and `!` are part of the syntax of Racket identifiers, same with `^` and `%` if they ever get to those.

gus-massa:
About the main part of this proposal:

I like the idea of meaningful indentation. It reduces the amount of parenthesis a lot, like in my  examples in samplings. I don't use paredit, so I rely in the parenthesis to detect the mismatch. I sometimes use `{}` for `define` and `[]` for long `begin` blocks to make the detection of mismatch easier. I will need to be more carefully when editing, and it will need more editor support to move the blocks to the left/right as automatically as possible. I like this anyway.

I'd like to add the magic `&` for `for`/`let` in addition of the magic `|` for `if`/`match` and the magic `:` for everything. But I think this addition is quite orthogonal and I don't expect to cause too much problems, so perhaps we can delay the discussion for later.

I'm not convinced of the optional  `:`. I prefer a strict rue of a `:` before every new indented block that is not a `|` block, and never a `:` before a `|` block. It makes the code core consistent.

I like the two spaces before the `|` blocks. Indenting and using the `|` is a little redundant, but it looks nicer.

I'm still not sure why the `{}` blocks must be different of the `()` blocks. But if they are equivalent I'm worried about the confusion of an operator that can be unary and binary an is at the begining of the line. I still can't make a good example, but something like this example. (Note the lack of the `,`, `display` is perhaps a bad election, and I need a unitary operator that has side effects and also can be used as a binary operator so I used `--` that is the nearest candidate.) 

    display(x
            -- y) 

    display{x
            -- y}

jackfirth:
> I also lead toward C-style because I am tired of having to explain that the - in x-y or the ! in set! is just part of the name; it means the syntax has already confused my audience, and I'm explaining to try to undo that confusion.

Now that I think about it, I've probably also had to explain that dozens of times. It would be nice to just... not have to do that. I'm coming around on the idea of C-style identifiers for Rhombus.

mflatt:
@gus-massa On the optional `:`, I'm sympathetic to your point. I started with the `#lang something` optional `:`, but maybe the choice there is related to how `:` enables indentation within `()`. Otherwise, I thought it might be annoying to have to delete a `:` when editing code in some cases. But it does seem like the reader can simply disallow `:` before `|`, `{`, or indentation, and providing a good error message seems straightforward, so maybe that's better. 

I also share your concern about the special rule for operators to continue a line within `()` and `[]` and the way that `{}` is different. Within `()` and `[]`, a compensating factor that we expect `,` to separate groups, as you say. The remaining risk of confusion seems a worthwhile trade-off to avoid `\` for breaking arithmetic across lines, but I'm not 100% certain.

I'm not initially enthusiastic about `&`, because getting rid of it seemed like one of the two big improvements over sapling notation (where the other is getting rid of the need for blank lines). But I encourage you to try/share examples and maybe try adjusting the parser.

michaelballantyne:
+1 for required `:` before indented blocks.

I'm not sure I like the idea of preserving commas on the parse. I'd prefer that either:

1. separating eg. function arguments with newlines be exactly the same as using commas
2. comma-separated groups be a special kind of group, where commas are always required, but dropped in the parse.

When would a group that mixes use or absence of commas be desirable?

````
f(a, b
  c, d,
  e)
```

mflatt:
@michaelballantyne Requiring commas in `()` and `[]` seems likely a good idea. I avoided that requirement originally, because I wanted to make the notation flexible — deferring when possible to a language built with the notation. But enforcing a use of commas in `()` and `[]` at the shrubbery-notation level seems consistent with the way the division of responsibility has evolved.

mflatt:
I've updated the description and parser:

* A `,` is now required between each group in `()` or `[]`. Extra `,`s are not allowed. I like how this removes the need to represent `,` in parsed forms.

* A `|` is now implicitly indented by half a column. Although @gus-massa liked how indentation was required for nested `|`, it feels a little less confusing to allow a `|` to line up with the "keyword" that starts a conditional form, because the shape of `|` makes it look a little indented already. More significantly, a half-column alignment side-steps the question of mixing groups that start with `|` and groups that don't. And since they can't be mixed, a block that contains `|` groups can be simplified to an `'alts` variant of `'block` in parsed form.

* A `:` (or `|`) is now required before indentation. I'm ambivalent. It often looks noisy and feels genuinely redundant to me. But a redundant `:` is useful as a kind of belt-and-suspenders notation to help detect earlier when indentation goes wrong — we like indentation, but don't really trust it? — and I think that's probably where the suggestion comes from. (@gus-massa  and @michaelballantyne can correct me if I'm wrong.)

Extra `:`s are still allowed. Requiring a `:` before indentation is redundant and slightly pendantic, but workable. Having the parser complain when you have an extra `:` (because you just inserted a newline after a `:`, for example) seems unnecessarily pedantic. A style guide and code-formatting tool should normalize `:`s` as well as indentation, of course.

jeapostrophe:
I really like all of these changes, because I like the greater specificity. Thank you!

yfzhe:
Are characters used as "operators" like "+", "-" not allowed to be shown in an identifier? I see some identifiers like `make_adder` and `color_posn` in the demo, and `in-list` in (#lang) racket is transformed to `in_list`.

sorawee:
> Extra `,`s are not allowed.

@mflatt does that mean that trailing comma (`[1, 2, 3,]`) is not allowed? I find that trailing commas are really useful in two cases: 1) writing prettifier is easier, because we can just `print(arg + ', ')`, and more importantly 2) in line-based diff, using trailing commas allows adding arguments in new lines to touch exactly only those new lines, so the diff is not noisy. So I hope that trailing commas will be allowed.

> Are characters used as "operators" like "+", "-" not allowed to be shown in an identifier? I see some identifiers like `make_adder` and `color_posn` in the demo, and `in-list` in (#lang) racket is transformed to `in_list`.

@yfzhe  Yeah, see https://github.com/racket/rhombus-brainstorming/pull/122#issuecomment-537505453

MarcKaufmann:
> So I hope that trailing commas will be allowed.

I strongly second this from the way I write Python code, where I always add trailing commas. Whenever I don't, I end up forgetting to add a comma to the formerly last row when adding a new one, which is annoying.

I have a question on how to evaluate different options. I find it difficult to even think through what goals of usability and clarity of the syntax we want to achieve - and for which subgroup of people? What *are* some actual studies showing what type of syntax find more natural along whichever metric? And do you have an ideal study where you'd say "I would change my mind on doing X rather than Y if you could show me that 1st-year CS students/long-time C coders after 1 day/1 month exposure with X rather than Y are happy/faster at writing syntactically correct code/can understand the meaning of written code?" Even if we don't find such studies or can't plausibly run them, it would help clarify what trade-offs we are making.

As an example, I find it annoying to write _ instead of - in identifiers. It's one of my main syntax-related frustrations when not programming in Racket -- I find it very non-ergonomic and slows my typing down where I consciously notice it (hitting <shift> and the - button). I can see why this may confuse lots of people however and require explaining it to them. However, if after explaining it once to them they understand it and are never confused, then so what? Alternatively, if they read code such as pair-vector1-vector2 as a pair - vector1 - vector2 even after some exposure to the syntax, and if we care more about readability, I would not just be happy to drop -, but in fact agree that having - in identifiers is bad. Similarly, I'd put some serious money on people being bad at parsing prefix mathematics even after quite some exposure to it -- but if you actually showed me that after 1 month of using prefix daily, people read and write it as easily as infix, I'd be surprised, but I'd be happy to grant that I am the odd one out, and I'd agree that prefix serves the goal better.

jackfirth:
> I can see why this may confuse lots of people however **and require explaining it to them**

Emphasis mine. In my experience, the costs of _teaching_ a notation dominate the costs of reading it, and both dominate the cost of writing it.

MarcKaufmann:
That makes sense. It also highlights why my opinion as a user, who primarily *writes* his own code, but rarely has to reread much of it, and even less teach it, is probably tilted in the wrong direction. Which is why it's useful to spell that out - and I can be happy without '-' in identifiers.

I would add that leading to more correct and maintainable code beats initial teaching costs, and for me personally infix maths is much less likely to lead to mistakes than prefix, despite me getting both.

mflatt:
The idea of allowing a trailing comma surprises me. I had not noticed the evolution on this point in other languages.

I could see allowing a trailing comma and having the standard format include a trailing comma when a closing `)` or `]` is written on its own line (as required in Go).

jackfirth:
Trailing comma support works best when there's an autoformatter involved, so it gets you better diffs without users having to actually think about where to put trailing commas and where not to.

Related, many languages now ship with a standard autoformatter. @mflatt, do you think defining some autoformatting rules as part of shrubbery notation would be a useful thing to do at this stage?

gus-massa:
I'm not sure about the last comma, but there are some example in Python anyway:

    bool(0) # ==> False (it's the number 0, that is falsy)
    bool(0,) # ==> False (again the number 0, the last comma is ignored)
    bool((0)) # ==> False (again the number 0, unnecessary parenthesis are ignored)
    bool((0,)) # ==> True (it is the tuple with one element that is zero)

Another unrelated idea for the comas is from VB.  Two comas separated by space mark a optional argument of a function that is replaced by the default value.

    MsgBox("Message", , "Title")  ' The second argument is missing so use the default.
                                  ' It selects the select the icon and buttons to display.

gus-massa:
About the spaces of indentation before `|`, I think that 1 or 2 is better than 0. I'm not sure if 1 is better than 2 or not. Using 2 look like too much indentation indeed. 

The problem is the editor while you are writing the program. Imagine you type

    if x>0 [enter] [tab] | ...

How many spaces should `[tab]` move? Is it necessary to press `[tab]`? With 2 spaces, [tab] always adds 2 spaces.

I think the objective of adding some syntax is to fix the shortcomings of the s-expressions. It's quite abstract, so I prefer a more concrete objective of removing the special rules for indentation and square bracket in DrRacket. It looks like this notation is good enough to eliminate the rules after the program is complete, but DrRacket will need some rules for the indentation while typing the program.

I think only the [L-expressions](https://github.com/racket/rhombus-brainstorming/pull/114) need no rules for a complete program and no rules while typing a program, but it needs to use too many continuation characters at the end of the lines to ensure that.

mflatt:
It turns out that `parse.rkt` already allows a trailing comma, because it wasn't specifically guarding against that. I'll adjust the text to say that it's allowed. @gus-massa Thanks for the VB note; I wouldn't go that way, but it's good to know about the precedent.

@jackfirth Yes, I think basic formatting rules belong at this phase. I tried to specify some already, characterizing certain patterns as "standard" style or indentation, and we can add more of that.

@gus-massa I very much agree on the goal of avoiding binding-specific indentation. As a starting point in this notation, I suggest 1 space for `|` indentation and 2 spaces for other indentation. For editor behavior, I had imagined that `[enter]` after `:` would indent, `[enter]` without `:` would not indent, and `|` would automatically (un)indent without requiring `[tab]`. Meanwhile, `[tab]` would cycle through plausible indentations (potentially unindenting). But there's probably more/better that others have worked out for Python-style syntax.

PaulMorris:
In this syntax proposal and the others, it would be interesting to see how syntax for optional type annotations would fit in.  It might be worth considering that sooner rather than later to avoid "painting oneself into a corner" when it comes to "Typed Rhombus".

sorawee:
That's a good point, but @mflatt has already included some examples. See https://github.com/racket/rhombus-brainstorming/pull/122/files#diff-ef7ebf39fed9f2714f31edee8c398140R169 for instance. I think it's still interesting to see examples of annotation on non formal argument position.

sorawee:
@mflatt Speaking of type annotation, is this a typo?

```
define exp(n: Integer, base: base = 2.718281828459045)
```

Probably it should be this, right?

```
define exp(n: Integer, base: Integer = 2.718281828459045)
```

samth:
@sorawee the base argument doesn't look like an integer to me ...

sorawee:
🤦‍♂maybe Real or Number then ...

mflatt:
When I wrote that, I was experimenting with having type-annotated and keyword arguments at the same time. Maybe it makes sense if types are syntantically distinct (e.g., uppercase), or maybe it doesn't make sense.

sorawee:
Oh, so that's intended to be a keyword argument? That is,

```
base: base = 2.718281828459045
```

is similar to 

```
#:base [base 2.718281828459045]
```

?

If so, the question then is that how do you provide a type annotation on top of this. `base: base = 2.718281828459045: Real` or `base: base: Real = 2.718281828459045` look really awful to me.

Incidentally, for languages that have keyword arguments, most of them only allow specifying identifier name, and the keyword will then be the same as the identifier name. Is there a real value of allowing them to differ in Racket? It looks like most uses of keyword argument will have keyword and identifier that are identical anyway:

```
~/racket $ ag '#:([^ ]*) \[\1' --stats
1818 matches
~/racket $ ag '#:([^ ]*) \[(?!\1)' --stats # this one has A LOT of false positives, so the actual number is lower.
592 matches
```

mfelleisen:
If these are type annotations, it suggests Rhombus wants a "micro optional type system". Is this intended?

samth:
@mfelleisen Typed Racket has basically the same optionality for type annotations as this syntax.

mfelleisen:
@samth Your reply to Oak suggests that one parameter of a function may come with a type spec and the other one may omit it. Am I reading this wrong?

samth:
Yes, you can do exactly that in Typed Racket as well.

mfelleisen:
@samth Got it. (1) In TYPED Racket, something like `(define (f {x : Integer} y) (+ x y))` is legal syntax but doesn't type-check. Fine. (2) The LanguageNameThatRhombus will use eventually will be derived from TYPED Racket not PLAIN Racket.

mflatt:
@mfelleisen Like S-expressions, shrubbery notation is intended to support multiple languages — some of them with types, some with optional types, and probably others that use a type-like position for other purposes.

@sorawee The advantage of the current syntax for keyword argument is that it's a little more consistently parallel to function calls, where there's a keyword and an argument expression (so, imitated by a keyword and argument name on the declaration side). But there are certainly trade-offs, and having comma-separated arguments may shift the trade-offs.

mfelleisen:
If shrubbery notation is like S-expressions, we don't need to specify type positions now. We just rely on syntactic extensions.

michaelballantyne:
Right. But we need to know that the reader notation we choose is flexible
enough that we can write type annotation-like syntax that will look
aesthetically pleasing.

mfelleisen:
Does this mean the parenthetical syntax is superior? We did not have to know in 1958 or 1995 that we'd add type annotations retroactively to Lisp/Racket. We just did. I would hope that Shrubbery allows such a post-hoc addition, too. You never know what you want to add later.

mflatt:
@mfelleisen I think you know that the goal here is to have an adaptable syntax. There are obviously tradeoffs for S-expressions versus other notations, and we don't need to rehash them in this thread. This thread is for a specific proposal, and it if you have thoughts on how this specific notation does or doesn't support extension, that feedback would be helpful here.

mfelleisen:
@mflatt I think you know that I raised a rhetorical question to tease out a point. My low-level conjecture is that the type examples snuck in via carelessness. My high-level point is that there's a rationale missing between the overall goal statement and this concrete proposal. So, here is what I think the unwritten rationale is: 

the Shrubbery notation is isomorphic (in a sense one could make precise) to S-expression syntax. 

It is all about writing 
```
#lang name-tbd 
shrubs 
```
vs 
```
#lang typed/name-tbd
typed shrubs 
```

If I am wrong, I'd like to see this intermediate rationale because contributing to this discussion is extremely difficult otherwise.

mflatt:
@mfelleisen I wonder if you're trying to think of `demo.shrb` as a program in one particular sapling-based language. That's not the intent. That file is meant to hold saplings that could represent program fragments in various languages. Maybe the fact that most of the demo saplings could go together in some program is giving the wrong impression.

The `interp.shrb` example is meant to be a program in a particular language—one that is similar to Plait. The reason to try out some fragments and programs as they might look in a typed language is that `:` has a particular roles in saplings and a traditional role with types; the examples are meant to probe how well those roles mesh.

mfelleisen:
... think[ing] of `demo.shrb` as a program in one particular sapling-based language. ...

This is precisely what I thought. But perhaps I am the only one who doesn't need this explanation.

bmitc:
I would encourage the team to take a look at F#'s syntax and implementation, especially regarding indentation and pattern matching. This could be potentially useful in two ways:

1. understand and learn from where the F# team struggled in their implementation of the indentation rules, which are rather complex in F# (this has been written about and I believe it's at least mentioned in @dsyme's _History of F#_ paper),
2. benefit from and hopefully improve upon F#'s clean syntax in its core language. F# is generally much cleaner than equivalent code in Python, Ocaml, SML, etc.

I'm making this comment mainly because I took a look at the linked examples, and I the use of `:` operator stood out to me as something that doesn't really look that great, especially for the result expressions of pattern matching.

Just as a point of comparison, I implemented some of the examples in F#.

The shrubbery examples:

```
define pi: 3.14

define identity(x): x

define fib1(n):
  match n
   | 0: 0
   | 1: 1
   | n: fib1(n-1) + fib1(n-2)

define fib2(n):
  match n | 0: 0
          | 1: 1
          | n: (fib2(n-1)
                + fib2(n-2))

define fib3(n):
  match n
   | 0:
       0
   | 1:
       1
   | n:
       fib3(n-1) + fib3(n-2)

define make_adder(n):
  lambda (m):
    printf("adding to ~a\n", m)
```

In F#:

```
let pi = 3.14

let identity x = x

let rec fib1 n =
    match n with
    | 0 -> 0
    | 1 -> 1
    | _ -> fib1(n-1) + fib1(n-2)

let rec fib2 n =
    match n with | 0 -> 0
                 | 1 -> 1
                 | _ -> fib2(n-1)
                        + fib2(n-2)

let rec fib3 n =
    match n with
    | 0 ->
      0
    | 1 ->
      1
    | _ ->
      fib3(n-1) + fib3(n-2)

let makeAdder n =
    fun m -> sprintf "adding to %A\n" m
```

I tried matching the indentation style where appropriate, and this is functional (i.e., working) F# code. I used the wildcard `_` operator, which is idiomatic F#, but F# would have supported me typing an `n` instead. A few comments. (These suggestions may be too specific to the examples given, but I thought I'd at least mention them. If there's a better place, let me know.)

- I personally dislike `_` as a name separator convention. Hyphens `-` are much more clean to read and convenient to type. I feel hyphens are better than underscores, camelCase, and PascalCase, with underscores being the worse (like in Python and Elixir). The hyphens in Scheme and Racket code have always been a plus to me over every other language.
- I personally dislike how `:` looks in the code. It's a "soft" visual character that causes things to bleed into each other or itself to just blend into its surroundings. I generally dislike it in Python and would encourage people to not look at it too much for its syntax. In general, I find F#'s syntax to be much cleaner than Python, even when writing nearly the same lines of code, including OOP (i.e., ignoring F#'s more advanced features that Python doesn't have).
- In F# there is the thing of using parentheses or not, so it's potentially good that a new language that forces them is better as F#'s slight flexibility in the matter causes inconsistencies in codebases. I've only used parentheses when required in the F# examples above, but the single argument functions could have used them, such as `let identity(x) = x`.
- Obviously syntax is personal, but I feel the F# code is cleaner primarily because of `=` and `->` replacing the usage of `:`. Also, `let` tends to be nice, but that's more of a preference. The `rec` I feel neutral on, which is required in F# for recursive functions. It takes some getting used to, but it's a nice reminder and note that the code defined is recursive.

cloudhan:
I'd like the mention nim lang here since it is not mentioned over all issues/PRs, see [create-a-simple-macro](https://nim-lang.org/blog/2018/06/07/create-a-simple-macro.html) to get some sense. Hopefully, it might give reader of this comment some inspiration of the syntax.

ShalokShalom:
You could use 'Tabs for indentation and spaces for alignment'

Tabs for indentation and spaces for alignment is the one way to check all the checkboxes in [your document](https://github.com/racket/rhombus-brainstorming/blob/master/resources/syntax-considerations.md#considerations-list)
[See also here]
(https://dmitryfrank.com/articles/indent_with_tabs_align_with_spaces)

Another reason is that alignment and indentation is something different, so I guess different symbols for different things make sense.

 It gives more of an impression, what is used when and why, especially for beginners. It is a check for the student.

sorawee:
A couple of questions:

- Is it possible to enforce that there must be no space between function position and argument list? I tried the parser and found that both `f(1)` and `f    (1)` give the same result.
- Is the `curried` function example meant to be called with `curried(1)(2)(3)`?

rooney:
>     * I personally dislike `_` as a name separator convention. Hyphens `-` are much more clean to read and convenient to type. I feel hyphens are better than underscores, camelCase, and PascalCase, with underscores being the worse (like in Python and Elixir). The hyphens in Scheme and Racket code have always been a plus to me over every other language.

@bmitc The problem is it makes e.g. `-x` ambiguous, and the options are both ugly:
* Not supporting unary `-` operator, or
* Not allowing names to start with `-`
What do you propose to resolve that?

AlexKnauth:
Forcing spaces between operators and arguments, consistently over both binary and unary operators

rooney:
That means the syntax also won't support things like `!foo`, `#bar`, or `@baz`

mflatt:
Some updates on this particular experiment:

 * The spec now pins down a syntax for things like numbers, identifiers, and operators. It includes `#{`...`}` as a limited S-expression escape for fancier numbers and identifiers.

 * The package https://github.com/mflatt/shrubbery-rhombus-0 implements `#lang shrubbery` with syntax coloring and indentation in DrRacket.

Using the package, if you run a `#lang shrubbery` program, it just prints the parsed representation. It's more interesting to hit Tab in DrRacket to cycle through possible indentations for a given line. (I recommend against trying Tab with multiple lines selected, at least until/unless a relevant DrRacket PR goes through to give a `#lang` more control.)

mflatt:
@sorawee 
 * Currently, the specified representation of a parsed shrubbery does not provide enough information that a language (built on the notation) could require a function expression to be adjacent to the opening parenthesis for a function call. It would make sense to add adjacency information in the parsed form, and it does seem worth considering as part of experiments with a language using shrubbery notation.
 * Yes, in the half-formed language suggested by the examples, I imaged that `curried` would be called as `curried(1)(2)(3)`.

mflatt:
@bmitc Do you have a pointer for more information on where the F# team struggled in their implementation of the indentation rules? The HOPL paper mentions that indentation-aware syntax was added. I looked through the referenced blog post, but it seems to just present the results. (Retrofitting indentation sensitivity to an existing language seems like a different game, but there may be related strategies.)

sorawee:
> If so, the question then is that how do you provide a type annotation on top of this. `base: base = 2.718281828459045: Real` or `base: base: Real = 2.718281828459045` look really awful to me.

The latest commit answers my above question. Thank you!

sorawee:
Can this be supported?

```
match 0:
| 0: 
  1
| n:
  n
```

Being forced to write:
```
match 0:
| 0: 
   1
| n:
   n
```

feels really weird to me.

Perhaps a special rule that if `:` comes right after `|`, count `|` as the starting column.

sorawee:
Also, slightly surprised that

```
(1 +
 2)
```

doesn't parse.

mflatt:
@sorawee

I'm not enthusiastic about trying to special-case indentation for a `|` plus `:` combination. Partly, it's because your match formatting looks wrong to me, but maybe I've been using the existing implementation too long already. :) But I'm also unclear on the suggestion. When you say "`:` comes right after `|`", what's the definition of "right after"?

For continuing a parenthesized group the next line without `\`, the current rule is that it can continue if an operator starts the new line. C code is more readable to me when a continuing operator is on the left of a new line, but I think the Python convention (not a rule) is to put it on the right of the earlier line. I can imagine changing the current rule or allowing both, and my opinion on this point is relatively weak.

rocketnia:
Hmm, is it happening because in this code,

```
| 0: 
  1
```

the `0:` and `1` lines are considered to have the same indentation level in the `|`-initiated block?

If it were caused by the behavior of `:`, I might have suggested that `:` consider the following line to be part of its block if it's merely more indented than the beginning of the `:`'s line. But that already appears to be the case in the examples:

```
hello:
  world
  universe
```

Personally, I think I'm fine with the current design except in one case.

Consider this `cond` from the examples:

```
define fib(n):
  cond
   | n == 0: 0
   | n == 1: 1
   | else: fib(n-1) + fib(n-2)
```

In a more sophisticated program, the conditions and branches might be more sophisticated, and we might put grouping parentheses around them both and put them on different lines:


```
   | (n == 1 && another_condition):
       (1 + 2 * 3)
```

In the s-expression world, indenting past an opening parenthesis like this usually coincides with that parenthesis being unmatched. That makes it easy to determine by eye how many `)))` to write at the end, by zig-zagging up the left edge. I think languages that break that convention start to rely on tooling to match parens.

I think this runs up against one of the points in the syntax considerations file:

> users (beginners in particular) should be able to read code in various environments, like GitHub and codeblock embeded in Markdown. Matching parens is not easy if it requires you to use rainbow parens, for instance.

On the other hand, it looks like one of the other considerations is

> dependent on advanced tooling/editor support

So perhaps being willing to depend on tooling like rainbow parens is a goal, rather than a non-goal. I'm not sure if I'm reading that right.

Anyhow, I didn't make this digression to say "how about s-expressions?" :-p If we could indent the branches like @sorawee was trying to do...

```
   | (n == 1 && another_condition):
     (1 + 2 * 3)
```

Now the branch body isn't indented further than the opening paren on the previous line. So, matching parens by eye might be a little easier with this indentation style.

sorawee:
> When you say ": comes right after |", what's the definition of "right after"?

I couldn't pinpoint the definition, but I have another thought.

My worry is mostly with the rightward drift issue. It's already a problem in Racket, and with the current Rhombus style, it could be worse. Compare:

```
(cond
xx[1
xxx2])
```

with:

```
cond:
x| 1:
xxxx2
```

What I just realized is that `block` has double functionalities. First, it allows users to group a bunch of things together. Second, it allows users to stop the group whenever they want to.

```
group_1:
  thing_1
  thing_2
thing_3 // // stop the previous block group
group_2: 
  thing_4
  thing_5
```

In the `cond` construct, I argue that `block` is not the appropriate grouping to use, because we don't need its second ability. That is, we don't need the ability to stop the group. We will never write:

```
cond:
 | c_expr:
    expr_1
    expr_2
   another_c_expr:
    expr_3
    expr_4
 | yet_another_c_expr:
    expr_5
    expr_6
```

The body of each branch is already grouped appropriately by the alt-block, making an additional block is redundant.

So perhaps what should really happen is to use an operator/keyword to separate conditional expression and the body in each branch:

```
cond:
 | c_expr =>
   expr_1
   expr_2
 | another_c_expr => expr_3
```

And ideally, we should encourage reducing indentation whenever possible, but my opinion on this is weak.

```
cond:
| c_expr =>
  expr_1
  expr_2
| another_c_expr => expr_3
```

mflatt:
@sorawee 

I agree that `cond` clauses don't inherently need a block to group the answer part of a clause. After trying it, though, I still think `:` works better. It could be just that I'm stuck on a notation, but here are some points in favor: 

* it maintains the use of blocks for starting a definition context;
* it avoids reserving another operator; and
* indentation can be a more helpful visual distinction than a tailing `=>` if the question expression gets large enough that a visual distinction matters.

@rocketnia

I kindof see what you mean, but I can't help thinking that the main problem in that example is unnecessary parentheses. Even in the S-expression world, I think some people use an always-indent-the-next-line style, but maybe that doesn't apply when the first component has parentheses.

benknoble:
My vote is for `=>`, not that it has much weight. Resembles SML's `case` arrow, also used for pattern-matching, and an existing Racket convention where some separate conditions from blocks with `;=>` on a line. The trailing arrow can move to a new line as a horizontal separator if the condition is long.

michaelballantyne:
Would it be possible to describe `{}` blocks, `()` / `[]` lists, and `|` alts as separate syntaxes? The writing of the proposal now suggests the three structures are broadly similar, but then creates exceptions that mean they really act quite differently.

Instead we could have roughly the following:

* Blocks are created by a pair of braces `{}`, a `:`, or implicitly in the body of an alternative. Within a block, newlines or semicolons separate groups.
* Lists are created by a pair of parentheses `()` or braces `[]`. Within a list, commas separate groups.
* An alternatives sequence is composed of several alternatives each beginning with a `|`.  The body of each alternative is a block.

Some related special cases:

> As a special rule, when a group is immediately within a () or [] group sequence, the group continues to the next line when that next line starts with an operator. 

Commas are already required to separate elements of a () or [] group sequence, so why consider line breaks alone as a potential separator? Only separating by commas allows

```
(my_favorite_number +
  your_favorite_number +
  their_favorite_numbers)
```

as well.

> As a special rule, if the block that would be created by : has one group that is itself a block, then : does not create a new block.

This seems like it could be confusing in its own way; I would be surprised that

```
a:
  {val x 5; x}
```

and

```
a:
  {val x 5; x}
  {x}
```

or

```
a: {val x 5; x}
```

and

```
a:  {val x 5; x}
     {x}
```

parse with different group depths. Having

```
hello: {world; universe}
```

create nested group structure is less surprising, to me. Allowing `hello: : : : : world` seems undesirable.

> A | is implicitly shifted half a column right (so, implicitly nested), and it is implicitly followed by a : that conceptually occupies same column as the |. When a | is not at the start of a group or its group is at the start of a block or after an opener, then the | is also implicitly preceded by a :. Note that the implicit : following a | is subject to the special rules for :, which is that it doesn't create a new block if one is already created (by indentation or {), and it can force an empty block.

> A | has its own special rule: if a | appears on the same line as a | and would start a block immediately within that |'s block, then | instead terminates the previous |'s block and continues its enclosing group sequence with a new | group that is not implicitly preceded by a :. The intent and consequence of this rule is that multiple |s can be used on a single line as an alternative to starting each | on its own line, making the following groups the same as the above groups:

I find these rules very difficult to follow.  What about making some of these error cases instead? It seems like the versions with `:` and extra `{}` are not preferred, so why not disallow them? One possibility to replace these rules and enforce unambiguous syntax would be to say that a group cannot start with an alternatives block. No explanation of a relationship to `:` would be needed, and the following would be illegal:

```
hello: | world
       | universe

hello { | { world }
        | { universe } }

hello: | world | universe

hello { | world | universe }

hello { | { world } | { universe } }
```

Alternatives only make sense to me when they follow something else that provides a context, so I can't think of a case where I would want to start a group with alternatives.

mflatt:
@michaelballantyne 

The idea of merging groups in `(` ... `)` or `[` ... `]` without a `,` does work, but I would be reluctant to go that way. A main goal is to have visual grouping reliably reflect form grouping. It seems bad to let the `fun` in

```
(map fun
 (x): x
 list)
```

consume the `(x): x` that looks like it's a group on its own. Also, in my limited experience writing programs with the `shrubbery-rhombus-0` candidate prototype experiment, I've appreciated the simple and clear "missing comma" errors when a comma got lost. The current rule of allowing a group to continue when the new line starts with an operator does not fully preserve the constraint that visual grouping reflects form grouping, but it sacrifices less, and it leaves at least some visual indication at the start of the new line to suggest that it may not be its own group.


The rules on blocks and redundant blocks came out of a goal to allow writing anything on one line that could be written on multiple lines. Consider this example:

```
x
 | a
 | b
y
```

It's easy enough to get rid of most of the line breaks without introducing `{`s:

```
x | a | b
y
```

But now we're stuck, because

```
x | a | b ; y
```

would mean

```
x
  | a
  | b
    y
```

We can get back to the original by using `{` and `}`:

```
x { | a | b } ; y
```

But this has one of the patterns that you've listed as illegal.

So, the one-line goal meant at least allowing extra `{` ... `}` around `|`s. Maybe that's the only place, though. Allowing extra blocks all over was an attempt at one kind of consistency (if they're allowed in one place, allow them everywhere), but it does seem to have bad consequences.

Another possibility is to require a `:` instead of having a `|` start a block for multiple `|`s. I see that you've written `:` after `cond`, and I've seen others do that. It looks redundant to me, especially in something like `if x == y: | 0 | 1`, but it also seems like a workable rule.

I'm inclined to try allowing redundant `{` ... `}` only around a block of `|`s (which would also mean that it's still ok to write `cond:`), but I'll wait for more opinions.

mflatt:
Making `:` required before `|` worked fine when I tried it in the Rhombus candidate prototype experiment. I've pushed that change here and in `mflatt/shrubbery-rhombus-0`, so others can try it out.

samth:
One relatively recent syntax that I like a lot is the F# (and now JavaScript and others) pipeline operator. I've tried to implement that in the `shrubbery-rhombus-0` prototype, and it works ok:

```
let format(f): fun(x): racket.format(f,x)  

operator (x |> f): f(x)

1 |> racket.add1 |> format("~a")
```



However, multi-line pipelines are less nice -- effectively, they have to be parenthesized as far as I can tell. For example, this is a syntax error.

```
def g2(x): x
           |> racket.add1
           |> fun (x): racket.format("~a",x)
```

And I would have to write it this way:

```
def g2(x): (x
            |> racket.add1
            |> fun (x): racket.format("~a",x))
```

I then tried to implement this as a macro, with the idea that I would consume the subsequent forms more eagerly:
```
expr.macro ?(??x |> ??f ??tail ...):
  values(?((??f)(??x)), tail)
```
But that didn't work. 

It's possible that this could be dealt with at the enforestation level
(and maybe that's where I should put this discussion) or maybe even
via my failed macro idea, but an alternative choice would be to relax
the rule about continuing groups on new lines with operators to apply
to any block.

I tried that, by commenting out the check for `state-paren-immed?` on
line 319 in shrubbery/parse.rkt, and it makes `g2` above work
correctly. That's far from a comprehensive test, though, and probably
we would want a more restrictive check.

mflatt:
You're right that a macro can't bridge the line-based separation here. Grouping in shrubbery notation is meant to constrain the reach of a macro in that way.

Within parentheses, an operator on a new line is allowed to continue a group, because groups must also be separated by `,` within parentheses. That is, there's no ambiguity for a line that starts with an operator about whether it should start a new group, because a missing `,` helps indicate that — assuming, at least, that the `,` wasn't omitted by mistake.

Concretely, commenting out the check for `state-paren-immed?` causes the `demo.rkt` file to not parse. The `¿` on line 237 is treated as a continuation of the preceding group. That leads to a “wrong indentation” error right now, and while it must be possible to change more so that it would be added to a preceding group, that's not the intent of the code. The same thing happens with the leading `?` on line 86. More generally, it would mean that a prefix operators don't work at the beginning of a group unless there's a `;` before.

It would be possible to sort operators into different categories, where `|>` would continue a group and `¿` or `?` would not. I don't know whether that would be a good idea, but it doesn't seem crazy to treat operators that start `|` specially.

It may be that line-based grouping fights too much against this kind of syntax and a different approach to surface syntax is better.

mflatt:
After trying it for a few weeks, requiring `:` before `|` still feels redundant to me. So, I'd like to try out disallowing `:` before `|`. This is not quote the opposite of the previous design change, because it keeps the simplifications otherwise; for example, it doesn't make the `:` optional.

Also, the default indentation of `|` now uses 0 spaces instead of 1 space.  In part, I was convinced by seeing 0 spaces as the convention for `match` in OCaml and F#.

michaelballantyne:
It seems like the `|>` issue could be resolved by indention:

```
x |> f
  |> g
```

would be one group because the second line is indented and begins with an operator;

```
x |> f
|> g
```

would be two (and an error at enforestation).

Indention is already significant in determining which block a line belongs to, so I don't think this would be too much of a departure from the spirit of the syntax.

samth:
@michaelballantyne that indeed seems nicer than either my change or making an exception for `|`.

mflatt:
The latest version, based on the discussion in #172 and the discussion here on `|>`, makes a few big changes:

 * `{}` is no longer used for groups. Instead, it's like `()` or `[]` and represented as a `braces` form.

 * `«»` can be used in combination with `:` and/or `|` to delimit blocks. The expectation and experience is that you practically never need to do that when writing shrubberies, though.

 * When an indented line starts with an operator and doesn't follow a `:`, then it continues the previous line's group. Each subsequent continuing line must have the same indentation as the first continuing line, also starting with an operator. The special continuing rule within `()` and `[]` has been removed.

michaelballantyne:
@mflatt What is the motivation for the design goal that it always be possible to express or print a shrubbery on a single line? Always using multiple lines for the complex nested block and alternative situations would avoid the need for the new Unicode delimiters.

mflatt:
@michaelballantyne So far, I ran into the single-line question when printing shrubbery syntax objects. There is surely a better approach to printing than my quick implementation, but being able to put things on a line — or, more generally, having a simple form that accomodates textual nesting — seems like something we probably take for granted in various settings. So, it's not an important constraint, and probably not important enough to spend ASCII characters on, but it seems worth spending a couple of non-ASCII characters to solve a small problem that shows up in different ways.

michaelballantyne:
Makes sense. I guess I'm more worried about a "huh, that's weird..." budget than a character budget. It seems unfortunate to have special syntax for an edge case.

mflatt:
Looking at this again, I see that the `«| ....»` case isn't necessary, so I've removed it to reduce the weirdness spend.

sorawee:
Is it worthwhile to have more than one "body" (like TeX) in the @-notation?

I've seen this feature request for a couple of times in the Pollen community. For example, one might want to write:

```
@braket{@phi}{@psi}
```

which is much cleaner compared to the current workaround:

```
@braket[@content{@phi} @content{@psi}]
```

mflatt:
@sorawee That seems possible. Since the body is delivered in an explicit list, it could work here more easily than with `#lang at-exp`.

zamora:
In the "Parsed represention" section of the description of shrubbery notation, it states:
> An element created by [] is represented by 'braces consed onto a group-sequence list.
> An element created by {} is represented by 'brackets consed onto a group-sequence list.

This seems backwards - I usually see [] referred to as brackets and {} as braces. Though a few comments up, it seems that both kinds may now be represented by 'brackets.

mflatt:
@zamora You're right — that description is backwards. The implementation is the other way around with `[]` => `'brackets`. I'll fix the description. Thanks!

gus-massa:
I'm not sure if this is a goo idea, but... 

One problem that I have in Python is when I comment the only instruction in some code that I use to debug or to have a verbose output, but I don't want it in the final version. (And I don't want to remove it, because I may need more debugging later.)

    #lang python 
    for i in range(10):
        print(i)
        if i == 7:
            #print(i, i*i)
            #continue # <-- why is this necessary

mflatt:
The latest version is an experiment changing `«` and `»` so that parsing the content inside is not sensitive to line breaks or indentation.

In some ways, this change brings us full circle to proposals that advocate a choice of equivalent notations, where one is indentation-sensitive and the other is not. A difference here, though, is that the indentation-insensitive notation isn't meant to be particularly convenient to type or pretty on its own, and so it can be closer to the indentation-sensitive syntax. That is, it's meant as a kind of "armor" mode to minimally adjust program text while ensuring that accidental line or indentation changes won't change the way the text parses (except for text in `@` notation, for now, but likely support there can improve).

One possible use of armoring is just before copying some program text to move it into a different context, where the text could be pasted and reindented in the target context. Unarmoring in the target context then ensures that it parses the same as before it was copied.

The `#lang rhombus` prototype binds Meta-A to armor a selected region of a program. Here's what the current "demo.rkt" looks like after armoring:

 https://gist.github.com/mflatt/b932084a4b2489abbe115022b2b81b9b

The next step would be to make Meta-A toggle armored text to unamored.

To support a sequence of groups that is not in a block, a `«` can now appear just after `;`. In that case, the groups between `«` and `»` are spliced into the enclosing context.

sorawee:
Correct me if I'm wrong, but mainly, there are two ways to style a colon.

1) If colon is at EOL, then the current indentation is increased by 2. For example:

```
yyyyyyy:
  // current indentation = 2
  xxxxx:
    // current indentation = 4
    foo
```

2) If colon is not at EOL, then there's something after colon. The current indentation is set the the column position of that "something". For example:

```
yyyyyy:
  // current indentation = 2
  xxxxx:     hello
             // current indentation matches hello's position
```

An opening paren-like (brace, bracket, paren) at EOL increases the current indentation by 2. For example:

```
yyyyyyy:
  // current indentation = 2
  hello(
    // current indentation = 4
```

```
yyyyyyy: hello(
           // current indentation matches hello's position + 2

```

## Question

In JS, people usually write something along this line:

```
yyyyyyy:
  // current indentation = 2
  val x: {
    // current indentation = 4
    a: 1,
    b: 2,
  }
```

Shrubbery notation can read this code just fine. But the style doesn't follow the "rules" I wrote above, and the current indenter will fight against this style.  

I guess the main question I have is, is this style endorsed? And if so, how should the "rules" by adjusted?

mflatt:
@sorawee That output might be implemented as "two more than whatever indentation starts the line with the opener". But that choice also leads to an example like this:

```
top:
  val x: form {
    inside
  }
         after
```

instead of

```
top:
  val x: form {
           inside
         }
         after
```

which seems like it might be a bad idea.

Special-casing indentation for an opener–closer pair at the end of a block might work, but indentation is currently determined only by looking before the line to indent. I'm not at all sure those are the only possibilities.

sorawee:
Consider:

```
// pre
fun foo(a, b): // qwe
  // www
  a // def
   + b // qqq
// post
```

Here's the raw information on each node:

```
#<syntax (top (group fun foo (parens (group a) (group b)) (block (group a (op +) b))))>: #f #f #f

#<syntax top>: () #f (" " "// qqq" "\n" "// post")

#<syntax (group fun foo (parens (group a) (group b)) (block (group a (op +) b)))>: #f #f #f

#<syntax group>: () ("\n" "// pre" "\n") #f

#<syntax:string:3:0 fun>: "fun" #f #f

#<syntax:string:3:4 foo>: "foo" (" ") #f

#<syntax:string:3:7 (parens (group a) (group b))>: #f #f #f

#<syntax:string:3:7 parens>: "(" #f (")")

#<syntax (group a)>: #f #f #f

#<syntax group>: () #f #f

#<syntax:string:3:8 a>: "a" #f #f

#<syntax (group b)>: #f #f #f

#<syntax group>: () ("," " ") #f

#<syntax:string:3:11 b>: "b" #f #f

#<syntax:string:3:13 (block (group a (op +) b))>: #f #f #f

#<syntax:string:3:13 block>: ":" #f #f

#<syntax (group a (op +) b)>: #f #f #f

#<syntax group>: () (" " "// qwe" "\n" "  " "// www" "\n" "  ") #f

#<syntax:string:5:2 a>: "a" #f #f

#<syntax:string:6:3 (op +)>: () (" " "// def" "\n" "   ") #f

#<syntax:string:6:3 op>: () #f #f

#<syntax:string:6:3 +>: "+" #f #f

#<syntax:string:6:5 b>: "b" (" ") #f
```

I will need to think more, but my first impression is that I'd really love for `// qqq` to associate with `b`, `// def` to associate with `a`.

sorawee:
Indenter nit:

The interaction of the indenter and automatic closer insertion is not great. Here are some examples, where `|` indicates the caret position.

```
//// Example 1 (there are stuff after opener)
// before
foo(arg,|) // enter

// after
foo(arg,
|)

//// Example 2 (opener is at [to-be] EOL)
// before
foo(|) // enter

// after
foo(
|)
```

Example 1 looks incorrect. There is no reason that `)` should be put there. Any other text would have been indented to align with `arg`, so I would expect the indenter to produce:

```
foo(arg,
    |)
```

Example 2 might appear to indent correctly, but in practice, programmers expect that after entering a newline, they will be able to continue coding right away. As it currently is, they would need to back up one line and enter a newline again to achieve:

```
foo(
  |
)
```

which is not ergonomic.

Most code editors will produce the above editor state right away. 

Note that the above observation ("programmers expect that after entering a newline, they will be able to continue coding right away") also validates:

```
foo(arg,
    |)
```

since they can add more code right away.

sorawee:
Thanks!

Speaking of `|`, when "Enable automatic parentheses, square brackets, and quotes" is enabled, and `|` is typed, it will create two `|`s. Is there a way to adjust this behavior based on `#lang`? The behavior makes sense for `#lang racket`, but does not make sense for `#lang shrubbery`/`#lang rhombus`.

rfindler:
@sorawee I've (slooooowly) been fixing that in DrRacket and I'll try to get something pushed soon that fixes that.

mflatt:
@sorawee I pushed the change to keep same-line whitespace and comments with the preceding token as a `'raw-suffix` property.

You mentioned in person that it would be useful to preserve the parsed structure of commented-out groups and alternatives. I'd rather not have that as the default, because the idea is that plain strings might be lightweight enough to preserve in syntax objects for compiled form (to be used later when reporting errors about macro expansions). So, there would need to be a special mode — and in that mode, would it be better to just leave the commented-out group as non-comments in the parse? The `#//` comment token could be in the commented-out term's prefix, and maybe there should also be an extra syntax property to indicate when something is commented out.

rocketnia:
Conceptually, I think quoting and commenting can interact. Within a deeply nested quotation, a comment can be intended to be preserved through several levels of quotation before the comment is supposed to kick in and erase it for the rest. This is a niche concern—virtually all mainstream string syntaxes don't even have a single way to write comments inside them, much less ones that interact carefully with nesting—but it's something that's been on my mind in my designs, and now seems like an apropos time to mention it for Rhombus.

For this to work in my design notes, I've treated comments as being a kind of escape sequence, and I've given escape sequences a prefix that allows them to be annotated with simple information relevant to determining the depth they apply at. (In the most straightforward cases this would be a number written in unary, but in my designs, I want to be able to specify labels to jump directly to certain nesting levels.) Unfortunately, the presence of these prefixes makes it hard to determine what kind of escape sequence is coming up by peeking for it, so for instance it's hard to write something that skips whitespace and comments but doesn't read into any other escape sequence. I've considered making up for this by giving them another prefix that acts as a forward declaration of what kind of escape sequence is coming up.

Frankly, it's messy, and the resulting comment notations are probably too verbose. But I thought it could be useful to share this experience report in case some of these concerns are relevant.

More to the point, if comments are part of the AST, then at some point someone might ask for comments that are *really* comments rather than being part of the AST. And then if that exists, someone might ask for a quoting operator that *really* quotes things, so that those comments are part of the AST too. I think the idea of different comments having different escaping depths can help explain how these competing concerns can interact and coexist.

mflatt:
@sorawee I made a further change that I think you'll approve of, but mentioning in case not: When  a block is followed by a comment that starts on the same column as the block's content, then the comment and intervening whitespace is kept with the block as a tail, instead of left to be a tail of an enclosing for or a prefix on the next form. (This rule makes a Scribble `rhombusblock` form work better.)

mflatt:
@rocketnia I'm not sure it's deeply related to what you have in mind, but something like that happens with `@` comments. When you group `@` content using `{` and `}`, then `@//` within the group is a comment that is not preserved with the rest of the content in `{` and `}` (although it's still stashed in properties). But if you  wrap that whole thing as `@` content using, say, `|<<{`  and `}>>|`, then the comment using `@//` will be preserved as text, while a comment using `@|<<//` would disappear.

sorawee:
@mflatt Thanks! I totally agree that this is better.

rocketnia:
@mflatt 

> But if you wrap that whole thing as `@` content using, say, `|<<{` and `}>>|`, then the comment using `@//` will be preserved as text, while a comment using `@|<<//` would disappear.

Oh yeah, I learned about that feature recently (in regular Racket-based Scribble rather than Rhombus), and I really like that it's available as a technique. I think it is related, not just in superficial terms but potentially in more intricate ways as well.

I think the essential difference is that Scribble's labels apply to lexer syntaxes rather than quotation levels. The difference has been pretty subtle for me as I've worked on Punctaffy, which deals mainly with higher-dimensional analogues of lexer syntax, but has an application as a building block for quasiquotation syntaxes. I've mixed up these concerns more than I'd like to admit. I think in the majority of practical cases, Scribble's syntax labels are sufficient for suppressing the syntactic features of arbitrary code, in the same way that a preprocessor is often a sufficient substitute for a function. And just like preprocessors and functions, these techniques each serve their own purposes.


### Statically scoped quotation labels

The mental model I have in mind is that each comment (whether labeled or not) should be statically associated with a quotation level, in a way that's as reliable as static lexical binding. If a comment is associated with a quotation level that's at a nonzero depth, that comment is suppressed; it represents its source text rather than doing anything comment-like.

Whether or not a comment is suppressed this way, it still begins and ends in the same places. We might consider those beginning and ending points to be determined *before* scope resolution figures out what level the comment's label refers to. The code of a suppressed comment might even be syntax-highlighted in a way that reflects its status as a quoted comment, like I started to describe in https://github.com/racket/drracket/issues/512#issuecomment-924508528.

Relevant excerpt:

> [...] each level could be tagged with the language that the editor is basing its editor functionality on at that level (usually Racket), and each pair of adjacent levels could be tagged with the kind of embedding that's going on between them (commenting-out or quotation).


### Statically scoped lexemes

Anyhow, if I understand right, the `|<<{` `|<<@` `}>>|` labeling in Scribble applies to lexer syntaxes. In this way, Scribble can suppress comments by changing what kind of syntax even counts as comment syntax in the first place. This prevents things from being comments so decisively that we don't even consult the comment syntax's logic to determine where these non-comments begin and end. If someone suppresses a comment this way, its source text can be given a surprising new interpretation.

In the following example, some code initially works because some erroneous (or perhaps malicious) notation is commented out. Once it's moved into another quotation that suppresses the comment, the notation contained in the comment gets interpreted with a different semantics that makes it actually do something:

```racket
#lang at-exp racket

; Here's an example:

(displayln
  @~a{
    hello
    @; Commented-out }>>|}) (displayln (string-upcase "code injection")) (displayln @~a{@~a|<<{ code.
    world})

; Here it is in a quoted string:

(displayln
  @~a{
    ; Here's an example:
    
    @~a|<<{
      (displayln
        @~a{
          hello
          @; Commented-out }>>|}) (displayln (string-upcase "code injection")) (displayln @~a{@~a|<<{ code.
          world})}>>|})
```

Output:

```
hello
world
; Here's an example:

(displayln
  @~a{
    hello
    @; Commented-out 
CODE INJECTION
 code.
world})
```


### Code injection?

Using "code injection" in that example might be catastrophizing. I mean, my concern is to be able to move code around without adjusting its escape sequences. If my concern here were security, just about every language under the sun would be vulnerable to something like this on account of the popularity of double-quoted strings (and single-quoted ones, etc.):

```
; Safely prints some code we don't want to run.
(displayln "(/ 1 0)")

; Does *not* safely print some code that would safely print some code we don't
; want to run.
(displayln "(displayln "(/ 1 0)")")
```

And if everyone's vulnerable to this, then someone must have thought hard about why double-quoted strings are actually okay... right? Or is that wrong? Well, personally, I don't know how to tell people it's a catastrophe even if it is, and what I do know is that I prefer the alternative quotation syntax approach I'm describing, which doesn't share the same usability gotchas.


### Shortcomings of my approach; and tying back to Scribble and Rhombus

I wouldn't say my statically scoped quotation levels are necessarily free of gotchas altogether. I think most of the gotchas are herded into a particular corner, though: Moving code around is still dangerous if it contains unmatched brackets or free variables which might interact differently with their new surroundings.

Of course, if nothing else, the code pasted into the string could simply contain an unmatched string-ending bracket sequence, which matches up in a potentially surprising way with the string boundary itself.

The need to avoid unmatched brackets is more pervasive than that, though. Since strings that track quotation depth need to recognize nested quotation forms to keep track of the quoting depth, they need to be able to match up brackets to figure out where those nested quotations begin and end. Unmatched or mismatched brackets can interfere with this, so they need to be meticulously escaped. And if these mismatched brackets are intended to affect the quoting level, then I suppose their effect on the quoting level and the extent of that effect has to be meticulously explained in another escape sequence (one I haven't ever prototyped yet).

I'm in a scenario where I have a whole mess of unmatched brackets to deal with, I'll probably prefer wrapping them in Scribble-style labeled brackets rather than escaping them all individually. That's where the "these techniques each serve their own purposes" synthesis comes in. :)

Another issue that's come up in my explorations is that when I have code that resembles English text, that becomes a *problem*, because it can't be distinguished as program syntax inside a string. But this concern is similar to Scribble, where `@` keeps a lot of things distinguishable from text. But I think even if Rhombus doesn't make itself pervasively distinctive the way Scribble does, there's still a viable approach: Explicit escape sequences can break up strings into regions of quoted code and regions of other text. Hmm, I suppose this might even mean my dreaded double-quoted strings are salvageable if they're extended with the right escape sequences. :)

ceving:
I am wondering if it possible to simplify the block syntax by replacing the pipe with a preceding colon. I think it might be possible to give the sequence `newline` `whitespace` `:` the meaning of the pipe.

~~~
if x > 0:
  : do-if-1-then-1
    do-if-1-then-2
  if x < 0:
    do-if-1-else-1-if-2-then-1
    : do-if-1-else-1-if-2-else-1
      do-if-1-else-1-if-2-else-2
~~~

mflatt:
#527

PR #202, opened 2021-12-16
Add Wraith syntax proposal

cwebber:
This proposes a different surface syntax which preserves s-expression structure under the hood but is indentation based, akin to Python.

See [wraith/wraith.md](https://github.com/cwebber/racket2-rfcs/blob/wraith-rebase/wraith/wraith.md) for a tutorial and examples, and comes with a working language.

As a bonus, [here is a substantial example of Wraith in use with Slideshow](https://gitlab.com/dustyweb/talks/-/blob/master/spritely/eu-presentation/eu-presentation.rkt).


slaymaker1907:
Is it a syntax error to provide more than one infix operation per grouping? I generally agree that separate operations should have grouping explicit by default, but only if this check can be done syntactically rather than at runtime.

In the reference section, I recommend also showing how the syntax is transformed into s-expressions like you do for your examples.

I know @mflatt really wanted support for algebra-style function calls.

cwebber:
It sounds like @AlexKnauth may be interested in carrying this in a more shrubbery-like direction, which is where it seems Rhombus will go.  I've expressed my reservations about that elsewhere, but I think deciding whether or not to leave this open or close it could depend on whether or not there's an interest in the shrubbery/wraith idea-merging.  I won't be contributing to that effort personally, but I think it's perfectly fine to continue using this PR for that purpose, if anyone wants to.

spdegabrielle:
You mentioned a hybrid approach briefly in the meeting on Thursday. I'd be interested in seeing what that could look like?

mflatt:
#527

PR #210, opened 2022-01-10
add plan as appendix

spdegabrielle:
an effort to reduce confusion about the project; the plan is currently available, but buried two levels down.


mflatt:
I agree that it makes sense to have this content directly in the repo, but would this be better as a separate plan document (linked from the README)?

spdegabrielle:
Closed replaced by #211

PR #215, opened 2022-04-01
Dynamic Dot - no closures

samdphillips:
This is a light modification of the existing dynamic dot lookup protocol.  When parens follow a dot expression, it sets a `fun?` flag and passes the arguments to `dot-lookup-by-name`.  This avoids the closure allocation but may instead cause a list allocation.

### DIFF
--- a/rhombus/private/dot.rkt
+++ b/rhombus/private/dot.rkt
@@ -9,6 +9,9 @@
          "static-info.rkt"
          "dot-provider-key.rkt"
          "realm.rkt"
+         "racket-class.rkt"
+         "parens.rkt"
+         "parse.rkt"
          "racket-class.rkt")
 
 (provide |.|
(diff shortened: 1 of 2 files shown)


mflatt:
I'm not sure sure about this one. I see how it lets dotted calls turn into method calls with less in between, but there are some small issues, at least. For one, it looks like there's a kind of mismatch between not handling  keywords in `dot-lookup-by-name` versus the use of `make-keyword-procedure` in `object-dot-lookup`. There is also potentially an issue that the function-call form `#%call` is bound differently than the default.

From the Rhombus meeting yesterday, I was left with the sense that it's better to not build function-call support into dynamic `.`, and instead rely on some static information to indicate that an expression produces an object (with certain methods?). Only then would `.` plus `()` be recognized as a method call (perhaps independent of the `#%call` binding).

samdphillips:
> I'm not sure sure about this one. I see how it lets dotted calls turn into method calls with less in between, but there are some small issues, at least. For one, it looks like there's a kind of mismatch between not handling keywords in dot-lookup-by-name versus the use of make-keyword-procedure in object-dot-lookup. There is also potentially an issue that the function-call form #%call is bound differently than the default.

Yes, this is more of an experiment to see how hard it would be to get something to play around with.  Using dynamic dot to access a function or closure (instead of a more complicated protocol) is probably better than trying to be clever.

> From the Rhombus meeting yesterday, I was left with the sense that it's better to not build function-call support into dynamic ., and instead rely on some static information to indicate that an expression produces an object (with certain methods?). Only then would . plus () be recognized as a method call (perhaps independent of the #%call binding).

I still think dynamic `.` could be useful but after more design of the object/class/method/whatever system is worked out.

mflatt:
Ok if we keep this open, though? That will make the experiment easier to find until we sort out everything.

mflatt:
#527

PR #222, opened 2022-05-27
scribble: add kind information

sorawee:
This depends on https://github.com/racket/scribble/pull/328
and will be effectively used by https://github.com/racket/racket/pull/4263


mflatt:
Part of 6762773f9 after similar changes as the linked PRs.

PR #224, opened 2022-06-23
Add Equality-related design documents (creating a folder for such documents)

countvajhula:
This adds the equality documents from the discussions a couple of months ago. They are in a `design` folder which can be used to hold other design documents and any supporting materials such as code.


countvajhula:
Can this be merged @mflatt ? The `design` folder can be used by subsequent PRs to add other design documents, including the process document.

PR #227, opened 2022-07-05
Add string case for ++ append

AlexKnauth:
Adds a string case for `++` append so that two strings can be appended with it. It does not coerce other values to strings like `&` does.


Lazerbeak12345:
That's really cool. Looks nice!

PR #228, opened 2022-07-06
Process docs

countvajhula:
This includes initial draft versions of:

- the process document, which describes Git workflows, the role of the design document and the editor, and how the editor is selected
~- an Editor's Guide, which is an FAQ for editors that conveys expectations and imparts advice (can be expanded over time with more lessons learned)~
~- a Contributor's Guide, which is an FAQ for contributors to design proposals and Rhombus in general~

These documents are in `design/meta/...`


mflatt:
Thanks for putting together this draft, @countvajhula! There are many good pieces here.

Still, the draft overall doesn't sound like what I understood from the meeting. The biggest difference is the separation here of a discussion phase from an implementation phase. I would like to see a process document that encourages implementation as a way of exploring design from the beginning. Also, my understanding of the end product was that it would be a kind of appendix to a complete design description. The appendix would relate ideas and alternatives that were discussed to the conclusion. In that way, the design document really should showcase a specific way of doing something as arrived at by consensus.

The draft document itself starts as a good example of what I think we want! It synthesizes an initial discussion into a concrete process design that's as close to an "implementation" as a process document can be. As a mere presentation suggestion, I would suggest dropping the "Why?" section, because it doesn't really say anything about the conclusion, but the rest is usefully concrete.

But I'm suggesting bigger changes than that. In the spirit of discussion through implementation, I offer a pull request on the proposal, https://github.com/countvajhula/rhombus-prototype/pull/1, which has the document https://github.com/mflatt/rhombus-prototype/blob/process/design/meta/process/process.md.

Small specific point: the proposed revision suggests Markdown and Scribble as document formats, dropping reStructuredText so that contributors don't have to learn so many formats.

countvajhula:
@mflatt I made the changes we discussed and a few others. Mind giving this another look? The fresh changes include:
- further emphasizing design document + prototype implementation as opposed to just the former
- tempering that by encouraging small POCs rather than Rhombus implementations at early stages of the design (to ensure that contributors are in a position to review the design at a high level before prematurely getting into the weeds)

mflatt:
@countvajhula — I'm happy with your latest version. Thanks!

countvajhula:
That's a great way of putting it @AlexKnauth . Couldn't SOP become combinatorial, though? i.e. for options A and B in two separate components, SOP would enumerate the four solutions AA, AB, BA, BB, whereas POS would be AB AB. It seems that this would grow rapidly with number of options.

AlexKnauth:
@countvajhula 
> Couldn't SOP become combinatorial, though? 

Yeah, but if it it would become too crazy or if the component-options are mostly independent, they could refer to previous component-options. That does disrupt the flow of that later option a bit though, so I expect that would be a last-resort.

For example: if there are components 1 and 2, which each have options a and b, then there could be four overall options: A = (1a, 2a), B = (1b, 2b), C = (1b, 2a), and D = (1a, 2b). In a sum-of-products structure that can refer to previous component-options, those overall options A and B could be described fully without breaking the flow, and then options C and D could refer to the previous component-options and just describe the consequences of them being combined differently.

countvajhula:
I could see this being a good format in some cases. I don't know how prescriptive we want to be about the format of the document, and I'd be inclined to leave this to the discretion of the editor, as long as it meets the stated goals in the process doc. Regarding the template, I think it would be fine to include more than one, if they are to serve as samples rather than strict requirements. Unless anyone has any objections, would you mind preparing a (minimal, like the existing one) SOP template to capture what you have in mind @AlexKnauth ?

AlexKnauth:
@countvajhula 
> would you mind preparing a (minimal, like the existing one) SOP template to capture what you have in mind?

Sure! I've made a PR: https://github.com/countvajhula/rhombus-prototype/pull/2

I based this second template more on the Typed Racket RFC template, with the biggest change being sections for multiple options.

countvajhula:
I've added a paragraph to the process doc mentioning the templates and the discretion of the editor regarding style. Let me know if that looks good to you @AlexKnauth .

@samdphillips @rocketnia @sorawee @notjack  👀 🔎 👍 ❓

sorawee:
LGTM. I agree that it should be left to the editor to find whatever format most suitable. 

Can't the two design templates be combined somehow? I find it confusing to have multiple.

Ironically, I find the second template somewhat difficult to follow, since I need to read Option A, B, C to get all content and paste them together. I think it might be easier to read if there's only one complete description, and the variants are just "...", like the first template.

On the other hand, I like that the second template is self-explanatory. Perhaps because GitHub shows the second template first, I expected the first template to be something similar, and was at first very confused "what the heck is 'A Relational Model of Data'?"

countvajhula:
@sorawee I'm not sure they can be combined since they may be relevant to two different kinds of proposals. In the way @AlexKnauth described them, they are essentially opposite of one another.

Re: the content of the two proposals being different, that's true. The POS template acts as a content example itself, while the SOP template describes the content to the reader. I don't have a strong opinion, but if we want, we could make them have the same content but presented in different structure. Personally, I would be inclined to change them both to resemble the POS content, since (a) some of the prescriptive material in the SOP template is already in the process doc, which is the main thing we are agreeing on here, so I am inclined to avoid duplication and conflicting advice here, and (b) the prescriptive content has some overlap with commonsense expectations and may be noise that the editor would ignore. In case we decide to go with this approach, then any prescriptive content that we feel is useful could instead be incorporated directly into the process doc.

AlexKnauth:
If you just want the templates to resemble each other more, I would prefer them both be self-describing, as the 2nd one was originally.

I've made a PR here: https://github.com/countvajhula/rhombus-prototype/pull/3

It also reorders/combines some sections such as Evaluation and Tradeoffs, and Prior art and References, and adds back the Drawbacks and Unresolved questions section to both templates, which was deleted from the 2nd one.

AlexKnauth:
It might be worth noting that the original template for racket2-rfcs here:
https://github.com/racket/rhombus-prototype/blob/7b1a4ef780ae1e8def01d62813e23a8cf5f61442/template.md

Was also self-describing in this way. I don't want to get rid of the benefits of that and replace it with "A Relational Model of Data" stuff.

countvajhula:
@AlexKnauth I have some reservations about putting prescriptive content in the templates. Self-describing is OK as long as it is empty of content, yet, it is a slippery slope to prescriptive since in the present case what is being described is the content of the document, and it becomes more prescriptive by becoming more descriptive!

As I see it, it runs the risks of:

- diverging in its guidance from the core documents such as the process document. For instance, although the newest versions you've created don't include anything conflicting (I'd need to take a closer look though - and this is partly my point), the earlier version mentioned things like "describe views expressed by others" -- when, rather, the point of the document is to _present the view of the group_, rather than the view of the editor with dissenters present, even if the editor is the one guiding the design, as will often be the case.
- being of lower quality since they are the _outcome_ of the agreement on core documents, and may not receive the same attention and scrutiny. Rebuttal: This is in our hands, and we can commit to giving the templates a lot of attention. Counter: This attention would be divided over all templates, and ensuring they are consistent with core documents is not easy. Why not just give all the attention to the core documents and let templates be empty structural shells? Counter-counter: They wouldn't be as useful if they were empty. Response: That's true. But what's important is usefulness of the documents taken together rather than the templates considered alone. And in this measure, the overall usefulness could be improved in other ways.
- discouraging editors from reading the core documents by presenting an appearance of completeness

Yet, I am not suggesting that the potential benefits in the prescriptive content should be "gotten rid of" but rather incorporated where it could be even more useful, comprehensive, and of high quality. I'm suggesting a dedicated Editor's Guide to hold this content instead, which we could see as being analogous to the web development practice of "fat models, thin templates". This has the following benefits and drawbacks:

Drawbacks:
- being one step removed from the template means it's less convenient and less in-context. Rebuttal: context would vary widely in practice, and editors may choose a quite different format. If the presentation is going to be irrelevant for their purposes, then they won't read the content either. And then we've missed our opportunity to convey useful advice to the editors that would aid high quality documents.
- It would be easier to just work with a template and avoid reading a whole other document. Rebuttal: being an editor is a lot of work. By capturing lessons and best practices, this guide could lighten the burden and save the editor time in the long run even if it is a modest additional upfront investment. I think the fact that you (Alex) are interested in preserving this content is an indication that we agree that such content is useful - the only question then is _where_ to put it.

Benefits:
- can be comprehensive and include advice and lessons learned that would not be appropriate for a template. We can teach the editors to fish and not just show them some fish
- as one focused document, easier to maintain than (content in) multiple distinct templates. Because of separation of concerns, this should keep the other documents (e.g. process document, in addition to the templates) leaner as well.
- gives us the flexibility to add as many templates as we like. Someone down the line may wish to recommend another format to future editors. The new template could be easily added without having to reconcile its content with that of the core documents, since it would have none (just what is necessary to convey the structure).

For these reasons, I would advocate for completely minimal templates that don't include any content, and I wouldn't be opposed to even removing them entirely. But, of course, I am keen to do what everyone feels is best, so if you disagree vehemently, or would like to agree vociferously, then please weigh in.

AlexKnauth:
An editor's guide will contain far more information than what's necessary for creating the first draft of a design document. Maintaining it as the design evolves, updating it after discussions, meetings, implementation changes, and so on.

But the template's purpose is different, more focused on convenience for getting the first version done.

countvajhula:
It can accomplish that goal without having prescriptive content, just headings -- which, if you prefer, we could make "self-describing". That is, headings only, and only structurally-necessary content.

AlexKnauth:
I'm not sure I understand your objections on it being too prescriptive.
> I don't know how prescriptive we want to be
> some of the prescriptive material in the SOP template is already in the process doc
> the prescriptive content has some overlap with commonsense expectations and may be noise that the editor would ignore
> any prescriptive content that we feel is useful could instead be incorporated directly into the process doc
> Re: the prescriptive content that was removed from the SOP template, ... But I feel that advice would be better incorporated into a dedicated "Editor's Guide" than a template.
> some reservations about putting prescriptive content in the templates
> it is a slippery slope to prescriptive
> it becomes more prescriptive by becoming more descriptive

I'm not a linguist, but from what I understand, prescriptive is about rules that must not be broken, while descriptive is more about describing things that have been said or can be said, without excluding other things. Web forms that must be filled out exactly as they are would be prescriptive, but a text document templates like this are more flexible than that, and seem more descriptive by nature. They're not rigid, these documents can be edited freely after copying the template.

I don't understand why this is an issue. I don't understand why the templates I've written and adapted from other design RFC templates are "prescriptive" or why they're a problem?

countvajhula:
I've accepted Alex's suggestions and trimmed the content down as we discussed. We could add content in response to concrete needs over time - that way, we can be sure that it's relevant and useful.

PR #229, opened 2022-07-07
Add str function, use with at-expressions

AlexKnauth:
Takes a single list argument because that's what at-expressions with `{}` expect.

```
> str(["the list ", [1, 2, 3], " has ", 3, " elements"])
"the list [1, 2, 3] has 3 elements"
> @str{the list @([1, 2, 3]) has @3 elements}
"the list [1, 2, 3] has 3 elements"
```

### DIFF
--- a/demo.rhm
+++ b/demo.rhm
@@ -216,7 +216,7 @@ bind.infoer 'build_reverse_cons_infoer($in_id, ($a_in, $b_in))':
   | '($a_ann, $a_id, $a_info, ($a_bind_info ...), $a_matcher, $a_binder, $a_data)':
       match bind_ct.unpack_info(b)
       | '($b_ann, $b_id, $b_info, ($b_bind_info ...), $b_matcher, $b_binder, $b_data)':
-          '($("matching((_ :: " & unwrap_syntax(a_ann) & ") <> (_ :: " & unwrap_syntax(b_ann) & "))"),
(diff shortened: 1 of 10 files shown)


mflatt:
The `str` function seems useful with `@`, but seeing the `&` examples converted to interpolation makes me think that interpolation is not always the best choice. Especially for someone new to the language, there's a lot more to explain with `@str{@also_x, @also_y}` than with `also_x & "," & also_y`.

There's even more to explain when changing `unwrap_syntax(a_ann)` to `unwrap_syntax[a_ann]`, but maybe that's a matter of improving `@` notation for Rhombus. Still, interpolation always seems to be complicated in the corners, since it's a new sublanguage. That complication can pay off for all but the most basic string-formatting problems, but basic string-formatting problems are common.

Formatting output with `<<` works nicely in C++, and `+` also works nicely in JavaScript when it's what you mean (but, of course, it's a problem that `+` is overloaded with addition). C++ has a slightly different `+` that avoids some of JavaScript's problems, but it's clumsier when you start with a string literal (a legacy of `char*` versus `std::string`). Experience with the good good cases in JavaScript and C++ is why I still lean toward something like a `&` operator for simple string- formatting, but with a better choice of operator than `&`.

As a intermediate point, if `str` could distinguish between regular arguments and arguments that come from `@....{}`, then `str(also_x, ",", also_y)` seems not bad, and it reduces the amount of stuff. Currently, though, an operator seems to me enough of an improvement to be worthwhile.

PR #231, opened 2022-07-12
revise syntax class documentation

mflatt:
Besides reorganizing the documentation a little, this patch drops `stx_class` in favor of `syntax.class`, which was also in
place before. (As `stx_class` was being developed, I had forgotten that a dummy `syntax.class` was in place already. Seeing it at the level of documentation made me think it might be better).

Part of the reorganization is a style for the reference area that is different than the guide area. A reference page is not so much meant to be read from top to bottom, but instead the description of something like `syntax.class` is meant to be all under the `doc` form for `syntax.class`. Another part of the reorganization splits things into separate sections; I'm not sure that's better, but I like it more after trying it.

Finally, `syntax` (to access `syntax.class`) moves here from `rhombus/macro` to just `rhombus`, since it's useful at phase 0 with `match` and similar. It gets imported at phase 1 by importing `rhombus/macro`, since `rhombus/macro` reexports all of `rhombus` at phase 1.

### DIFF
--- a/demo.rhm
+++ b/demo.rhm
@@ -700,9 +700,9 @@ for:
   ~each (key, val): {3: "c", 4: "d"}
   displayln(key & " -> " & val)
 
-// stx_class
+// syntax.class
 
-stx_class Arithmetic
+syntax.class Arithmetic
 | '$x + $y'
 | '$x - $y'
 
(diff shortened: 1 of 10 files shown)


tcallred:
This looks good 👍

PR #232, opened 2022-07-18
Custom attributes for syntax classes

tcallred:
This PR introduces a feature to syntax classes that allows users to define custom explicit attributes on top of the one implicit in the patterns. This also allows users to write arbitrary definitions after a pattern alternative and then choose definitions to expose with the `~attr id: value` form.

### DIFF
--- a/rhombus/private/syntax-class-syntax.rkt
+++ b/rhombus/private/syntax-class-syntax.rkt
@@ -37,11 +37,15 @@
                  ([g (in-list (syntax->list #'(groups ...)))])
          (syntax-parse g
            #:datum-literals (group block)
-           [(group #:attr id (block in-block ...))
-            (with-syntax ([(other ...) (reverse other-forms)])
+           [(group #:attr attr-id (block in-block ...))
+            (with-syntax ([(other ...) (reverse other-forms)]
+                          [((id id-ref) ...) (append idrs attrs)])
               (values
                other-forms
-               (cons #`[id (rhombus-body other ... in-block ...)] attrs)))]
+               (cons
+                #`[attr-id (let ([id id-ref] ...)
+                             (rhombus-body-at attr-id other ... in-block ...))]
+                attrs)))]
            [other
             (values (cons #'other other-forms) attrs)])))
      (define all-attrs (append idrs explicit-attrs))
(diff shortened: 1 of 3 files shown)


rhombus/private/syntax-class-syntax.rkt
```diff
@@ -1,30 +1,58 @@
...
+                  [attrs null]
+                  #:result (reverse attrs))
+                 ([g (in-list (syntax->list #'(groups ...)))])
+         (syntax-parse g
+           #:datum-literals (group block)
+           [(group #:attr attr-id (block in-block ...))
+            (with-syntax ([(other ...) (reverse other-forms)]
+                          [((id id-ref) ...) (append idrs attrs)])
+              (values
+               other-forms
```
mflatt, line 44: I think you meant `null` here, otherwise the forms already added via `other ...` get duplicated

rhombus/scribblings/ref-syntax-class.scrbl
```diff
@@ -36,17 +36,23 @@
...
 
  A variable bound with a syntax class (within a syntax pattern) can be
  used without dot notation. In that case, the result is a sequence of
  syntax objects corresponding to the entire match of a
- @rhombus[syntax_pattern], as opposed to an indvidual attributes within
+ @rhombus[syntax_pattern], as opposed to an individual attributes within
  the match. Use @rhombus[...] after a @rhombus[$]-escaped reference to
  the variable in a syntax template.
 
+Custom attributes of a syntax class can be defined in a block following a
```
mflatt, line 50: Besides this text, the `doc` form that defines the `syntax.class` grammar should be updated. Probably you want to replace `syntax_pattern` with `clause`, where clause is defined to be either `syntax_pattern` or `syntax_pattern: pattern_body; ...` and `pattern_body` is an expression or definition of `~attr` form. See `def` for an example of how to define helper nonterminals like `clause`.

rhombus/private/syntax-class-syntax.rkt
tcallred: I'm not sure if we will want that. If I replace `other-forms` with `null` here then it won't let any explicit attrs refer to earlier definitions. For example 
```
  | '~foo $x':
      def c:
        '14'
      def g:
        24 + unwrap_syntax(c) + unwrap_syntax(x)
      ~attr y:
        '$g'
      def a:
        5 + unwrap_syntax(y) + unwrap_syntax(c)
```
If we use `null`, this won't work because after `~attr y` the definition of `c` is not going to continue to be available for `a`.

tcallred: I see the problem now. Let's discuss a better approach to this.

PR #233, opened 2022-07-24
Repetition positions for ellipses in bindings and uses

mflatt:
Rendered form: https://github.com/mflatt/rhombus-prototype/blob/dots/design/sequence/repetition.md

The Rhombus prototype implementation and documentation are updated as part of this proposal.

### DIFF
--- a/demo.rhm
+++ b/demo.rhm
@@ -150,7 +150,7 @@ match cons(7, 8)
 
 match '(z + y, {[10, 11, 12]})'
 | '(x $a)': a
-| '($a + y, {[$n, ...]})': cons(a, n)
+| '($a + y, {[$n, ...]})': cons(a, [n, ...])
 
 cond
 | #true: 17
@@ -163,7 +163,7 @@ fun
 | factorial(n): n*factorial(n-1)
 
 expr.macro '$a *! $tail ...':
-  values('(factorial($a))', tail)
+  values('(factorial($a))', '$tail ...')
 
 10*!
 
@@ -184,7 +184,7 @@ just_five ++ " is the result"
 // another way to write that
 
 expr.macro 'also_prefix_plus $e ...':
-  match e
+  match '$e ...'
   | '$a $b $c ...':
       values(a, '+ $b $c ...')
   | ~else:
@@ -269,7 +269,7 @@ rx
 // or a block of definitions and a sequence of expressions
 
 defn.macro 'define_eight $e ...':
-  match e
+  match '$e ...'
   | '$name':
       'def $name: 8'
        
@@ -277,7 +277,7 @@ define_eight ate
 ate
 
 defn.macro 'define_and_use $e ...':
-  match e
+  match '$e ...'
   | '$name: $rhs ...':
       'def $name: $rhs ...
        literal_syntax($name)'
@@ -288,7 +288,7 @@ nine
 // declaration form
 
 decl.macro 'empty_import $e ...':
-  match e
+  match '$e ...'
   | '':
       'import:«»'
 
@@ -369,7 +369,7 @@ annotation.macro 'AlsoPosn': values('Posn', '')
 Posn(1, 2) :: AlsoPosn  // prints Posn(1, 2)
 
 bind.macro 'AlsoPosn ($x, $y) $tail ...':
-  values('Posn($x, $y)', tail)
+  values('Posn($x, $y)', '$tail ...')
 
 annotation.macro 'Vector':
   values(annotation_ct.pack_predicate('fun (x): x is_a Posn',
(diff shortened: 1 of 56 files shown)


demo.rhm
```diff
@@ -478,8 +480,8 @@ bob_loc
 def Map(symbol(alice), alice_loc2, symbol(bob), bob_loc2): local_map
 [alice_loc2, bob_loc2]
 
-def [ps :: Posn, ...] : [Posn(1, 2), Posn(3, 4)]
-ps[0].x
+def [p :: Posn, ...] : [Posn(1, 2), Posn(3, 4)]
+[p, ...][0].x
```
AlexKnauth, line 484: For subscripting with `[0]`, it would make sense to me to let `p[0]` refer to the first element of the repetition `p`.

Can array access accept a repetition without creating ambiguity?

AlexKnauth: Perhaps the repetitions inside array access could be restricted to just variables and further nested array access.

AlexKnauth: This case shows a conflict:
```
def [ps :: List.of(Posn), ...] : [[Posn(1, 2), Posn(3, 4)], [Posn(5, 6), Posn(7, 8)]]
[length(ps[0]) + ps[0].x, ...]
```
If it's greedy from the bottom up, then the `length(ps[0])` will be fine as a single access but the `ps[0].x` will be wrong.
If it goes from the top down, then the `length(ps[0])` will be wrong while the `ps[0].x` is a repeated access.

mflatt: Allowing a repetition before `[0]` maybe could probably work, but with the current implementation, it would have to be limited somehow.

The intent with the current design is that a term is in an expression position or a repetition position based on its context. The implementation relies on that constraint; concretely, an attempt to match the `:repetition` syntax class internally either succeeds with an expansion or fails with an error, and the same is true of `:expression`. (In contrast, `:definition` is set up to fail with a non-match if the immediate term doesn't start a definition form, so it makes sense to try `:definition` and then `:expression` in a body sequence.) As you suggest, there would have to be some sort of constraints on either repetitions or expressions in an overlapping position so that expansion could pick the right one.

More generally, though, I think your example of ambiguity points to the difficulty of making two different kinds of inference work together: inferring a repetition versus an expression and inferring a repetition from an expression-like term in a repetition position. I'm not really sure either of those is a good idea, although the latter seems more useful, and I would be inclined (so far) to avoid the former.

mflatt:
There are still open questions about how to make repetitions convenient (especially: inferring repetitions from expression-shared terms in a repetition position), but the general approach here of repetition contexts doesn't seem to be controversial. So, I'm planning to merge soon.

mflatt:
The latest prototype supports a lot more `...`.

```
def [x, ...] = ["a", "b", "c"]
def [[z, ...], ...] = [[1, 2, 3], [4, 5, 6]]

// consecutive `...`s flatten, like Racket templates
[z, ..., ...]  // => [1, 2, 3, 4, 5, 6]

// operators defined by `operator`, including predefined ones like `+&`,
// work as repetition forms; they have subrepetitions instead of
// subexpressions
[x +& x, ...]  // => ["aa", "bb", "cc"]

// mixing depths also works as in templates
[z +& x, ..., ...]  // => ["1a", "2b", "3c", "4a", "5b", "6c"]

// literals and variables work as depth 0 repetitions, which
// allows them to be combined with deeper repetitions
[z+1, ..., ...]  // => [2, 3, 4, 5, 6, 7]

// `.` is like an operator
class Posn(x, y)
def [p, ...] = [Posn(1, 2), Posn(3, 4)]
[p.x, ...]  // => [1, 3]

// ditto for `[]`
def m = {"a": "apple", "b": "banana", "c": "coconut"}
[m[x], ...]  // => ["apple", "banana", "coconut"]

// ditto for function calls
fun greet(s):
  "hi " +& s
[greet(x), ...]  // => ["hi a", "hi b", "hi c"]

// maps and sets support `...` to splice in repetitions
{z: x, ..., ...}  // => {1: "a", 2: "b", 3: "c", 4: "a", 5: "b", 6: "c"}
{x, ...}          // => {"a", "b", "c"}
{x, ..., "a"}     // => {"a", "b", "c"}
{x, ..., z, ..., ...} // => {1, 2, 3, 4, 5, 6, "a", "b", "c"}

// maps and set bindings also support `...` (but only at the end,
// which is the same constraint as for list bindings currently):
def {key: val, ...} = {z: x, ..., ...}
[key, ...]  // => [1, 2, 3, 4, 5, 6]
```

Making things like operator uses, function calls, and `.` work as repetitions relies on specifically implementing those forms as repetitions. There's no attempt to automatically coerce an expression into a repetition, especially not with subrepetitions. That choice avoids potential ambiguity and dealing with macros that treat some subforms specially. The drawback, though, is that we have to pick forms to act as repetitions; for example, it would probably make sense to add `::` and `-:`, but I didn't consider it until just now. Also, macro implementers have to consider whether they want to provide a repetition variant of each new form and figure out how to implement it (when I add the repetition protocol to the meta API).

I've refrained from adding repetition variants of forms that include binding or body blocks (which might have definitions), because that gets complicated. It would make sense to support constrained variants of some forms with body blocks, such as an `if` variant that requires a single repetition in each branch. I'm not sure whether those constrained forms would be a good idea, so I've left them out to keep things simpler.

PR #234, opened 2022-07-28
Syntax class description

tcallred:
This adds the optional `description` field for syntax classes. It is equivalent to the `#:description` attribute on Racket syntax classes, so the field can be populated with any expression that evaluates to a string or `#false` and it will appear in error messages where a term is rejected by the syntax class.

### DIFF
--- a/rhombus/private/syntax-class-syntax.rkt
+++ b/rhombus/private/syntax-class-syntax.rkt
@@ -98,7 +98,7 @@
                   (generate-pattern-and-attributes alt-stx))])
     (list
      #`(define-splicing-syntax-class #,class-name
-         #:description #,(if description #`(rhombus-body #,description) #f)
+         #:description #,(if description #`(rhombus-body #,@description) #f)
          #:datum-literals (block group quotes)
          #,@patterns)
      #`(define-syntax #,(in-syntax-class-space class-name)
@@ -142,7 +142,7 @@
         ;; Specify patterns with "pattern"
         [(form-id class-name
                   (block
-                   (~optional (group description (block class-desc)))
+                   (~optional (group description (block class-desc ...)))
                    (group pattern (alts alt ...))))
          (generate-syntax-class stx #'class-name (syntax->list #'(alt ...)) (attribute class-desc))]
         [_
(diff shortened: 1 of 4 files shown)


rhombus/private/syntax-class-syntax.rkt
```diff
@@ -138,10 +138,12 @@
...
         ;; Classname and patterns shorthand
         [(form-id class-name (alts alt ...))
-         (generate-syntax-class stx #'class-name (syntax->list #'(alt ...)))]
+         (generate-syntax-class stx #'class-name (syntax->list #'(alt ...)) #f)]
         ;; Specify patterns with "pattern"
         [(form-id class-name
-                  (block (group pattern (alts alt ...))))
-         (generate-syntax-class stx #'class-name (syntax->list #'(alt ...)))]
+                  (block
+                   (~optional (group description (block class-desc)))
```
mflatt, line 145: I think the block for `description` should be allowed to have multiple groups.

rhombus/scribblings/ref-syntax-class.scrbl
```diff
@@ -10,12 +10,16 @@
...
   defn.macro '«syntax.class $name:
+                $maybe_description
                 pattern
                 | $clause
                 | ...»',
   defn.macro '«syntax.class $name
                | $clause
                | ...»',
+  grammar maybe_description:
+    description: $body
```
mflatt, line 21: Along the same lines as the implementation, this should be `$body; ...` to allow multiple groups.

rhombus/scribblings/ref-syntax-class.scrbl
```diff
@@ -33,7 +37,12 @@
  @rhombus[::]. The @rhombus[pattern] subform is optional in the sense
  that pattern alternatives can be inlined directly in the
  @rhombus[syntax.class] form (but the @rhombus[pattern] subform makes
- room for additional subforms in the future).
+ room for additional subforms in the future). The optional
+ @rhombus[description] subform can be used before @rhombus[pattern] to define
+ the description of the syntax class. This description is used to produce 
+ clearer error messages when a term is rejected by the syntax class. 
+ It takes a block with an expression that may evaluate to a string or 
+ @rhombus[#false].
```
mflatt, line 45: Maybe start a new paragraph for `The optional @rhombus[description]`? I'd use `must` instead of `may`. With the other changes, I'd write `a block whose result is` (since there's not just one expression). Clarify that `#false` is equivalent to not specifying `description`, which means that the error message is not customized.

mflatt (changes requested):

PR #235, opened 2022-07-29
Use hashalw for maps and sets instead of hash

AlexKnauth:
Since the default equality operator `==` has equal-always behavior, maps and sets should be the same.


mflatt:
Thanks!

PR #236, opened 2022-07-29
Keyword Rest implementation

AlexKnauth:
An implementation of https://github.com/racket/rhombus-prototype/pull/225 option A: Rests with `&` and `~&`, Repetitions with `...`, Strings with `+&`.

Things left to do:
- [x] Implement keyword-apply-like function calls with `~&` keyword rest
- [x] Document new features
- [x] Test `~&` keyword rest thoroughly
- [x] Implement `&` rest for more data constructors
  * [x] List exprs/binds with `[e, ..., & l]` and `List(e, ..., & l)`
  * [x] Map exprs/binds with `{k: v, ..., & m}`
  * [x] Map exprs/binds with `Map{k: v, ..., & m}`
  * [x] Map expressions with `Map([k, v], ..., & l)`
  * [x] Set expressions with `{e, ...+, & s}`
  * [x] Set expressions with `Set{e, ...+, & s}`
  * [x] Set expressions with `Set(e, ...+, & l)`
- [x] Implement empty map binding, usable for something like `{k: v, ..., & Map.empty}`
- [x] Resolve issues with `Map()` and `&` rest not working the same way as `{}` and `&` rest (`Map{}` macro, `Map()` function with keys/values grouped together instead of alternating, no `&` rest in `Map()` binding for now, but with the possibility of resolving that in the future via entry-sequences suggested by @jackfirth)
- [x] Resolve issues with matching `&` rest of mutable data (~error, pattern-fail~, copy, ~or alias-slice?~)
- [x] Implement `~&` keyword rest for `|` alt functions
- [x] Implement keyword arguments for `|` alt functions, cooperating with `~&` keyword rest
- [x] Check `raco expand` on test cases like the one below with `(#%app f temp8 temp7)`
- [x] Test everything thoroughly
- [ ] Clean up code that ends up unused
- [x] Rebase / Fix merge conflicts

From the Rhombus meeting on 2022-08-18:
```
#lang rhombus

fun f(x, ~b):
  b

fun g():
  f(1, ~b: 2)
```
the expansion for `g` should ideally contain something like `(if (#%app variable-reference-constant? (#%variable-reference f6)) (#%app f temp8 temp7) ....)` in its body.


demo.rhm
```diff
@@ -216,7 +216,7 @@ bind.infoer 'build_reverse_cons_infoer($in_id, ($a_in, $b_in))':
   | '($a_ann, $a_id, $a_info, ($a_bind_info ...), $a_matcher, $a_binder, $a_data)':
       match bind_ct.unpack_info(b)
       | '($b_ann, $b_id, $b_info, ($b_bind_info ...), $b_matcher, $b_binder, $b_data)':
-          '($("matching((_ :: " & unwrap_syntax(a_ann) & ") <> (_ :: " & unwrap_syntax(b_ann) & "))"),
+          '($("matching((_ :: " +& unwrap_syntax(a_ann) +& ") <> (_ :: " +& unwrap_syntax(b_ann) +& "))"),
```
sorawee, line 219: If we all agree that `&` should be changed to `+&` to make room for the rest operator, can we make this change as a separate PR? I think it unnecessarily makes this proposal difficult to read.

AlexKnauth: Sure. I made it as a separate commit so I can do that.

AlexKnauth: Okay, I made a separate PR: https://github.com/racket/rhombus-prototype/pull/237

PR #239, opened 2022-08-05
updates to `for`

mflatt:
Most of the changes here are based on trying suggestions from the [January 21 meeting](https://github.com/racket/rhombus-prototype/discussions/180#discussioncomment-2010674).

First, the term "folder" has been replaced with "reducer".

A reducer can now be specified at the end of a `for` block using `~into`, like this:

```
for:
  ~each i: 1..4
  i
  ~into List
```

But a reducer can also still be specified at the top, as in `for List: ....`. Sometimes it reads better to have the reducer first, and sometimes it reads better at the end, so I decided to try allowing both.

The changes here don't yet try to add accumulators into the body of `for`. I think that makes sense for adding extra accumulators that aren't part of the result, but haven't yet tried that change.

A new `~do` form supports macro extension of `for` bodies, analogous to `#:splice` in Racket's `for`. The names `each`, etc., are bound as for-clause forms that just expand to `~each`, etc. Alternatively, you can consider `~each` as a
shorthand for `~do each`. The rationale here is that recognizing a for-clause form like `each` by itself (without `~do`) would mean that expressions and for-clause forms overlap, which I consider an anti-pattern because it constrains other syntactic choices. Requiring `~do` keeps the two kinds of positions separate. Meanwhile, the predefined `for` controls are common enough that it seems worth having shorthands.

Instead of `~and` as a variant of `~each` that must follow an `~each`, the `~each` form supports an immediate block containing binding sequences to run in parallel. Compare

```
for List:
  ~each i: 0..4
  ~each j: 0..4
  [i, j]
```

to

```
for List:
  ~each:
    i: 0..4
    j: 0..4
  [i, j]
```

I think I originally went with `~and` because I didn't want the bindings for sequences to look like they're in a more nested scope. But now that seems less important than seeing clearly how sequences are grouped. Also, this form plays better with `~do each`.

Finally, the changes here rename `~break` and `~final` to `~break_when` and `~final_when`, because it seems clearer and reads better, and the extra `_when` is a small price to pay for rarely used forms.


Lazerbeak12345:
A thought: instead of calling it `~break_when` and `~final_when` have `~break~when` and `~final~when` or `~break-when` and `~final-when` been considered? (the last pairing I like the most, perhaps due to racket's current naming scheme)

mflatt:
The current choice `~break_when` is a single keyword. In contrast, `~break~when` is two keywords (with no space in between), and `~break-when` is a keyword, the minus operator, and an identifier. I don't have strong opinions about the name, but probably we want it to be a single keyword. If it's two words within a keyword, then using `_` follows Rhombus conventions — so far, at least.

gus-massa:
Is it possible to apply two `reducers` and get two results?

Option a) I want to calculate the `min `and `max` of a bunch of numbers

    for:
      ~each i: 1..20
      i*i mod 100
      ~into Min, Max

Option b) In each step generate two things and fold each one in a different way: 

    for:
      ~each i: 1..20
      values(i, even(i))
      ~into ~values(Sum, And)

    for:
      ~each i: 1..20
      values(i, even(i))
      ~into Sum
      ~into And

Option c) Combine a) and b) and get 3 or 4 results

mflatt:
> Is it possible to apply two `reducers` and get two results?

That sounds potentially useful, but instead of something that's built into `for`, I expect that to be a general reducer that combines other reducers.

mflatt:
Merged as 9c930ae5ab51a6f1efe1afc6aa1db35cfc801dd9.

PR #241, opened 2022-08-07
Add Map.empty a binding as well as an expression

AlexKnauth:
An empty map. Differs from `{}` and `Map()` because while those as binding forms match any maps including ones with extra keys, `Map.empty` as a binding only matches empty maps, with no keys.

Also changes `Map.empty` to a `hashalw`, not an equal-based `hash`.


gus-massa:
I'm out of the loop here, so sorry if my question makes no sense. What's the result of this?

    match {"x": 7, "y": 8}
    | {"x": 7}: 0
    | {}: 1
    | _: 2

AlexKnauth:
The result of that is `0` because the first pattern `{"x": 7}` matches the target value `{"x": 7, "y": 8}` by `{}` patterns allowing extra keys.

PR #242, opened 2022-08-17
nested bindings and unified`import`

mflatt:
This set of commits is an experiment with a form for created nested bindings (along the lines of `List.length` and `List.cons`) and unifying those names with module imports as much as possible. Example (edit: adjusted capitalization):

```
   // defines `math` with `math.tau` and `math.complex`:
   nest math:
     export:
       tau
       Complex
     val pi: 3.14
     val tau: 2 * pi
     class Complex(real, imag)

   math.tau                   // prints 6.28
   // math.pi                 // would be an error
   math.Complex(0, math.tau)  // prints Complex(0, 6.28)

   // more nesting:
   nest subject:
     export:
       math
       english
     nest english:
       export: greeting
       val greeting: "Hello"

   subject.english.greeting  // prints "Hello"
   subject.math.tau          // prints 6.28

   // unpack a nested namespace using `import`, 
   // but put `.` in from of the defined nested name to
   // distinguish it from a module path
   begin:
     import: .subject open
     english.greeting        // prints "Hello"
     math.tau                // prints 6.28

   // `import` works in any block
   val also_pi:
     import: .math open
     tau / 2
   also_pi

   // `.` works on a module path in `import`, too, as a kind of shorthand
   begin:
     import: rhombus.List open
     length(["a", "b", "c"])

   // a prefix bound by `import` works like a name bound by `nest`,
   // and it can be exported
   nest lang:
     import:
       rhombus as rh
       racket/base as rk
     export: rk rh    
   lang.rh.Array.length        // prints #<function:Array.length>
   lang.rk.#{vector-length}    // prints #<function:Array.length>
```

Some notes and issues:

 * This is not about submodules. As illustrated above, a `nest` binding can be used later in the same module, but its body can see enclosing bindings. It's more like Chez Scheme's `module` in that sense. We'll also want submodules, TBD.

 * Despite the veneer of a unified import and export syntax, modules and nesting names behave a differently in the corners. You can only use `for_meta` on a reference to a module, for example, although that reference could take the form of a prefix given to an import, like `rk` and `rh` above. Also, module imports can be shadowed by definitions in the same body, while nesting imports are like definitions.

 * On the leading `.` to distinguish nesting names versus collection-based module paths: I experimented with inferring whether a name refers to a binding like `List` or a module path like `rhombus`, but it was messy enough that I suspect that it's a bad idea. Even if making a syntactic distinction is a good choice, the syntax of a module path and/or a nesting reference could be different.

 * The implementation is complex, and I don't think it's quite right, yet. (Also, tests sometimes fail in a way that makes me suspect a bug in the expander.) The complexity mostly has to do with implementing things twice for modules versus nestings (needs to preserve the representation optimizations for module imports and exports as well as deal with phase shifting), while also dealing with each import and export operations being implicitly across all phases, and also making sure that as much as possible is done via portal bindings so that it's documentation-linking friendly.


mflatt:
More notes:

 * The definition form `nest` could have a different name, of course.

 * More generally, the draft terminology "nesting" names is awkward. There's a relationship to "dot provider", but that's yet another mechanism internally; surely the terminology can be integrated better, if not the implementation.

mflatt:
I accidentally pushed this to the main branch. Since it's in a working state, I'm leaving it pushed for now — but the intent is that this discussion is still open!

mflatt:
#243 is the design document — originally planned to be part of this PR, but not now that one takes over.

PR #249, opened 2022-10-29
classes, fields, constructors, and methods

mflatt:
Evolving design document is at https://github.com/mflatt/rhombus-prototype/blob/class/design/class/class.md

This is still very much a work in progress, and methods aren't in place, yet. But this overall approach to the syntax seems to be working, and field declaration and constructor-customization part is maybe all there.


design/class/class.md
```diff
@@ -0,0 +1,359 @@
...
+
+Open Issues
+-----------
+
+An uncooperative custom constructor for a non-final class might return
+an instance of itself or some other subclass when called on behalf of
+the constructor for a different subclass.
+
+Class clause macros should be able to receive information about the
+fields of the enclosing class. Moving `extends` out of the clas sbody
```
Lazerbeak12345, line 342:
```suggestion
fields of the enclosing class. Moving `extends` out of the class body
```

mflatt: Fixed - thanks!

design/class/class.md
```diff
@@ -0,0 +1,394 @@
...
+simple pattern-matching macro transformations, which is useful with
+`binding` and `anotation`. (When `constructor` is specified, then
+typically `binding` and `annotation` should also be specified.) The
+`entry_point` for `binding` and `annotation` is a meta-time
+expression.
+
+A class is `final` unless `nonfinal` or `extends` is present (i.e., by
+default, classes do not have subclasses). A `nonfinal` clause is
+implied when `extends` is present and `final` is not present (i.e., by
+default, a subclass can have surther subclasses).
```
sorawee, line 130: further

mflatt: Fixed - thanks!

mflatt:
Now with methods and interfaces — and worth a closer look from anyone who is interested.

Probably most of the details will seem obvious, although specific syntactic choices are always debatable. The least obvious parts of the design (still with some precedents) are (1) the handling of constructor, binding, and annotation customization with subclassing, and (2) the notion of implementing an interface privately.

Remaining work includes defining interfaces for equality and printing, sorting out method access via a class name (as opposed to an object annotated with the class), and writing overview documentation and examples in the reference.

mflatt:
Merged as c917717e710ddb626643626b85301cbf11daab0d

PR #255, opened 2022-12-15
allow `=` instead of `:` in `def`, `let`, and `field`

mflatt:
Using `=` for `def` and similar forms reduces the sea of `:`s, and it connects better to using `=` for default-value expressions for function arguments and class fields. When I compare the two `namespace` blocks below, the `=` form is clearly more readable.

```
class Posn(x, y)

namespace A:

  def origin: Posn(0, 0)
  def p :: Posn: origin
  def Posn(x, y): origin

namespace B:

  def origin = Posn(0, 0)
  def p :: Posn = origin
  def Posn(x, y) = origin
```

Using `:` is still allowed, and it looks nicer for multi-line definitions than having to use `begin`.

```
namespace A:

  def one_thirty:
    def pi = 3.14
    Posn(cos(pi/4), sin(pi/4))

namespace B:

  def one_thirty = begin:
    def pi = 3.14
    Posn(cos(pi/4), sin(pi/4))
```

The difference between `:` and `= begin:` is much smaller than `:` versus `=`. We could support only `=`, but somehow it feels more consistent to me to still allow `:` in `def`, like other forms. I think of `=` as a readable shorthand, similar to way that `fun f(): ....` is a preferred shorthand to `def f = fun(): ....`.

For consistency, `:` is also allowed instead of `=` when providing a default-value expression for an optional functional argument or constructor field. I am not sure this matters.

There is no change to shrubbery notation to support `=`. An alternative is that `=` becomes a block-creating character like `:`. Changing shrubbery notation in that way would make parsing/matching on `=` forms easier, and it would impose a shrubbery-level grouping on uses of `=`. But it would also make shrubbery notation more specific to Rhombus. Since`=` is going to be use in many fewer places than `:`, and since constraining `=` to a single-line continuation might be a good choice anyway, this proposal leaves parsing to the small number of places that will recognize `=`: `def`, `let`, and `field` (in addition to the function-argument and field forms with defaults, which already had to do this kind of parsing).

Disadvantages of this proposal:

 * There's more than one way to write `def` to `let` or `field`, so a programmer has to pick.
 * Syntax matching on `=` forms is clumsy compared to matching `:`.

Advantages of this proposal:

 * There's more than one way to write `def`, etc., so a programmer gets to pick.
 * Using `=` in the relevant positions is more readable than `:`.
 * Typing something like `def x = 5` in a REPL is much nicer than `def x: 5` followed by an extra newline.
 * Allowing `:` preserves a connection to most other forms, and is sometimes tidier.

### DIFF
--- a/demo.rhm
+++ b/demo.rhm
@@ -58,17 +58,17 @@ md2(Posn(1, 4))
 
 // more definitions
 
-def π: 3.14
+def π = 3.14
 π
 
-def (((ππ))): π * π
+def (((ππ))) = π * π
 ππ
 
 def (ma, mb, mc):
   values("1", "2", "3")
 mb
-  
-def List.cons(ca, cb): List.cons(1, [2])
+
+def List.cons(ca, cb) = List.cons(1, [2])
 ca
 
 fun
@@ -82,10 +82,10 @@ fun
 size(Posn(8, 6))
 size(1, 2)
 
-def Posn(px, py) -: Posn: Posn(1, 2)
+def Posn(px, py) -: Posn = Posn(1, 2)
 List(px, py)
 
-def identity: fun (x): x
+def identity = fun (x): x
 identity(1 + (fun (x): x) (99) )
 
 size
@@ -210,8 +210,8 @@ bind.macro '$a <> $b':
                    ($a, $b))')
 
 bind.infoer 'build_reverse_cons_infoer($in_id, ($a_in, $b_in))':
-  def a: bind_meta.get_info(a_in, '()')
-  def b: bind_meta.get_info(b_in, '()')
+  def a = bind_meta.get_info(a_in, '()')
+  def b = bind_meta.get_info(b_in, '()')
   match bind_meta.unpack_info(a)
   | '($a_ann, $a_id, $a_info, ($a_bind_info, ...), $a_matcher, $a_binder, $a_data)':
       match bind_meta.unpack_info(b)
@@ -262,7 +262,7 @@ bind.binder 'build_reverse_cons_bind($in_id, ($a, $b, $a_part_id, $b_part_id))':
 // an expression operator that's consistent with the pattern
 expr.rule '$a <> $b': ~parsed_right; 'Pair.cons($b, $a)'
 
-def rx <> (ry :: Integer) : "2" <> 1
+def rx <> (ry :: Integer) = "2" <> 1
 rx
 
 // definition form, which returns either a block of definitions
@@ -271,7 +271,7 @@ rx
 defn.macro 'define_eight $e ...':
   match '$e ...'
   | '$name':
-      'def $name: 8'
+      'def $name = 8'
        
 define_eight ate
 ate
@@ -304,7 +304,7 @@ let accum: accum+1
 let accum: accum+1
 accum
 
-def ok_later: "ok"
+def ok_later = "ok"
 check_later()
 
 // Nested class types with contracts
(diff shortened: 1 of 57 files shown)


mflatt:
Merged as d458a16eb1605932d397827c6630387bceab6122, but leaving the PR open for a while in case there's discussion.

jackfirth:
Thoughts so far:

1. I expected `def foo() = 42` to work. I was thinking this would be something along the lines of Kotlin's expression body functions, where `lhs = rhs` is a substitute for `lhs { rhs; }` in many contexts, including function and method definitions.

2. I also expected this to work:

```
def foo =
  bar()
```

My thinking is that `=` looks best when used for cases where the right-hand side is a single expression, while `:` looks best when it's multiple expressions. I can still get this to work by keeping the first part of the expression on the same line, but that leads to difficult to read code like this:

```
def someLongButUsefulVariableName = someLongButUsefulFunctionName(
  arg1, arg2, arg3)
```

That looks... very weird, mostly because I usually try to write code in a style that's consistent with the [Rectangle Rule](https://github.com/google/google-java-format/wiki/The-Rectangle-Rule).

3. What if we just disallowed a `:` with a same-line right-hand side entirely? That would eliminate the possibility of ambiguity.

mflatt:
>     1. I expected `def foo() = 42` to work. 

If I understand what you mean, I'm not in favor of this. We just got rid of a `def` that tires to infer function binding versus other binding forms (i.e., when `foo` is a binding form), and it seems like the wrong direction to go back down.

>     2. I also expected this to work:
> 
> ```
> def foo =
>   bar()
> ```

This cannot work with changing the meaning of `=` at the shrubbery level, which I don't think we should do.

> My thinking is that `=` looks best when used for cases where the right-hand side is a single expression, while `:` looks best when it's multiple expressions.

The guideline that fits well with shrubbery notation overall is `=` when the whole definition is one line and `:` when the whole definition takes multiple lines. Your guideline sounds fine by itself, but the shrubbery-consistent guidelines make everything else fit together, too.

>  I can still get this to work by keeping the first part of the expression on the same line, but that leads to difficult to read code like this:

I would write that with `:`:

```
def someLongButUsefulVariableName:
  someLongButUsefulFunctionName(arg1, arg2, arg3)
```

>     3. What if we just disallowed a `:` with a same-line right-hand side entirely? That would eliminate the possibility of ambiguity.

This kind of constraint should be at the shrubbery level, but there are lots of places where where want the part after `:` to be allowed on the same line.

AlexKnauth:
Would `fun foo() = 42` working instead of `def foo() = 42` make sense then?

mflatt:
Allowing`fun foo() = 42` could work. Here are some small arguments against:

* The `=` doesn't have same clarifying role as it does in `def` to separate a somewhat complicated binding pattern from the right-hand side, since the parentheses around the function arguments already do that. There could be a complicated result annotation for a function, but that's a rarer case that can be clarified by using multiple lines.

* Allowing `=` in the expression form of `fun` doesn't seem great to me (and maybe we wouldn't want to do that).

* From my uses so far, such as updating `demo.rhm`, `=` seems useful as a redundant signal for `def` as opposed to `fun`.

jackfirth:
`fun foo() = 42` is what I had in mind, not `def foo() = 42`. Well, really, my use case was `method foo() = 42` and in translating it to not be class-specific I forgot that `def` no longer does what it used to.

sorawee:
I recall in a meeting last month, there's an error when you use `=` in a macro definition, but there's no error when you use `:`. I thought that was weird because it means `=` and `:` are not truly interchangeable. 

@mflatt, can you show that example (at the very least for the record)? Why is the restriction needed?

mflatt:
The error that we hit during the meeting no longer happens, because multi-term syntax objects now splice into term positions in a template.

In general, you can still get an error by trying to use a multi-group term in a term position. That couldn't have happened with our example in the meeting, though, because the matching half would always produce a single-group but possibly multi-term match).

rocketnia:
I love this change, thank you. I suspect it will make Rhombus codebases more approachable at first glance for people coming from the countless other languages that use `=` or `:=` for variable initialization.

PR #265, opened 2023-01-06
Add a basic `Equatable` interface

jackfirth:
This interface doesn't provide much more than the bare bones of an equality API. All it provides is this:

```
interface Equatable:
  method equals(other) :: Boolean
  method hashCode() :: Fixnum
```

That's it. No support for a `recur` argument, distinguishing between `equal?` and `equal-always?`, or an equality key is provided. This is all that's needed to get Rhombus collections working. Objects that implement this interface will get an implementation of `prop:equal+hash`, but that implementation will not use the `recur` or `mode` arguments.

I **strongly recommend** we avoid adding more to this API. But I don't want to litigate that in this pull request; for now, I'm just implementing what's needed to make generic collections work.

### DIFF
--- a/rhombus/private/equatable.rkt
+++ b/rhombus/private/equatable.rkt
@@ -1,0 +1,91 @@
+#lang racket/base
+
+(require (for-syntax racket/base
+                     "interface-parse.rkt")
+         "name-root.rkt"
+         (submod "annotation.rkt" for-class)
+         (submod "dot.rkt" for-dot-provider)
+         "realm.rkt"
+         "class-dot.rkt"
+         (only-in "class-desc.rkt" define-class-desc-syntax))
+
+(provide Equatable
+         (for-space rhombus/class Equatable)
+         (for-space rhombus/annotation Equatable))
+
+
+(define-values (prop:Equatable Equatable? Equatable-ref)
+  (make-struct-type-property
+   'Equatable
+   #false
+   (list (cons prop:equal+hash (lambda (_) bounced-equal+hash-implementation)))))
+
+(define-values (prop:Equatable-public Equatable-public? Equatable-public-ref)
+  (make-struct-type-property 'Equatable #false (list (cons prop:Equatable values))))
+
+
+(define (bounce-to-equal-mode-proc this other recur mode)
+  (equals-internal-method this other))
+
+(define (bounce-to-hash-mode-proc this recur mode)
+  (hashCode-internal-method this))
+
+(define bounced-equal+hash-implementation
+  (list bounce-to-equal-mode-proc bounce-to-hash-mode-proc))
+
+
+(define-syntax Equatable-internal
+  (interface-desc #'Equatable-internal
+                  #false
+                  #'()
+                  #'prop:Equatable
+                  #'Equatable-ref
+                  (vector-immutable (box-immutable 'equals) (box-immutable 'hashCode))
+                  #'#(#:abstract #:abstract)
+                  (hasheq 'equals 0 'hashCode 1)
+                  #hasheq()))
+
+(define-class-desc-syntax Equatable
+  (interface-desc #'Equatable
+                  #'Equatable-internal
+                  #'()
+                  #'prop:Equatable-public
+                  #'Equatable-public-ref
+                  (vector-immutable (box-immutable 'equals) (box-immutable 'hashCode))
+                  #'#(#:abstract #:abstract)
+                  (hasheq 'equals 0 'hashCode 1)
+                  #hasheq()))
+
+(define-name-root Equatable
+  #:fields
+  ([equals equals-method]
+   [hashCode hashCode-method]))
+
+(define-annotation-syntax Equatable
+  (identifier-annotation
+   #'equatable-interface #'Equatable-public? #'((#%dot-provider equatable-instance))))
+
+(define-dot-provider-syntax equatable-instance
+  (dot-provider-more-static (make-handle-class-instance-dot #'Equatable #hasheq() #hasheq())))
+
+(define (get-equatable who v)
+  (define vt (Equatable-public-ref v #false))
+  (unless vt
+    (raise-argument-error* who rhombus-realm "Equatable" v))
+  vt)
+
+(define equals-method
+  (let ([equals (lambda (this other)
+                  ((vector-ref (get-equatable 'equals this) 0) this other))])
+    print))
+  
+(define hashCode-method
+  (let ([hashCode (lambda (this)
+                 ((vector-ref (get-equatable 'hashCode this) 1) this))])
+    hashCode))
+
+(define (equals-internal-method v op)
+  ((vector-ref (Equatable-ref v) 0) v op))
+
+(define (hashCode-internal-method v op)
+  ((vector-ref (Equatable-ref v) 1) v op))
--- a/rhombus/private/equatable.rkt
+++ b/rhombus/private/equatable.rkt
@@ -77,7 +77,7 @@
 (define equals-method
   (let ([equals (lambda (this other)
                   ((vector-ref (get-equatable 'equals this) 0) this other))])
-    print))
+    equals))
   
 (define hashCode-method
   (let ([hashCode (lambda (this)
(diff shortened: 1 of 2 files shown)


rhombus/private/equatable.rkt
```diff
@@ -0,0 +1,91 @@
...
+(define (get-equatable who v)
+  (define vt (Equatable-public-ref v #false))
+  (unless vt
+    (raise-argument-error* who rhombus-realm "Equatable" v))
+  vt)
+
+(define equals-method
+  (let ([equals (lambda (this other)
+                  ((vector-ref (get-equatable 'equals this) 0) this other))])
+    print))
```
bennn, line 80: `equals` ?

jackfirth: Oops. Good catch.

jackfirth:
Question for @mflatt: I saw that the implementation of `Printable` (which I copied for this PR) uses separate `prop:Printable` and `prop:Printable-public` properties. Why is that? Am I correct in assuming I need to do that for `Equatable` too?

jackfirth:
> Will we be able to get correct `chaperone-of?` behavior without `recur`? If we use `equal-always?` rather than `recur` for comparing the elements of a collection, then `chaperone-of?` will probably return `#t` even in cases where chaperones have been stripped from the elements rather than added to them.

That sounds possible. Can you put together an example?

rocketnia:
I'll give it a try. :)

mflatt:
> Question for @mflatt: I saw that the implementation of Printable (which I copied for this PR) uses separate prop:Printable and prop:Printable-public properties. Why is that? Am I correct in assuming I need to do that for Equatable too?

That allows `Printable` to be privately implemented, so that the `print` method is called by printing primitives but can't be abused through direct calls.

I am particularly interested in being able to privately implement an interface like `Equatable` so that the `equals` part doesn't have to start by checking that `other` is an instance of the same class. But I don't think it's important for this initial implementation.


The current `chaperone-of?` implementation does rely on `equals` implementations using the recur argument.

rocketnia:
Here's a test Racket file showing the difference between the `chaperone-of?` behavior of a struct that uses `recur` and one that uses `equal-always?` in its place. It also shows how `box-immutable` aligns with the behavior of the `recur` version.

The code comment is pretty much just a recap of the context in this thread so that the example is more self-contained.

```racket
#lang racket/base


(require (only-in racket/contract/base -> any/c))
(require (only-in racket/contract/combinator make-chaperone-contract))
(require (only-in racket/contract/region define/contract))
(require (only-in racket/match define/match))
(require (only-in rackunit check-exn check-not-exn check-true))


; We're demonstrating a situation where a data structure has unideal semantics for `chaperone-of?` as
; a result of comparing its elements using `equal-always?` instead of `recur`. In particular, we'll be
; showing that it treats an instance as a chaperone of another instance even if the first instance's
; elements have *fewer* chaperones on them than the second one's do.
;
; It seems like the point of `chaperone-of?` is to check for a certain kind of bug that may arise in
; the implementation of a contract combinator. As such, for the demonstration, we'll be using a
; contract combinator that's specifically prone to this kind of misbehavior so that it can be caught
; by the check. To keep it simple, we'll just have a combinator `cheat-using-const/c` that ignores the
; value passing through and replaces it wholesale with whatever value we want.
;
; To demonstrate that the only difference is the choice between `recur` and `equal-always?`, we
; implement the same data structure twice (`recur-box` and `equalw-box`). We also check the behavior
; of `box-immutable` to show that it aligns with the `recur` behavior.


(define (cheat-using-const/c result)
  (make-chaperone-contract
    #:name `(cheat-using-const/c ,result)
    #:first-order (lambda (v) #t)
    
    #:late-neg-projection
    (lambda (blame)
      (lambda (v missing-party)
        result))))


(define e1 values)
(define/contract e2 (-> any/c any/c) e1)

(check-true (and (chaperone-of? e2 e1) (not (chaperone-of? e1 e2)))
  "The example element's contract-protected version is a proper chaperone of it")


(struct recur-box (contents)
  #:methods gen:equal-mode+hash
  [
    (define/match (equal-mode-proc a b recur now?)
      [((recur-box contents-a) (recur-box contents-b) _ _)
       (recur contents-a contents-b)])
    (define/match (hash-mode-proc v recur now?)
      [((recur-box contents) _ _)
       (recur contents)])
    ])

(struct equalw-box (contents)
  #:methods gen:equal-mode+hash
  [
    (define/match (equal-mode-proc a b recur now?)
      [((equalw-box contents-a) (equalw-box contents-b) _ _)
       (equal-always? contents-a contents-b)])
    (define/match (hash-mode-proc v recur now?)
      [((equalw-box contents) _ _)
       (equal-always-hash-code contents)])
    ])


(let ()
  (define b1 (box-immutable e1))
  (define b2 (box-immutable e2))
  (check-exn exn:fail:contract?
    (lambda ()
      (define/contract b3 (cheat-using-const/c b1) b2)
      b3)
    "A chaperone contract on an immutable `box?` does not let a merely `equal-always?` element pass"))

(let ()
  (define b1 (recur-box e1))
  (define b2 (recur-box e2))
  (check-exn exn:fail:contract?
    (lambda ()
      (define/contract b3 (cheat-using-const/c b1) b2)
      b3)
    "A chaperone contract on a `recur-box?` does not let a merely `equal-always?` element pass"))

(let ()
  (define b1 (equalw-box e1))
  (define b2 (equalw-box e2))
  (check-not-exn
    (lambda ()
      (define/contract b3 (cheat-using-const/c b1) b2)
      b3)
    "A chaperone contract on an `equalw-box?` actually does let a merely `equal-always?` element pass"))
```

jackfirth:
That seems fine to me. Contracts returning values that are merely `equal-always?` instead of `chaperone-of` seems like a reasonable thing to allow.

mflatt:
> That seems fine to me. Contracts returning values that are merely equal-always? instead of chaperone-of seems like a reasonable thing to allow.

I think the issue is not that a contract without `recur` is inherently wrong, but that a contract may intend to mean something that can only be implemented if it has `recur` available (and that also seems like the common case).

AlexKnauth:
> That seems fine to me. Contracts returning values that are merely equal-always? instead of chaperone-of seems like a reasonable thing to allow.

It allows a cheating contract like this `cheat/c` to go around other contracts and negate their chaperones:

```racket
(define ws (make-hashalw))

(define cheat/c
  (make-chaperone-contract
    #:name `cheat/c
    #:first-order (lambda (v) #t)
    #:late-neg-projection
    (lambda (blame)
      (lambda (v missing-party)
        (hash-ref! ws v (λ () v))))))
```

An example of it negating a `(box/c number?)` chaperone contract:

```racket
(define b0 (box 0))
(define/contract wb0 cheat/c (equalw-box b0))
(define/contract b1 (box/c number?) b0)
(define/contract wb1 cheat/c (equalw-box b1))

(set-box! (equalw-box-contents wb1) 'not-a-number)
(unbox (equalw-box-contents wb1))
```

This cheat passes through it only because `equalw-box` implements chaperone-of with equal-always instead of recur. Replacing `equalw-box` with `recur-box` in this example lets the contract system catch this cheating at the `(define/contract wb1 cheat/c (recur-box b1))` stage.

rfindler:
@AlexKnauth I don't quite understand the code that you've written, possibly because some `equalw` functions are missing? What are they?

AlexKnauth:
It's building on @rocketnia 's earlier comment, specifically the definitions of `equalw-box` and `recur-box` here:
```racket
(struct recur-box (contents)
  #:methods gen:equal-mode+hash
  [
    (define/match (equal-mode-proc a b recur now?)
      [((recur-box contents-a) (recur-box contents-b) _ _)
       (recur contents-a contents-b)])
    (define/match (hash-mode-proc v recur now?)
      [((recur-box contents) _ _)
       (recur (list struct:recur-box contents))])
    ])

(struct equalw-box (contents)
  #:methods gen:equal-mode+hash
  [
    (define/match (equal-mode-proc a b recur now?)
      [((equalw-box contents-a) (equalw-box contents-b) _ _)
       (equal-always? contents-a contents-b)])
    (define/match (hash-mode-proc v recur now?)
      [((equalw-box contents) _ _)
       (equal-always-hash-code (list struct:equalw-box contents))])
    ])
```

PR #266, opened 2023-01-06
quoting revisited

mflatt:
Until now, I've tried to avoid adding a general quoting (in the Lisp/Scheme/Racket) sense to Rhombus, emphasizing syntax quotes via `'....'`, instead.

The trouble with Scheme-style quoting is that creates a pressure to have reader notations for different kinds of values. Those reader notations get hardwired, and they are different then the usual constructor notation. For example, the quoted form of Racket `(hash 'a 1)` is `'#hash((a . 1))`, which no one likes to write. Even worse, these datatypes get tangled up with syntax-object representations, so anything that traverses syntax has to deal with them to some degree.

Still, I'm also finding it difficult to do without symbol and keyword literals. Before this commit, there are `symbol` and `keyword` forms that expect an immediate identifier or keyword argument to quote, respectively:

```
> symbol'apple'
symbol'apple'
> keyword'~apple'
keyword'~apple'
```

There's no particular reason that these forms need the `'`s, but the intent is to clarify that the identifier or keyword is quoted. (Parentheses are also allowed, currently, but then it looks too much like a function call, so let's ignore that.) Symbols and keywords show up enough that this is annoyingly verbose.

This commit tries out out an implementation where `#'` is parsed as an operator shrubbery (even though `'` would not normally be allowed in an operator name), and then it's defined as a prefix operator that quotes symbols and keywords:

```
> #'apple
#'apple
> #'~apple
#'~apple
```

So far, I find this convenient and readable, and it's easy to get used to the idea that the roles of `#'` and `'` have sortof swapped relative to Racket. I think I might like `#'apple'` better, but making the closing `'` work is a much larger change to shrubbery notation and the representation of shrubbery terms, so a prefix-operator `#'` notation seems like a good compromise. To generalize, each of the non-grouping special characters could be allowed as an operator when prefixed with `#`, which means adding `#,`, `#;`, `#:`, and `#|`.

A prefix `#'` is enough of an improvement that I'm less tempted to use strings for map keys:

```
{#'apple: 1, #'banana: 2}
```

I'm inclined to go further and allow `#'` to construct lists, sets, and maps in a `quote`-like way:

```
#'[apple, banana, coconut]
#'{apple: 1, banana: 2, coconut: 3}
#'{apple, banana, coconut}
```

Note that `#'{apple: 1, banana: 2}` is different from `'{apple: 1, banana: 2}'`. The latter is a syntax object has no actual hash tables in it, but it reflects code that would construct a hash table; the former would produces a hash table. In other words, Rhombus `#'{apple: 1, banana: 2}` would not be like Racket `'(hash apple 1 banana 2)`. By being less like a Lisp/Scheme/Racket quote, maybe `#'` ends up being more useful and intuitive.

The obvious next step is to say that `#'` is more like quasiquote and define `#,` as unquote.

I imagine that the datatypes supported by a predefined `#'` are tied to the set of datatypes that have predefined shorthand constructor notation. Attempting to quote a `MutableMap{apple: 1}` form under `#'`, for example, would be a syntax error. We could support "quote macros" just like any other kind of syntactic category, though, and if you want to introduce or shadow a data shorthand, you could also define a quoting rule.


jackfirth:
I'd _really_ like to avoid quoting for non-syntax objects. It's caused endless headaches in Racket and is probably responsible for a significant fraction of all questions from confused users in the Racket discord.

Are symbols and keywords the only use case for this?

mflatt:
> I'd really like to avoid quoting for non-syntax objects. It's caused endless headaches in Racket and is probably responsible for a significant fraction of all questions from confused users in the Racket discord.

I think we probably all understand that quoting has various drawbacks, which is why I've been avoiding it, too. But the text above distinguishes the proposal from Racket quoting. It's not obvious to me that the questions you have in mind would show up with this notation and in a setting where quoting is much less emphasized. I'm also not clear that the remaining questions wouldn't show up with syntax quoting, meanwhile. Examples would be welcome.

> Are symbols and keywords the only use case for this?

It's often useful to write a quick map with symbol keys, sometimes with nested maps, and it's much more convenient to quote the whole thing than individual symbols. You can write a macro each time, of course, but `quote` is a nice and general macro when properly deployed. Maybe it's possible to keep the good parts.

jackfirth:
I expect we'll get questions about why this doesn't work:

```
let x = 42
#'{apple: 1, banana: x, coconut: 3}
```

This is a frequent question asked about quoted lists in Racket. The issue doesn't show up with syntax quoting because beginners rarely construct syntax objects.

The `symbol'foo'` form is pretty wordy, so I feel your pain. This seems to be the worst when constructing maps, since maps with symbol keys are pretty useful.

What if we limited symbol quoting to map construction only? So `{apple: 1, banana: 2, coconut: 3}` would be short for `Map(symbol'apple', 1, symbol'banana', 2, symbol'coconut', 3)`.

rocketnia:
While I agree with @jackfirth that quoting for non-code is a source of confusion, I also like the "quote macros" approach. When every quotable thing has a syntax deliberately designed for that purpose, it avoids a lot of awkwardness. I think if there's any good way to have quoted data, it's what you've proposed here or something like it.

Hm, I briefly thought of asking if a sugar for switching to `#lang racket` syntax could serve the same purpose as this sugar does. However, considering that Rhombus has taken such a plunge to diverge from Racket's lexical syntax decisions, it probably wouldn't make sense to moor it to Racket's syntax just for this.


## Syntax quote macros

In my own dream language design notes, I've been considering the quote macros approach for quoting *syntax*, actually. That way, quoted syntax would be highlightable and statically analyzable like the rest of the syntax, perhaps facilitating a type system something like MetaOCaml's.

However, expanding quote macros for quoted syntax might pose a challenge semantically: When macros have optional arguments and such, it's common for some of their usage modes to expand into normalized calls to their other modes. Let's call that a staged recursive macro. If every quoted piece of syntax has to be macroexpanded, then the expansion of the definition of a staged recursive macro might have to expand quoted calls to the same macro that's being defined, before its definition is fully known. This seems hard to reconcile with Racket's phase system. I've been considering whether this should deter me enough to give up on the idea in my dream language notes, but I'm not sure yet.

Incidentally, if an acceptable semantics for this were found, each new piece of user-defined syntax would have to document all the locations where its users can `unquote` through it. This extra specification burden on macro authors (and reading comprehension burden on their users) would make it all the more necessary to have something like `syntax-parse` syntax classes to factor out the repeated design choices. Since Rhombus and Racket will have syntax classes already, they might be better equipped to explore this than a lot of other languages would be.

---

> @jackfirth: What if we limited symbol quoting to map construction only? So `{apple: 1, banana: 2, coconut: 3}` would be short for `Map(symbol'apple', 1, symbol'banana', 2, symbol'coconut', 3)`.

I like it.

Would there be a way to refer to the variable named `banana` in a key position, or would it just be a symbol now? I think Groovy and JavaScript programmers disambiguate this by putting brackets or parens around a key when it's an expression.

jackfirth:
Some sort of expression escape like `{apple: 1, [expr]: 2, coconut: 3}`, might make sense. But if there's a `Map(...)` constructor that takes expressions then I'm not sure it's worth it. People will end up writing `{["apple"]: 1, ["banana"]: 2, ["coconut"]: 3}` instead of just writing `Map("apple", 1, "banana", 2, "coconut", 3)`. That already happens today in Racket with hash literals.

mflatt:
I tried variations of implicit quoting for the key position in a map early on, and the more pieces fell into place, the better the expression option seemed to work out. So, I'm not eager to revisit that.

But I agree that confusion shown in the specific example above would likely be the same for people new to the language.

So let's not add a general quote. If no one is bothered too much by `#'` as notation restricted symbols and keywords, I'm inclined to keep that (i.e., the part actually implemented in the commit). Maybe one day we'll add a library that's not in `#lang rhombus` but provides a more general `#'` for those cases when you'd otherwise write a macro approximating it, but not have that be the usual way to write things.

rocketnia:
Sounds great!

> @mflatt: (To clarify, most of the places I've ended up using symbols or keywords were not in the key position of a map, so just having quoting there wouldn't help.)

Ah. Yeah.

gus-massa:
```
> #'apple
#'apple
> #'~apple
#'~apple
```

IIUC it is interpreted as a keyword if the name stars with a ~ and as a symbol otherwise? 

Is `~` part of the keyword name or just a part of the syntax representation? 

(I guess it's possible to make symbols with a `~` using some workaround, like `|~apple|` in Racket.)

mflatt:
> Is `~` part of the keyword name or just a part of the syntax representation?

It's the syntax and not part of the name. So, `~apple` is the same keyword as written `#:apple` in Racket S-expression notation.

PR #270, opened 2023-01-09
Add an equality macro

jackfirth:
This is a draft PR I'm leaving up as a reference. It implements a hypothetical `equality:` macro for specifying what components of a class are relevant when comparing its instances for equality. This code:

```
class Foo(a, b, c):
  equality:
    a
    b
    c
```

Expands into this, roughly:

```
class Foo(a, b, c):
  implements Equatable

  override method equals(other):
    (other is_a Foo)
      && (a == other.a)
      && (b == other.b)
      && (c == other.c)

  override method hashCode():
    let code = 1000003
    let code = 31 * code + a.hashCode()
    let code = 31 * code + b.hashCode()
    let code = 31 * code + c.hashCode()
    code
```

You can pass arbitrary expressions into `equality`, not just fields, so this works:

```
class TwoElementIntSet(a :: Integer, b :: Integer):
  equality:
    min(a, b)
    max(a, b)
```

I'm drafting this PR for reference in future discussions. I'm not sure this is the way to go, but I did want to make this code available. Consider it a proof of concept.

### DIFF
--- a/rhombus/data/equality.rhm
+++ b/rhombus/data/equality.rhm
@@ -1,0 +1,44 @@
+#lang rhombus/and_meta
+
+
+export:
+  equality
+
+
+import:
+  lib("racket/base.rkt"):
+    meta
+    expose: #{generate-temporaries}
+  lib("racket/base.rkt").#{equal-hash-code}
+  lib("racket/unsafe/ops.rkt").#{unsafe-fx+/wraparound}
+  lib("racket/unsafe/ops.rkt").#{unsafe-fx*/wraparound}
+
+
+class_clause.macro 'equality: $key_expr; ...':
+  def [key_id, ...] = #{generate-temporaries}([key_expr, ...])
+  'internal PrivateForEquals
+   implements Equatable
+
+   private method $key_id(): $key_expr
+   ...
+
+   override method equals(other):
+     (other is_a PrivateForEquals)
+       $$('&& (this . $key_id() == (other -: PrivateForEquals) . $key_id())') ...
+
+   override method hashCode():
+     let code = 1000003
+     let code:
+       #{unsafe-fx+/wraparound}(
+         #{unsafe-fx*/wraparound}(code, 31), #{equal-hash-code}(this . $key_id()))
+     ...
+     code'
+
+
+class Foo(a, b, c):
+  equality:
+    a
+    b + c
+
+Foo(1, 5, 10) == Foo(1, 10, 5) // true
+Foo(1, 5, 10) == Foo(1, 5, 5) // false


hashim-hivery:
Are the a b c on separate lines, a list? If not, would they be nicer in a list? Then they can be in one line
```
class Foo(a, b, c):
  equality: [a, b, c]
```

AlexKnauth:
```
class Foo(a, b, c):
  equality:
    a
    b
    c
```
is an equivalent tree to
```
class Foo(a, b, c):
  equality: a; b; c
```

rhombus/data/equality.rhm
```diff
@@ -0,0 +1,44 @@
...
+  def [key_id, ...] = #{generate-temporaries}([key_expr, ...])
+  'internal PrivateForEquals
+   implements Equatable
+
+   private method $key_id(): $key_expr
+   ...
+
+   override method equals(other):
+     (other is_a PrivateForEquals)
+       $$('&& (this . $key_id() == (other -: PrivateForEquals) . $key_id())') ...
```
mflatt, line 27: Ok if I change `$$` to `$&` (so this code would need to be updated)?

I think `$&` connects better to a splicing `&` in repetitions. That is, `$&` is a combination of an escape and a splice.

jackfirth: Yup, fine by me. I like `$&` better too.

Lazerbeak12345:
What are the magic numbers here doing?

AlexKnauth:
It would be good to replace the magic numbers in the hash-code method with a call to a hash-code-combine function similar to [`hash-code-combine` in `racket/src/cs/rumble/hash-code.ss`](https://github.com/racket/racket/blob/c28159300049ef2532f2e8dd62e8eba9c6b89a84/racket/src/cs/rumble/hash-code.ss#L315-L317).

jackfirth:
For the 31 constant, see [this stackoverflow post](https://stackoverflow.com/a/299748/3496273). For the 1000003 constant, see [this one](https://stackoverflow.com/questions/51072644/why-autovalue-annotation-uses-the-specific-integer-1000003-for-calculating-hash?noredirect=1&lq=1). I just copied what `@AutoValue` does, since it seemed fine in practice. I'm guessing the reason for picking a large starting constant is to avoid hash codes for small record types being biased to small-ish positive integers.

sorawee:
Effective Java uses 1000003 in the `AutoValue` section _without using 31_. On the other hand, the `hashCode` recipe section uses 31 _without using 1000003_. Using both 1000003 and 31 is probably fine, but it's unclear why it needs to be like that.

I agree with @AlexKnauth that exposing and calling `hash-code-combine` would be better.

Lazerbeak12345 (commented):
I looked into as many edge cases that came to mind without actually running this, and aside from recommendation to make use of a pre-existing hash method (preventing code duplication and reducing magic number use), I can't see any other issues.

We need to be sure to document that the `other` value is first getting passed through `(_ -: PrivateForEquals)` before getting the properties, and everything that entails. But that's for once we actually start writing docs. Unsure if docs would belong in this pr.

AlexKnauth:
For `hash-code-combine`, I've made a PR to racket here: https://github.com/racket/racket/pull/4546

rhombus/data/equality.rhm
mflatt: And now it's just `$`. Unlike S-expressions, a group or a list of terms doesn't correspond to a term, so `$` can just infer splicing.

jackfirth: Oh nice, that's even better

AlexKnauth:
Some updates for recur and hash-code-combine: https://github.com/racket/rhombus-prototype/pull/274

mflatt:

PR #272, opened 2023-02-02
Add recur argument to Equatable methods

AlexKnauth:
Adds a `recur` argument to the `Equatable` methods, and also changes the method names to reflect that. The `Equatable` interface is intended to be privately implemented, so this also removes the `Equatable` namespace and its "public" method access points, as well as the `Equatable` annotation since a class that privately implements `Equatable` won't match the annotation anyway.

Using the `recur` argument allows `Equatable` to specify correct `chaperone-of?` behavior, as well as other benefits like cycle-detection and other things enabled by `equal?/recur` and `equal-always?/recur`, such as looser equality for inexact numbers for `check-within` testing purposes.

See also
 - https://github.com/racket/rhombus-prototype/pull/265#issuecomment-1373079389
 - `cheat/c` from https://github.com/racket/rhombus-prototype/pull/265#issuecomment-1374000621
 - Racket discord, rhombus channel, 2023-01-04: https://discord.com/channels/571040468092321801/871953739982995547/1060369305314807929
 - Racket discord, rhombus equality channel, equal-within thread, 2023-02-03: https://discord.com/channels/571040468092321801/1071150855123914782/1071156782761783416

### DIFF
--- a/rhombus/private/equatable.rkt
+++ b/rhombus/private/equatable.rkt
@@ -9,9 +9,7 @@
          "class-dot.rkt"
          (only-in "class-desc.rkt" define-class-desc-syntax))
 
-(provide (for-spaces (rhombus/namespace
-                      rhombus/class
-                      rhombus/annot)
+(provide (for-spaces (rhombus/class)
                      Equatable))
 
 (define-values (prop:Equatable Equatable? Equatable-ref)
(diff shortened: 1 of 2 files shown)


mflatt:
This seems generally ok with me, but how about sticking with just `equals` and `hash_code` instead of `equal_recur` and `hash_code_recur`?

PR #273, opened 2023-02-05
disallow `:` at end of an operator and change `-:` to `:~`

mflatt:
Every once in a while, I want to write `<operator>:`, but instead I end up with just an operator that ends with `:`. Adding a space is ugly, and I tend to start out confused for a few seconds.

This problem is now worse with

```
operator <operator> : 
  <option>; ...
  match 
  | <case> 
  | ...
``` 

as a way to lift options that should span all operator cases out of an individual case, as motivated by improving an example we looked at in the February 2 meeting:

```
operator ! :
  ~stronger_than: ~other
  match
  | (0!): 1
  | (n!): n * (n-1)!
```

It seems less confusing to disallow `:` at the end of an operator name, unless the operator name is all `:`s. This is the same restriction as for `+`, `-`, and `.`. Then, the `! :` above could be written `!:`.

This new rule would disallow `-:`, which is currently the variant of `::` that associates static information without implying a run-time check. This commit renames to `:~` (which became available when we decided to allow `~` in operator names).


mflatt:
Merged as 47bcbe17606d507dc700a3a4b09d00ea11833f7b.

PR #274, opened 2023-02-06
Update equality macro for recur and hash-code-combine

AlexKnauth:
Changes from @jackfirth's original #270:
 - [x] Rhombus unsyntax-splicing `$$` is now just `$`
 - [x] Rhombus assume-annotation `-:` is now `:~`
 - [x] Equatable uses recur now
 - [x] `private implements Equatable` is encouraged
 - [x] Uses Rhombus's new `hash_code_combine`
 - [x] Uses type tags in hash codes to distinguish different but similarly-shaped types
 - [x] Adds tests


mflatt:

PR #275, opened 2023-02-07
Rest argument hash_code_combine & unordered

AlexKnauth:
Changes `hash_code_combine` and `hash_code_combine_unordered` to use a rest argument list instead of a single-argument list. This also lets `hash_code_combine` on one argument mix a single hash code just like `hash-code-combine`.


mflatt:
Thanks!

PR #280, opened 2023-03-07
For each parens

AlexKnauth:
Allows `each` clauses in `for` macros to use parentheses like
```
each (x: l)
```
Which allows some `for` forms to be written on one line more easily:
```
for List: each (x: l); f(x)
```
Without needing to use special characters like `«»`.

This also allows multiple sequences:
```
each (x: xs, y: ys)
```
Equivalent to
```
each:
  x: xs
  y: ys
```

### DIFF
--- a/rhombus/private/for.rkt
+++ b/rhombus/private/for.rkt
@@ -110,8 +110,9 @@
                                                          #`((#,group-tag any ...))
                                                          #'rhs-blk)
                           . bodys))]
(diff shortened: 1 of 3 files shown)


jackfirth:
I'm not sure writing `for` forms in one line is a good idea in the first place.

AlexKnauth:
I don't want to be forced to write it on multiple lines just for it to parse correctly.

sorawee:
#349 is merged, so I think this is resolved?

PR #291, opened 2023-04-20
Change `to_string` to convert mutable strings to immutable.

samth:
Since `String` only corresponds to immutable strings in Rhombus,
`to_string` should always produce them.

### DIFF
--- a/rhombus/private/string.rkt
+++ b/rhombus/private/string.rkt
@@ -56,7 +56,6 @@
 (define/arity (to_string a)
   #:static-infos ((#%call-result #,string-static-infos))
   (cond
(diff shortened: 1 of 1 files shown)


rhombus/private/string.rkt
```diff
@@ -56,7 +56,8 @@
 (define/arity (to_string a)
   #:static-infos ((#%call-result #,string-static-infos))
   (cond
-    [(string? a) a]
+    [(immutable-string? a) a]
```
jackfirth, line 59: `string->immutable-string` already has this optimization, there's no need to do it here too.

PR #299, opened 2023-04-28
block-alts design proposal

AlexKnauth:
A design proposal for an addition to Shrubbery to allow `|` alts after the end of a `:` block.

Design document: [design/block-alts/block-alts.md](https://github.com/AlexKnauth/rhombus-prototype/blob/block-alts/design/block-alts/block-alts.md)

An implementation of option A is at: https://github.com/AlexKnauth/rhombus-prototype/tree/block-alts-impl-A

An implementation of option C is at: https://github.com/mflatt/rhombus-prototype/tree/alts

Option C is the one we've decided to try for now.

### DIFF
--- a/demo.rhm
+++ b/demo.rhm
@@ -128,8 +128,8 @@ also_first_one([1, 2, 3, 4])
 
 fun add1(x) :: Int:
   match x
-   | n :: Int : x + 1
-   | ~else: x
+  | n :: Int : x + 1
+  | ~else: x
 
 add1(100)
 // would trigger an error, since "oops" is not an `Int`:
--- a/design/block-alts/block-alts.md
+++ b/design/block-alts/block-alts.md
@@ -397,6 +397,37 @@ syntax_parse stx_expr:
 
 Instead of needing some identifier like `pattern` or `match` separating the options from the cases.
 
+### Option C: `|` pipe indentation equal to group head indentation
+
+For a shrubbery like:
+
+```
+a:
+  b
+| c
+```
+
+To parse as:
+
+```
+(group a
+       (block (group b))
+       (alts (block (group c))))
+```
+
+The indentation of the `|` before `c` must be equal-to the indentation of `a`, less than the indentation of `b`. If it's less than `a`, then it might belong to a parent of `a` instead of `a`'s group. If it's between `a` and `b`, then it's an indentation error. If it's equal to `b`, then it would belong to `b`'s group instead of `a`'s group, and if it's more than `b` it's an indentation error.
+
+This indentation restriction applies to normal alts as well, so:
+
+```
+a
+  | c
+```
+
+is an indentation error too.
+
+The grammar, written/armored forms, and possible modifications to Rhombus forms are the same as written in option A.
+
 Evaluation and Tradeoffs
 --------------
 
@@ -486,6 +517,49 @@ Might "silently" fail, with the parser interpreting the same text as a different
 Instead of giving an error or some other immediate feedback to prompt users to fix their code.
 
 Rightward drift may also become a problem with option B, when programs get larger, more complicated, and more nested, the little bits of extra indentation that option B forces on `|` alts may start to add up.
+
+### Good of C
+
+Making `|` alts indentation between `a` and `b` an error prevents the problem in Bad of A. If
+
+```
+a:
+  b
+ | c
+```
+
+Is an indentation error, then there's no confusion between that and 
+
+```
+a:
+  b
+  | c
+```
+
+Option C also has less "rightward drift" than option B.
+
+### Bad of C
+
+If there are Rhombus programs that use rightward drift for alts:
+
+```
+a
+  | c
+```
+
+instead of
+
+```
+a
+| c
+```
+
+then they will get indentation errors.
+Fortunately, those errors could be made easy to read,
+and any given program should be simple to correct.
+Simple enough that an automatic tool similar to resyntax
+could probably do it.
+But it will still take more than zero effort.
 
 Prior art and References
 ---------
--- a/design/block-alts/block-alts.md
+++ b/design/block-alts/block-alts.md
@@ -1,0 +1,283 @@
+block-alts
+==========================
+
+Summary
+-------
+
+An addition to Shrubbery to allow `|` alts after the end of a `:` block.
+
+Motivation
+----------
+
+ * Syntax class options, including kind and fields, without `pattern` in between
+ * Operator options, including precedence, without `match` in between
+ * Macro options, including precedence and op_stx, without `match` in between
+ * Function cases with a common result annotation, without `match` in between
+ * Using shrubbery syntax for a whitespace-lisp
+
+Design Options
+--------------
+
+### Option A
+
+Describe a complete approach to the design, introducing its components
+and how they might interact.
+
+#### Parsing shrubberies
+
+```
+‹group›	::= ‹term›* ‹tail›
+
+‹tail› 	::= ‹term›
+       	 |  ‹block›
+       	 |  ‹alts›
+       	 |  ‹block› ‹alts›
+
+‹block›	::= `:` ‹group›*
+
+‹alts› 	::= `|` ‹block›*
+```
+
+```
+_group	= (group _term ... _tail)
+
+_tail 	= _term
+      	| _block
+      	| _alts
+      	| _block _alts
+
+_block	= (block _group ...)
+
+_alts 	= (alts _block ...)
+```
+
+For a shrubbery like:
+
+```
+a:
+  b
+| c
+```
+
+To parse as:
+
+```
+(group a
+       (block (group b))
+       (alts (block (group c))))
+```
+
+The indentation of the `|` before `c` must be greater-than-or-equal-to the indentation of `a`, and less than the indentation of `b`. If it's less than `a`, then it might belong to a parent of `a` instead of `a`'s group. If it's at `b` or more, then it would belong to `b`'s group instead of `a`'s group.
+
+#### Parsing written/armored forms of shrubberies
+
+The current code in `rhombus-prototype/shrubbery/write.rkt` is already good.
+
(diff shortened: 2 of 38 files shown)


mflatt:
I appreciate the problem that this change is trying to solve, but I find

```
main_id:
  option1
  option2
| case1
| case2
```

difficult to read. The `main_id` part feels like it ends with `option2`. More generally, having two levels of indentation associated with a blocking form seems complex. The contrast with

```
main_id:
  option1
  option2
  | case1
  | case2
```

is especially worrying to me, where the alternatives end up counting as being under `option2`.

If there's appetite for changing the rules of `|` more generally, I would be in favor of (1) removing the implicit "0.5" column that's associated with `|`, which means that `|` needs to be more indented than the start of its enclosing group, and (2) having the block and alternative parts in this PR's extension be at the same indentation. As a result, the second example above would be the way to write `main_id` with two options and two cases, and the first one above would be ill-formed.

To put it another way, things that belong to `main_id` would start at the same level of indentation. It also seems relatively easy to explain that a block can end with any number of `|`s, and that `|`s can appear only at the end.

Nesting `case1` and `case2` further would then require more indentation:

```
main_id:
  option1
  nested_id
    | case1
    | case2
```

The revision I'm suggesting would also mean that `match` would be written like

```
match x
  | pattern1: result2
  | pattern1: result2
```

I would further be in favor of allowing `:` before `|`, so that 

```
match x:
  | pattern1: result2
  | pattern1: result2
```

would be allowed.

AlexKnauth:
I've added option B to the design document for parsing
```
main_id:
  option1
  option2
  | case1
  | case2
```
as
```
(group main_id
       (block (group option1)
              (group option2))
       (alts (block (group case1))
             (block (group case2))))
```

And I put some of my thoughts on it in the evaluation and tradeoffs section.

Overall I think I still prefer option A though.
```

**Comment by sorawee** [comment]
The third option, which I have been advocating, is option A, but `:` is _mandatory_ (and not simply _allowed_). 

The problem that Matthew worries about is fixed under this option.

```
main_id:
  option1
  option2
  | case1
  | case2
```

is a parse error, because if the alternatives are to be placed as under `option2`, there must be `:` after `option2`.

Furthermore, it provides a consistent syntax w.r.t. presence of options.

```
main_id:
  option1
  option2
| case1
| case2
```

and 

```
main_id:
| case1
| case2
```

both have `:`. When we delete all options, we don't need to remove `:`. When we add options (where previously there were none), we don't need to add `:`.

Here are a couple of scenarios with option A + `:` mandate


```
a:
  b
| c
```

is read as:

```
(group a
       (block (group b))
       (alts (block (group c))))
```

----------------


```
a:
| c
```

is read as:

```
(group a
       (alts (block (group c))))
```

----------------

```
a:
  b
  | c
```

is a parse error. There must be `:` after `b`, or else `|` should be unindented.

----------------

```
a:
  b:
  | c
```

is read as:

```
(group a 
       (block (group b (alts (block (group c))))))
```

----------------

```
a:
  b:
| c
```

is a parse error because an empty block is not allowed.

**Comment by AlexKnauth** [comment]
If `:` is mandatory, then typing
```
a:
| b
```
will be a pain, since after typing `a:`, it will insert indentation making
```
a:
  | b
```
instead. This is why allowing `:` directly before `|` makes more sense for option B than it does for option A.

**Comment by sorawee** [comment]
Not necessary. Pyret, for example, have something along that line that I proposed. Consider:

```
fun test():%
```

where `%` is the caret position.

After entering a newline, I get:

```
fun test():
  %
```

If I type `v`, I will get:

```
fun test():
  v%
```

On the other hand, if I type `|`, I will get:

```
fun test():
|%
```

That is, typing `|` automatically unindents.

OCaml is a bit different, but the result is similar. Consider:

```
let _ =
  match a with
    %b
```

This code is in the "normal form" for indentation. Re-indenting the whole program does not change anything.

If I type `|`, I get:

```
let _ =
  match a with
  |%b
```

That is to say, `|` automatically unindents.

Pyret is a bit more faithful to what I proposed, since `:` will by default add indentation. `match` in OCaml on the other hand does not add any indentation, but it knows that non `|` should be further indented, so the final results are the same.

**Comment by sorawee** [comment]
And in general, you can use tab to toggle levels anyway.

For example, in:

```
match x:
| y:
    match z:
    | a: 1
    | b: 2%
```

Tabbing would alternate between the above code and

```
match x:
| y:
    match z:
    | a: 1
| b: 2
```

Note that currently, there is an extra possible position.

```
match x:
| y:
    match z:
    | a: 1
         | b: 2
```

But this position is invalid in my proposal because there's no `:` after `1`. So mandating `:` also allows us to remove these spurious positions.

**Comment by AlexKnauth** [comment]
Making `:` mandatory doesn't completely solve the problem in "Bad of A," it just shifts it to slightly more complicated cases like
```
a:
  b: c
  | d
```
vs.
```
a:
  b: c
 | d
```

**Comment by sorawee** [comment]
I see nothing wrong with the two programs that you presented. Both are valid, and rightfully so. Same situation as:

```
match x:
| y:
    match z:
    | a: 1
    | b: 2
```

and

```
match x:
| y:
    match z:
    | a: 1
| b: 2
```

**Comment by gus-massa** [comment]
Is the idea to add `else`  clauses like in Python?

```
#lang python
def sqrt2(n):
  for i in range(10):
    if i * i == n:
      print(i)
      break
  else:
    print("Not found")


sqrt2(9)  # ==> 3

sqrt2(8)  # ==> Not found
```

My problem is that I never liked them in Python, so I'm biased. I don't like that the main flow includes a `break`.

Also, I think it's difficult to decide how to indent the `|`. In spite of the official reason to create rhombus, my personal reason is that I think that it's impossible to indent/format Racket unless you add a million of rules like DrRacket does. (And another million of rules for brackets.)  So my wish is that rhombus has foolproof indentation rules that are human friendly and machine friendly.

**Comment by AlexKnauth** [comment]
With this proposal there's nothing *syntactically* preventing
```
fun sqrt2(n):
  for:
    each i: range(10)
    when i * i == n
    | print(i)
      break
  | else:
      print("Not found")
```
But instead a cultural aversion, limiting control flow like this to dedicated forms like a `try` form, which, if desired, this would allow `try` to look like:
```
try:
  body
  ...
| handler_clause
| ...
```

**Comment by gus-massa** [comment]
Using the `|` with `try` looks like a good idea.

Can the first part get a `|` too, like in my new example? (It looks bad for me.) Can the macro forbids it?

```
try:
| body              // <-- here
  ...
| handler_clause
| ...
```

**Comment by sorawee** [comment]
They are read differently, so yes, the `try` macro can recognize one and forbid the other.

```
try:
  body1
  body2
| handler_clause1
| handler_clause2
```

is read as:

```
(group
 try
 (block (group body1) (group body2))
 (alts (block (group handler_clause1)) (block (group handler_clause2))))
```

But 

```
try: // assuming : is allowed/mandatory here
| body1
  body2
| handler_clause1
| handler_clause2
```

is read as:

```
(group
 try
 (alts
  (block (group body1) (group body2))
  (block (group handler_clause1))
  (block (group handler_clause2))))
```

**Comment by sorawee** [comment]
Re @mflatt's suggestion.

Compare:

```
(syntax-parse v
  [(_ a)
   (generate a)])
   ;; 3
```

with

```
syntax_parse v:
| '_ $a':
    'generate($a)'
    // 4
```

and

```
syntax_parse v:
  | '_ $a':
      'generate($a)'
      // 6
```


Currently, we gave up a space relative to Racket's counterpart. That doesn't seem ideal to me already. So I am really averse to sacrificing 2 more spaces.

**Comment by mflatt** [comment]
Another variant of A to consider: `|` is not allowed to be more indented (or must have indentation 0.5, if you think of `|` has having a 0.5 indentation). In this variant,

```
a
  | c
```

is not valid, so there's less possible confusion with

```
a:
  b
  | c
```

Grouping the alternative with `a` would be written as

```
a:
  b
| c
```

We could allow `:` before `|`, but I don't think there would be a need to require it.

With this approach, existing shrubbery forms that are indented the standard way (i.e., no extra spaces before `|`) would continue to be well-formed.

AlexKnauth:
I added that as option C. I personally like both option A and option C equally.

sorawee:
I like the "| is not allowed to be more indented" constraint (which I implicitly always follow for my above proposed variant as well). 

The only unresolved topic, to me, is whether `:` should be (1) always required or (2) optional or (3) absent when not needed.

The advantage of (1) is consistency of `:` w.r.t. presence of options. The disadvantage is that the IDE support needs to handle the indentation of `|` after `:`, but I don't really see any disadvantage for language users.

(2) has the same disadvantage as (1), but it does not provide the consistency, which could be annoying for language users.

(3) is what the current PR proposes. It doesn't need the extra IDE support, but does not provide the consistency.

mflatt:
This PR now includes a full implementation and revised documentation. Small changes were needed to fix some existing examples and tests where there was leftover extra indentation for `|`s.

The revised parsing rules seem no more difficult to explain in the documentation than the old rules. I think the shrubbery documentation has improved by shifting terminology and presentation in a way that works with `:` plus `|`.

Changes at the Rhombus level for `:` plus `|`:

* The `operator` and `macro` forms (and related forms like `expr.macro`) take advantage of the support for combining `:` and `|`.

* The syntax of `fun` is simplified. Since there are no function options, so far, the revised `fun` doesn't actually need a combination of `:` and `|`, but it now uses `|` in a way that is consistent with `operator`. 

*  The `syntax_class` form has not changed, yet. It could make sense to drop the `pattern` clause form, but unlike the `match` hack formerly in `operator` and `macro`, `pattern` is specifically bound as a `syntax_class` clause form, and it would make sense for newly defined forms to expand to `pattern`. But it would make sense to support the shorthand of omitting `pattern`, even when other clauses are present, and I've left that to a future change.

Based on the discussion at the May 11 Rhombus meeting, I think we're ready to merge this change and see how it works with more programs.

PR #301, opened 2023-05-03
add converter annotations

mflatt:
When a converter annotation `annot` is applied in an expression like `expr :: annot`, then instead of just returning the value of `expr`, the annotation can return a value derived from the value of `expr`. Similarly, with `bind :: annot` as a binding, a candidate value for binding is first converted by `annot` before it is matched to `bind`.

As an example, the new `StringViewAsString` annotation accepts both immutable and mutable strings like `StringView`, but it converts a mutable string to a mutable one. This annotation could be handy when wrapping Racket functions, for example, since declaring a `StringViewAsString` annotation as the wrapper's result will cause Racket strings (too often mutable) to get converted to immutable strings.

The underlying protocol for converter annotations piggy-backs on the binding protocol, since the issues of matching are pretty much the same. When using a converting annotation, it might make sense to delay some work until a set of matches is confirmed (e.g., other arguments to a function are found to match). Or it might be that checking does much of the work of conversion anyway, and so a result from checking should be saved to use if matching succeeds overall. Those phases correspond to the `matcher` and `committer`+`binder` split in the binding protocol. So we reuse the binding protocol inside the annotation protocol.

The primitive annotation form to make a coverter annotation is `converting`. A converter needs three pieces: a binding pattern, a conversion body, and an optional result annotation; the syntax is currently

```
  converting(fun ($bind) $maybe_ret_ann: $body; ...)
```

That's a literal `fun` (or a name bound to the same thing), which only serves as a reminder of how the rest of the syntax goes, since a converter needs the same three pieces as a function. The `maybe_ret_ann` part is the same as for `fun`, but constrained to a single-value result pattern.

Predicate annotations can be automatically promoted to converter annotations, but a distinction is still sometimes needed. For example, non-final methods can only have predicate result declarations (otherwise there would be a ordering issue with multiple inheritance via interfaces, and probably other problems). Similarly, only predicate annotations can be used with `:~`, since conversions are not optional. When an annotation constructor wants to work with input annotations, it can choose to either recognize and specialize the case of predicate annotations or to implicitly promote by always using `annot_meta.unpack_converter`.

Annotation constructors in `rhombus` generally accept both predicate and convert annotations and create a predicate annotation when all given annotations are predicate annotations. For example, `List.of(String)` is a predicate annotation, while `List.of(StringViewAsString)` is a converter annotation that maps `StringViewAsString` over a list. The `&&` annotation constructor uses its left-hand argument in match-only mode (skipping conversion work, if any can be skipped) and returns whatever conversion the right-hand argument performs. The `||` annotation constructor tries in order, so it can return a converted value from either the left-hand or right-hand annotation.


jackfirth:
The `&&` operator's behavior seems questionable to me. My feeling is that it should either not allow the left hand side to be a converter annotation at all (to avoid silently not converting inputs) or it should apply both converters one after the other. I lean towards the former, but I can imagine uses for the latter. For instance, imagine an `PrintableAsString` annotation that first converts the value to a string using `print` and `with-output-to-string` and then wants to ensure that string is immutable with `StringViewAsString`. It could use `converting(...) && StringViewAsString` to do that.

mflatt:
That change to pipe conversions through `&&` makes sense to me. It's possible to compose converters currently by nesting `::`, but the order is unusual and counterintuitive (something like contravariance).

PR #303, opened 2023-05-03
rhombus/draw: `Point`, `Size`, and `Rect`

mflatt:
... and `PointLike`, etc. Drawing methods take `PointLike` values. See the updated "gui_demo.rhm" for examples in the style of using `PointLike` lists instead of explicit constructions of `Point`, etc.

A `PointLike` is either:

 * a `Point`

   Has fields `x` and `y`, which are `Real`s and supplied by position to the `Point` constructor. (The convention and ordering `x` then `y` is so established that keyword arguments seems unneecssary and possibly counterproductive.)

 * a list containing two `Real`s

   The lightweight `[]` syntax for lists makes this about as much to type as the `racket/draw`-style plain seqeuence of number arguments, as in `p.curve_to([20, -10], [80, -10], [100, 0])`.

 * a map with (at least) `#'x` and `#'y` mapped to `Real`s

   The intent here is to support Clojure-like data patterns, which is convenient and flexible for some kinds of applications.

The `PointLikeAsPoint` annotation accepts anything that `PointLike` accepts, but it converts to a `Point`. That conversion makes the glue methods much easier to write. The conversation is also an internal detail, and when the wrong kind of value is supplied to a method, the error message should be about `PointLike`, not `PointLikeAsPoint`. To achieve that goal, the implementation says `PointLike`, but in a context that imports `PointLikeAsPoint` with the name `PointLike`. So, the conversion happens as needed, but error messages (which are based on the source text of the annotation) say "PointLike". This is a little tricky, so there's room for confusion, but it works so nicely that it may be a good general pattern.

A `Size` has fields `width` and `height` (by position) that are `NonnegReal`s.

A `Rect` has fields `x`, `y`, `width`, and `height`, but also a `point` property that returns `Point(x, y)` and a `size` property that returns `Size(width, height)`. While most names here are meant to immitate NextStep (`NSPoint`, `NSSize`, and `NSRect`), I have always found `NSRect`'s name `origin` for the point to be weird and not in parallel to the `size` field. The name `topleft` would be more descriptive of what the point means for the rectangle, but `point` still seems more consistent and easier to remember. Also, I always find it tedious to construct a separate `NSSize` and `NSPoint` to create an `NSRect`, so that's why `Rect` supplies all four parts directly, and its constructor is overloaded with a 4-argument and 2-argument version. Similarly, a `RectLike` can be a list of four numbers, a list of a `PointLike` and `SizeLike`, etc. A case could be made that the `point` then `size` order is less obvious that `x` then `y` or `width` then `height`, and so keyword arguments would make sense, but a consistent order plays better with `RectLike` constructions as easy-to-write lists.

The `RadialGradient` and `LinearGradient` constructors take combinations of `PointLike` and `NonegReal` or `Real.in(0.0, 1.0)` and `Color`, still expressed as plain lists. They seem too niche to bother whole additional classes and converters.

Probably `Point` and similar classes will get new methods in the future.

### DIFF
--- a/gui_demo.rhm
+++ b/gui_demo.rhm
@@ -6,9 +6,9 @@ import:
 // ----------------------------------------
 
 fun draw_face(dc :: draw.DC, config :: Map):
-  let [w, h] = dc.size
+  let draw.Size(w, h) = dc.size
   let s = math.max(10, math.min(w, h) - 20)
-  
+
   let scale = config["scale"] / 50
   let w = w / scale
   let h = h / scale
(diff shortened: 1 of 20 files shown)


sorawee:
A bit off-topic, but I would propose changing `NonnegReal` to `Real.nonNeg`.

mflatt:
> A bit off-topic, but I would propose changing `NonnegReal` to `Real.nonNeg`.

I can see why that might make sense. But in #304, and given that `NonnegReal` would make sense as an interface, I've so far left this as `NonnegReal`.

PR #305, opened 2023-05-09
add `Sequence`, `Sequenceable`, and `sequence`

mflatt:
The `Sequence` annotation is satisfied by any sequence, which coincides with Racket's `sequence?`. The `Sequence.make` function is a keywordized and more direct (removes a thunk layer) variant of Racket's `make-do-sequence`, while the combination of `Sequence.instantiable` and `Sequence.instantiate` supports the same instantiation indirection as `make-do-sequence`.

Implementing the `Sequenceable` interface requires supplying a `to_sequence` method that returns a sequence, possibly created by `Sequence.make`.

The `sequence` class-clause form specializes `each` expansion in `for` for a right-hand side that has the class-as-annotation's static information (and ditto for interfaces). The `sequence` form is like `define-sequence-syntax`, except that it essentially forces a `:do-in` (or `#false`) result and doesn't expose a new identifier that might be used as an expression. The specialization's result is translated to `:do-in`.


rhombus/scribblings/ref-sequence.scrbl
```diff
@@ -0,0 +1,218 @@
...
+  @rhombus(continue_at_pos_val) needs to be @rhombus(#false) (otherwise
+  the value must be retained for that predicate).}
+
+ @item{@rhombus(continue_at_pos_val): An optional function that takes
+  both the current position and elements (which may me multiple values
+  supplied as multiple arguments) and reports whether iteration should
+  continue with that element. A @rhombus(#false) for
+  @rhombus(continue_at_pos_val) is equivalent to a function that always
+  returns @rhombus(#true). This function is applied after the element is
+  exposed to iteration. The position is receives has been converted by
```
sorawee, line 108: I can't parse this last sentence.

sorawee:
Can the `sequence` form also get a "keywordized" treatment, just like how `Sequence.make` is a "keywordized" treatment of `make-do-sequence`?

rhombus/scribblings/ref-sequence.scrbl
```diff
@@ -0,0 +1,218 @@
...
+  to an iteration, and it is called before @rhombus(continue_at_value). It
+  is intended for cases where retaining the position after extracting an
+  element would incorrectly retain the element via the position.
+  Typically, if @rhombus(early_next_pos) advances a position, then
+  @rhombus(next_pos) will return its argument, and
+  @rhombus(continue_at_pos_val) needs to be @rhombus(#false) (otherwise
+  the value must be retained for that predicate).}
+
+ @item{@rhombus(continue_at_pos_val): An optional function that takes
+  both the current position and elements (which may me multiple values
```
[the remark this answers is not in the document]
mflatt: will fix

rhombus/scribblings/ref-sequence.scrbl
mflatt: will fix - thanks!

mflatt:
Yes, using keywords for `sequence` is better, and I'll make that change.

PR #313, opened 2023-05-29
change "ref" to "index"

mflatt:
Use "indexable" more generally, instead of "map", for anything that supports indexing with `[...]`. As an annotation, `Indexable` now matches all indexable values, inlcuding maps and lists, and not just instances of classes that implement the `Indexable` interface.

Renames:

```
 #%ref                         =>   #%index
 Refable                       =>   Indexable
 Map.ref                       =>   Map.at
 statinfo_meta.map_ref_key     =>   statinfo_meta.index_key
 statinfo_meta.map_set_key     =>   statinfo_meta.index_assign_key
 statinfo_meta.map_append_key  =>   statinfo_meta.append_key
 Indexable.ref                 =>   Indexable.at
 Indexable.set                 =>   Indexable.assign_at
```

For now, `Map.ref`, `statinfo_meta.map_ref_key`, and `statinfo_meta.map_append_key` are still defined for backward compatibility, but they will go away.


Gopiandcode:
What's the motivation for:
```
Map.ref                       =>   Map.at
```
Is that a deliberate choice to diverge from Racket? I've never seen `at` used to represent retrieving from a mapping

mflatt:
In choosing `at`, I may have been trying too hard to avoid `ref`. Also, I looked for alternatives to `set` (again departing from Racket) to avoid a terminology overlap with the `Set` datatype.

I think we want the name for `ref` to be the same in `Map` and `Indexable`, but I don't have a strong opinion about what the name should be.

Python uses `__getitem__` and `__setitem__`. Are there other existing conventions to consider? Are `ref` and `set` already best?

Gopiandcode:
Maybe `.ref` would be good enough, or `.get` or maybe `.find`?
Or, maybe, to be consistent with the interface, it should be `.index(...)`? That seems to fit nicely

mflatt:
I like `get` and `set`.

mflatt:
Th commit here now uses `get` and `set`.

samdphillips:
> What's the motivation for:
> 
> ```
> Map.ref                       =>   Map.at
> ```
> 
> Is that a deliberate choice to diverge from Racket? I've never seen `at` used to represent retrieving from a mapping

Smalltalk uses `at:` in this way.

Gopiandcode:
Oh, I see, then maybe it's not too bad ~ I've never used smalltalk, so the convention was foreign to me.

mflatt:
I'm happy to know there's a precedent for `at`, but `get` and `set` still seem like the most obvious choices, so let's go with that — for now, at least.

PR #319, opened 2023-06-10
add docs for ways to type guillemets

AlexKnauth:
Relies on https://github.com/racket/gui/pull/297 and https://github.com/racket/scribble/pull/367

The DrRacket way works only with the above PRs.

The Mac way and the Windows way work on my Mac and Windows machines.

I do not have a keyboard with AltGr on it, so I don't think I can test whether AltGr + `[` or AltGr + `]` work on Windows, or whether AltGr + `Z` or AltGr + `X` work on Linux either.

Edit:

I have tested the GNU/Linux, BSD way on a Linux computer, and found that I did not need to press Enter, it entered `«` as soon as I finished typing `00AB`, and entered `»` as soon as I finished typing `00BB`. Pressing Enter after that behaved as normal Enter, not as "finish hex code sequence".

On that Linux computer I also tested the Compose key versions found in https://en.wikipedia.org/wiki/Guillemet#Keyboard_entry. Although I had to go into the keyboard settings to set the Compose key to something, it worked as expected after that.

### DIFF
--- a/shrubbery/scribblings/example.scrbl
+++ b/shrubbery/scribblings/example.scrbl
@@ -66,7 +66,7 @@ way, too, and it's intended for quoting shrubbery terms.
 A @litchar{;} or @litchar{,} acts as a group
 separator, even within a single line. A @litchar{:} or @litchar{|}
 treats remaining item on the same line like a new indented line, which
-forms a block of nested groups. A guillemot pair @litchar{«} and @litchar{»} can be
+forms a block of nested groups. A guillemet pair @litchar{«} and @litchar{»} can be
 used in rare cases to explicitly bracket subgroups formed by
 @litchar{:} and @litchar{|} without line breaks. A @litchar{\} continues
 a line, effectively shifting all columns on the next line as if they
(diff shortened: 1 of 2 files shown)


AlexKnauth:
After testing on a Linux computer, I have updated the GNU/Linux, BSD entry to remove pressing Enter, and added a new entry for the Compose key.

benknoble:
In Vim it's Control-K then << or >>. Should we include other editors?

mflatt:
I agree that it makes sense to add other editors as we get information on them.

One reason I haven't merged this, though, is that the placement doesn't seem right. I think it might make sense to have a new "Tools"  section at the end of the shrubbery manual, plus a margin note or other forward reference near the start of the `«»` section. I'm trying that out, and I'll report back.

samdphillips:
> In Vim it's Control-K then << or >>. Should we include other editors?

Same in Evil mode.

mflatt:
Here's a variant with this information moved to an "Editor Support" section (and edited in an attempt to make things clearer and fit some existing key-combination conventions): https://github.com/mflatt/rhombus-prototype/tree/editor

PR #320, opened 2023-06-12
add `recur` as a named-let form

mflatt:
Differences from Scheme named `let`:

 * Initial-value expressions can refer to earlier argument variables, which makes things more consistent with default-value expressions for `fun` arguments.

 * Keyword arguments are allowed, and they are optional in recursive calls. When not supplied, each keyword argument defaults to its value from the enclosing call, which means that keyword-argument values are automatically propagated in recursive calls.

Example from the docs:

```
  > recur sum_halves(l = [1, 2, 3, -4, 5, -6],
                     ~pos_sum: pos_sum = 0,
                     ~neg_sum: neg_sum = 0):
      match l
      | []: values(pos_sum, neg_sum)
      | [n, & ns]:
          if n > 0
          | sum_halves(ns, ~pos_sum: pos_sum + n)
          | sum_halves(ns, ~neg_sum: neg_sum + n)
  11
  -10
```

A further tweak on keyword arguments might include an expression or update function that is applied to the current iteration's argument for the next iteration. In the interest of keeping things simpler, this commit doesn't do that.

Besides `recur`, the commit adds a few things that are need to implement `recur` in Rhombus, such as the `bind_meta.Argument` syntax class.

### DIFF
--- a/enforest/main.rkt
+++ b/enforest/main.rkt
@@ -123,7 +123,9 @@
          tl-decl ...
          (define-syntax-class form-class
            #:attributes (parsed)
-           (pattern ((~datum group) . tail) #:attr parsed (transform-out (enforest (transform-in #'tail)))))
+           (pattern ((~datum group) . tail)
+                    #:cut
+                    #:attr parsed (transform-out (enforest (transform-in #'tail)))))
 
          ;; For reentering the enforestation loop within a group, stopping when
          ;; the group ends or when an operator with weaker precedence than `op`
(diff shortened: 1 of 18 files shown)


sorawee:
I like the automatic propagation, which matches the `#:inherit` mode in https://docs.racket-lang.org/loop/. :)

https://docs.racket-lang.org/loop/ also supports the `#:default` mode, which I occasionally find it useful. Is it worth supporting it?

mflatt:
I see that `#:default` in `loop` is different than what I had in mind as a potential extension (as "further tweak" above). Both possibilities seem useful, but maybe `#:default` is more useful. Still, I would be inclined to keep things simpler, at least for now.

Issue #318, opened 2023-06-09
RFC: Improving Syntax.replace_scopes


Syntacticlosure:
Intuitively, one may think id and '$id' have the same behavior if it is just an identifier because in racket they do.

In rhombus, the statement is true in most cases, however, it is not true in the case of `Syntax.replace_scopes`.

The following two pieces of codes are almost identical, but get different results:
```rhombus
import rhombus/meta open
expr.macro 'test: $expr':
  ~op_stx h
  def y: 'y'.replace_scopes(h)
  'block:
     def $y: 2
     $expr'

test: y
```

```rhombus
import rhombus/meta open
expr.macro 'test: $expr':
  ~op_stx h
  def y: 'y'.replace_scopes('$h')
  'block:
     def $y: 2
     $expr'
```
This is because '$y' is not an actual identifier but `(group y)` packed by the function `unpack-group`.

If we extract the identifier inside the `group` structure, we can get the correct result:
```rhombus
import rhombus/meta open
expr.macro 'test: $expr':
  ~op_stx h
  def y: 'y'.replace_scopes('$h'.unwrap_group().first)
  'block:
     def $y: 2
     $expr'

test: y
```
-----------------
I noticed that the current implementation of `Syntax.replace_scopes` is just a recursive wrapper of `datum->syntax`. I don't think it is suitable for rhombus because you don't know whether your syntax object is wrapped by some implicit structures that generated by quasiquote.

I propose that we can restrict the context parameter of `Syntax.replace_scopes` to a single term object or an identifier and apply `unpack-term` to it to get the real scopes.

If you like my proposal, I can make a pull request as soon as possible.
```[tasklist]
### Tasks
```

mflatt:
I agree — that sounds like the right repair to `Syntax.replace_scopes`.

PR #321, opened 2023-06-12
improve Syntax.replace_scopes by unpack-term

Syntacticlosure:
see #318


mflatt:
I expected to see `unpack-term` added to in `strip_scopes` within `synatx-object.rkt`. Is there some reason that the `syntax/strip-context` library needs to be copied to put the call there?

Syntacticlosure:
I updated the code.

mflatt:
Thanks! I plan to merge this, but then refine further.

I see that some other functions are using `extract-ctx`, and that seems like the right thing to use here. However, the functions that use `extract-ctx` (and the error message from `extract-ctx`) should say `Term` instead of `Syntax`. Furthermore, in the case of an operator, parentheses, or other compound term, really the context should be extracted from something that's inside the term (such as a `parens` identifier), instead of using term's outer syntax-object form.

PR #322, opened 2023-06-12
add `all` and `any` as reducers

mflatt:
[Edit: some of this isn't what I think, now, as reflected by the change in names by a force-pushed commit.]

Most of this PR is a straightforward generalization of the reducer macro API to support early termination of an iteration, which is needed for "and" and "or" reducer (analogous to `for/and` and `for/or`). The macro API is also extended to support operators, instead of allowing only prefix identifiers as reducers.

These changes allow `&&` and `||` as reducers. The new reducer don't accept any arguments in a reducer position, and I'm not sure that no-fix reducer operators are a good choice. I considered using `and` and `or` for the reducer names, instead, but `&&` and `||` match the name used for "and" and "or" everywhere else, so I went with those for a first draft.

There's also the `every`/`all` versus `any` convention, but because I'm more used to `andmap` and `ormap`, I usually have to work out which of `every` or `any` is "and" and which is "or". But maybe that's just a question of what I'm used to. Then again, `Any` has a stronger precedent for us as an annotation or contract.

Example:

```
fun lex_greater_or_equal(l1 :: List.of(Real),
                         l2 :: List.of(Real)):
  l1.length() >= l2.length()
    && (for &&:
          each:
            e1: l1
            e2: l2
          e1 >= e2)
```

### DIFF
--- a/enforest/main.rkt
+++ b/enforest/main.rkt
@@ -123,7 +123,9 @@
          tl-decl ...
          (define-syntax-class form-class
            #:attributes (parsed)
-           (pattern ((~datum group) . tail) #:attr parsed (transform-out (enforest (transform-in #'tail)))))
+           (pattern ((~datum group) . tail)
+                    #:cut
+                    #:attr parsed (transform-out (enforest (transform-in #'tail)))))
 
          ;; For reentering the enforestation loop within a group, stopping when
          ;; the group ends or when an operator with weaker precedence than `op`
--- a/rhombus/main.rkt
+++ b/rhombus/main.rkt
@@ -3,6 +3,7 @@
 
 (bounce "private/core.rkt"
         "private/core-macro.rkt"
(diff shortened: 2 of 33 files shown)


mflatt:
After trying out `all` and `any`, I now think those are much better reducer names — despite the overloading of `any` and `Any`. I blame my earlier choice on too many years with `ormap` and `andmap`. I've force-pushed an update to use `all` and `any`.

Here's the example again, but with `&&` replaced by `all`:

```
fun lex_greater_or_equal(l1 :: List.of(Real),
                         l2 :: List.of(Real)):
  l1.length() >= l2.length()
    && (for all:
          each:
            e1: l1
            e2: l2
          e1 >= e2)
```

sorawee:
`for some` for `ormap` is another possibility, if overloading `any` is a concern.

PR #325, opened 2023-06-16
add `with` for functional update of objects

mflatt:
Example:

```
 class Posn(x, y)
 Posn(1, 2) with (x = 10)
 // prints `Posn(10, 2)`
```

The general syntax is `obj_expr with (field_id = expr, ...)`. For an `obj_expr` with static information for class C, the result is obtained by calling the constructor for C with arguments from `obj_expr`'s field values, except that for each mentioned `field_id`, the result of the corresponding `expr` is used, instead.

A `class` declaration can replace `with` or fill it in for a custom constructor by using the new `udpdate` class-clause form, which is similar to `expression` or `annotation`.

The `with` form requires static information from `obj_expr` to identify a class. Otherwise, a syntax error is reported.

When a class is abstract or has a custom constructor, then there's no default implementation for `with`, but `update` can supply one.

The `with` syntax here is based on OCaml, but using parentheses instead of curly braces. I first tried `.with` and specialization via the existing `dot` clause in a class, but using `with` as a kind of magic name turned out to be clumsy (unsurprising in retrospect).

As of the initial PR, `update` is not allowed within `interface`, but probably it should be. A sensible implementation for an interface might call a method, instead of using a constructor.

### DIFF
--- a/rhombus/private/class-clause-parse.rkt
+++ b/rhombus/private/class-clause-parse.rkt
@@ -114,6 +114,10 @@
               [(#:dot name block)
                (hash-set options 'dots (cons (cons #'name (extract-rhs #'block))
                                              (hash-ref options 'dots null)))]
+              [(#:update name block)
+               (when (hash-has-key? options 'update-rhs)
+                 (raise-syntax-error #f "multiple update clauses" orig-stx clause))
+               (hash-set options 'update-rhs (extract-rhs #'block))]
               [(#:nonfinal)
                (when (hash-has-key? options 'final?)
                  (raise-syntax-error #f "multiple finality clauses" orig-stx clause))
--- a/rhombus/private/class-clause-primitive-meta-macro.rkt
+++ b/rhombus/private/class-clause-primitive-meta-macro.rkt
@@ -21,7 +21,9 @@
 (provide (for-spaces (rhombus/class_clause
                       rhombus/interface_clause)
                      dot
-                     static_info))
+                     static_info)
+         (for-spaces (rhombus/class_clause)
+                     update))
 
 ;; see also "class-clause-primitive-macro.rkt"; this one has only
 ;; forms that need meta-time bindings, so we don't want a meta-time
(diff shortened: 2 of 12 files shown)


samth:
> The with form requires static information from obj_expr to identify a class. Otherwise, a syntax error is reported.

Why not fetch this information dynamically if it's not available statically? Or is this a change you were planning for the future? (The inability to do this is a long-standing challenge in Racket.)

AlexKnauth:
> Being able to write `l.head with (x = 1)` would be really nice

An expression like `l with (head.x = 1)` would be more intuitive as an update of `l` with a subfield of a field changed

gus-massa:
I prefer

```
Posn(1, 2) with (.x = 10)
l with (.head.x = 1)
```

Perhaps I'm biased by vb6 https://learn.microsoft.com/en-us/dotnet/visual-basic/language-reference/statements/with-end-with-statement#example-1 It was to access an object multiple times instead of a functional update, but I like how the `.` shows that it's a field of the object instead of an free variable.

mflatt:
Revised to a dynamic choice of class to instantiate:

The general syntax is still `obj_expr with (field_id = expr, ...)`, where each `field_id` is a public field of the object that is represented in the constructor. A new object is generated by calling a "reconstructor" for the class that the object instantiates; by default, the reconstructor calls the class's constructor.
    
In the case that static information is available for the target object, the update can be more efficient. For a final class with the default reconstructor, a static update expands to a direct call to the constructor. Otherwise, a static update is the same as a dispatch to a method, where the method has optional arguments with defaults drawn from the target object's fields.
    
If no static information is available, then a dynamic lookup is used for `with`. In that case, the fields must be compared against the class's field list at run time, and an `apply` is eventually used to call the reconstrtcutor. No keywords are involved, however, since all fields are supplied to a reconstructor by position.
    
A `class` declaration can specialize update with a `reconstructor` clause, similar to writing a method.

The reconstructor protocol is not as inherently extensible as the constructor protocol. Specifically, when you define a reconstructor in a subclass, there's no built-in way to chain to the superclass reconstructor. I considered replicating the constructor protocol for reconstructors, but then you have two parallel implementations of the same protocol, and that can push work back on programmers. So far, just using a class's constructor seems right for almost anything I can imagine, and a class can define its own method-based update protocol in the rare case that something fancier is needed.

mflatt:
Here are some rough timings (single run on my M1 laptop), making 2M instances either by calling the constructor (= `make`), using `with` on a single field and static information on an object (= `static update`), and using `with` where the right-hand side is wrapped in `dynamic` to hide static information (= `dynamic update`). The `2+1 fields` rows are for a subclass where the base class has 2 fields and the subclass adds 1 more, and the subclass is either final or nonfinal.

Allocation is fast enough and the objects are small enough that the number of fields ends up not being measurable for `make`, but you can see the difference when fields need to be copied for functional update. The main point of interest is the cost of final update (calls constructor) versus nonfinal update (method dispatch), because the latter is the cost of dynamic dispatch; it's about a factor of 2. Completely dynamic (no static information) update is 10-20 times as slow.

```
                    2M loop time in msec

                           static   dynamic
                    make   update   update
   final 1 field      6       6       97
   final 3 fields     6      15      209

   final 2 fields     6      10      152
nonfinal 2 fields     6      23      156

   final 2+1 fields   6      19      211
nonfinal 2+1 fields   6      38      213
```

mflatt:
Further generation: a `reconstructor_fields` declaration allows the field names that work with `with` to be different from the representation's fields. This extension allows `Font`, `Pen`, `Brush` and `Color `from `rhombus/draw` to work with `with`, instead of having a separate `~like` constructor-argument convention.

PR #329, opened 2023-06-27
fix in-common-superinterface? gather condition

AlexKnauth:
To avoid an error message like
```
.../rhombus/private/class-method.rkt:454:19: Map.get: no value found for key
  key: #'size
```
while it's looking for `size` in a super-interface that doesn't have `size`, like `S` in
```
interface S

interface CV:
  extends S
  // read-only property
  property size :: Int
```


mflatt:
This looks like a good repair to me — thanks!

PR #338, opened 2023-07-13
revise `Printable` interface for pretty printing and cycle detection

mflatt:
This PR is aimed at adding pretty printing to Rhombus, including support for detecting graphs and cycles when printing, and making it easier for `Printable` implementations to cooperate.

The new `Printable` interface's only method here is `describe`, which takes a printing-mode symbol and a `recur` function. The mode is `#'text` or `#'expr`. The `describe` method doesn't print anything, but instead returns a `PrintDesc` value that describes what should be printed, possibly offering multiple formatting options that a pretty printer can select among. The `recur` function is for getting a description for a component while detecting cycles and graph references.

A plain string satifies `PrintDesc`, so easy cases are easy:

```
  class Posn(x :: Int, y :: Int):
    private implements: Printable
    private override describe(mode, recur):
      "⟨⟨⟨" +& x +& ", " +& y +& "⟩⟩⟩"
```

Helpers like `PrintDesc.list` can also make common patterns easy:

```
  class Posn(x, y):
    private implements: Printable
    private override describe(mode, recur):
      PrintDesc.list("⟨⟨⟨", [recur(x), recur(y)], "⟩⟩⟩")
```

The `PrintDesc.list` helper uses primitives `PrintDesc.concat`, `PrintDesc.newline`, and `PrintDesc.or` to build up a pretty-printing description that will print `Posn(1, 2)` as either

```
   ⟨⟨⟨1, 2⟩⟩⟩
```

or

```
   ⟨⟨⟨1,
      2⟩⟩⟩
```

or

```
   ⟨⟨⟨
     1,
     2
   ⟩⟩⟩
```

depending on the pretty printer's configuration and choices. The print-description primitives are based on the `pretty-expressive` package and associated OOPSLA'23 paper by @sowaree et al., and the implementation uses `pretty-expressive`.

This commit also replaces `print` and `display` with just `print` plus a `~mode` argument to `print`. Along similar lines, `to_string` get a `~mode` argument. The context parameters `current_input_port`, etc., move to `Port.current_input`, etc.

Timing results below try to show how much the intermediate `PrintDesc` data structure costs compared to direct printing, and how much the `pretty-expressive` algorithm costs in its default configuration.

The compared implementations:

 * old: Rhombus before these changes. Printing is always single-line, and there's not an intermediate data structure for graph detection.
 
 * dumb: new `Printable` API with graph detection (but there are no graphs in the tests), but using a naive pretty printer that tends to put everything on a new line in "multi" mode; single-line mode is selected through a Rhombus parameter that causes the pretty-printing description sent to `pretty-expressive` to have only a single-line option (and that's also true for the "dumb" variant)
   
 * expressive: as in the PR, which is the new `Printable` API with graph detection and using `pretty-expressive`, with the same parameter-based selection of single-line mode; non-single-line mode is "pretty" instead of "multi", because the pretty printer actually prints pretty

 * Racket: a rough comparison with plain and pretty printing in Racket; the pretty-printing API is difficult to use, and my benchmark implementation is sketchy, but it's maybe close enough to give a rough comparison

Overall, the cost of pretty printing with `pretty-expressive` is enough here that a programmer might need to be aware of `~pretty: #false` for `print` or to set the `Printable.current_as_pretty` context parameter. The deep trees and longish lisst are especially slow. But maybe the gap can be reduced by fixing something wrong in the use of `pretty-expressive` or tuning at some level.

```
                 old         dumb         expressive        Racket
                single   single  multi   single  pretty  plain  pretty
posn_tree 1       1         5     11        5      54       2      26
          3       7        24     85       26     626      18     115
          5      57       131    268      145    4820      72     777
          7     123       435   1067      576   30744     331    6834

    list  3      99       192    360      197    1586      64      761
         20     413       531    967      529   19953     264     2597

     string      12        12     12       12      12      12      243
```

Benchmark code: [print.zip](https://github.com/racket/rhombus-prototype/files/12034917/print.zip)

For the "old" column, use  only commit 4fb42b78cd490b2e6b675eb8f26b6b010f58d51b in this PR.

For the "dumb" column, look for "dumb" in two places in "print-desc.rkt" in Rhombus's implementation.

The posn_tree benchmark is "posn_tree.rhm", "posn_tree_old.rhm", "posn_tree.rkt", and "posn_tree_pretty.rkt". Change the argument to `blab` for the tree size, change `~pretty` to select single- or multi-line/pretty mode.

The list benchmark is "list.rhm", "list_old.rhm", and "list.rhm". Similar changes apply as in posn_tree to get variants.


mflatt:
Some revised numbers. I was able to speed up single-line printing a little by improving the graph-detection pass. I also made that pass normalize the printing description so that it works as input to a simple greedy printer; that printer aims for the same kind of behavior as Racket's pretty printer, and its performance compares favorably.

Since there's a significant cost to pretty printing as and a large cost to optimal printing via `pretty-expressive`, I've made non-pretty mode the default (except for the REPL, which is pretty by default) and made the greedy printer the default for pretty mode.

There's room for improvement, but overall, I think the performance results, ease of use, and flexbilty for graph detection and pretty printing justify this `Printable` protocol.

```
                 old        greedy         expressive        Racket
                single   single  pretty  single  pretty  plain  pretty
posn_tree 1       1         3     10        4      40       2      20
          3       7        19     72       16     532      13     115
          5      57       104    351       71    3967      68     752
          7     123       441   1632      310   24617     318    6834

    list  3      99        89    303      136    1570      61     760
         20     413       497    844      446   20016     259    2608
        100     413      2283   4043     1986  108440    1868    7484

     string      12        12     12       12      12      12      243
```

PR #342, opened 2023-07-27
scribble/rhombus: fix braces case in spacer

usaoc:
This fixes the incorrect rendering of braces expression with brackets.


mflatt:
Thanks!

PR #349, opened 2023-08-31
add initial-`each` shorthand for `for`

mflatt:
This commit changes `for` so that an initial `each` clause can be written with parentheses just before the `for` block. So, this change is sortof a step back toward Racket's `for`, but without a `for*` variant.

Examples

```
  for:
    each i: 1..4
    println(i)
  //  =>
  for (i: 1..4):
    println(i)

  for List:
    each i: 1..4
    i+1
  //  =>
  for List (i: 1..4):
    i+1

  for values(sum = 0):
    each i: 1..4
    sum+i
  //  =>
  for values(sum = 0) (i: 1..4):
    sum+i

  for:
    each:
      friend: ["Alice", "Bob", "Carol"]
      index: 1..4
    println(index +& ". " +& friend)
  //  =>
  for (friend: ["Alice", "Bob", "Carol"],
       index: 1..4):
    println(index +& ". " +& friend)
```

For all the current reducer forms, there turns out to be just one case of ambiguity. Only a `values` reducer can end in parentheses, and it requires `=` for each binding, while an `each` shorthand must use `:`. But in `for values(): ....`, the empty parentheses are parsed as an empty `each` shorthand, instead of an empty sequence of `values` bindings. This is a concern, and it's potentially a long-term concern as new reducers are written. So far, though, it seems like an ok compromise to me, since `values()` would be very rare as a reducer. It could be an issue for a macro that expands to `for values(....): ....`, though, where the macro writer would have to be careful to expand to `for values(....) (): ....`.

```
  for values(): // syntax error `values` expects more terms
    values()

  for values() (): // ok
    values()
```

### DIFF
--- a/rhombus/private/array.rkt
+++ b/rhombus/private/array.rkt
@@ -112,14 +112,23 @@
    (lambda (stx)
      (syntax-parse stx
        [(_ #:length e ...+)
-        (values (reducer/no-break #'build-array-reduce
-                                  #'([i 0])
-                                  #'build-array-assign
-                                  array-static-infos
-                                  #'[dest
-                                     (rhombus-expression (group e ...))
-                                     i])
-                '())]
+        ;; If `e` ends with parentheses after at least one term, then
+        ;; let those parentheses be for the initial `each` shorthand,
+        ;; instead of treating it as part of the length expression
+        (define (finish len-g tail)
+          (values (reducer/no-break #'build-array-reduce
+                                    #'([i 0])
+                                    #'build-array-assign
+                                    array-static-infos
+                                    #`[dest
+                                       (rhombus-expression #,len-g)
+                                       i])
+                  tail))
+        (syntax-parse #'(e ...)
+          [(e ...+ (~and p (_::parens . _)))
+           (finish #'(group e ...) #'(p))]
+          [_
+           (finish #'(group e ...) #'())])]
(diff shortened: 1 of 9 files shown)


AlexKnauth:
I like that this allows `for` forms to be written on one line more easily than https://github.com/racket/rhombus-prototype/pull/280 would!

mflatt:
The latest push revises parsing to put reducer macros in control, and both of these are now valid:

```
  for values(): // ok
    values()

  for values() (): // ok
    values()
```

There is still one subtle case involving an ambiguity that I had overlooked before:

```
  for Array ~length pick_the_size():
    each i: 1..10
    i
```

The `Array` reducer macro now must decide how to treat that closing `()`, either as part of the expression after `~length` expression or as the `for` shorthand. Right now it makes a simple choice that an ending `()` not by itself is left for `for`, and so an expression after `~length` might need to be wrapped in  parentheses. A better choice seems possible, more along the lines of letting the expression enforester decide whether it wants the parenthesized term. But the more general point is that the choice is up to the reducer macro instead of `for`; that has trade-offs, but overall it seems better than before.

mflatt:
I ended up pushing this to correct something that I must have gotten wrong by reordering commits. Given the discussion so far, maybe it makes sense to try it out at this point, anyway.

PR #330, opened 2023-06-27
WIP: Collection interfaces with mutableSublist producing a write through view

AlexKnauth:
An updated version of https://github.com/racket/rhombus-prototype/pull/259, rebased onto https://github.com/racket/rhombus-prototype/pull/329 to fix an issue that came up while working on it, with some of the ideas from the `mutableSublist` discussion implemented into it.

From a discord discussion on 2023-05-03:
 * Branching off from a discussion about naming interfaces like `ListView`, `StringView`, etc. into a discussion about `sublist` and copies vs. unmodifiable read-through views vs. write-thorugh views around here: https://discord.com/channels/571040468092321801/871953739982995547/1103365932753113149
 * The option for the `sublist` method to produce an unmodifiable read-through view while a separate method `mutableSublist` produces a write-thorugh view was first brought up here: https://discord.com/channels/571040468092321801/871953739982995547/1103395023845855293

PR #351, opened 2023-09-05
Propose a design for generic lists

jackfirth:
This builds off #201, #221, #259, and #330, providing a more concrete design and plan for a generic interface for immutable lists. Currently still in draft status.

I'm working on this with @OxSon, who is handling most of the RRB tree implementation and benchmarking.

[Rendered](https://github.com/racket/rhombus-prototype/blob/list-proposal/design/list/list.md)

### DIFF
--- a/design/list/list.md
+++ b/design/list/list.md
@@ -1,0 +1,89 @@
+Generic Immutable Lists
+=======================
+
+This document proposes a design for generic immutable lists in Rhombus, as well as a plan for implementing them.
+
+Motivation
+----------
+
+Lists come in many flavors, with many different performance characteristics. Forces programmers to interact with each of these flavors with different APIs leads to complexity, confusion, and inconsistencies. It hampers expressive power, as algorithms that operate on lists must tie themselves to specific implementations. Performance also suffers, as a lack of support for the most optimal list implementation for a task can lead programmers to choose suboptimal implementations instead. A generic interface for list processing can solve some of these problems.
+
+Out of scope
+------------
+
+This design document focuses solely on immutable lists, and leaves open the design of mutable lists as well as other collections. Discussion on these topics can be found in #201 and #221.
+
+Summary
+-------
+
+A `List` interface is provided from a `rhombus/list` module alongside various implementations of it. The basic structure of the interface is this:
+
+```
+interface List:
+    method get(index)
+    method set(index, element)
+    method add(element) # Adds to the end of the list
+    method insert(index, element)
+    method remove(element) # Removes first occurrence of element from list
+    method removeAt(index)
+    method sublist(start, end) # Returns a view
+```
+
+Various other methods that can be derived from these may be included, but this is the core interface that list implementations must supply. The following implementations will be provided:
+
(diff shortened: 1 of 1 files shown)


samth:
A few scattered thoughts:

- translating `[1,2,3]` to builder notation seems like it would be a significant performance hit; can't we translate it to something like `List.unsafeFromVector(racket.vector(1,2,3))` (or use quoted vectors for the argument)?
- what does `cons` on various kinds of lists do? Does it copy? 
- does it make sense to have hybrid representations to avoid copying?

jackfirth:
>     * translating `[1,2,3]` to builder notation seems like it would be a significant performance hit; can't we translate it to something like `List.unsafeFromVector(racket.vector(1,2,3))` (or use quoted vectors for the argument)?

Sure. Alternatively, we could translate it to `List(1, 2, 3)` and then do whatever optimization we do for `List(x, ...)`. Now that I think of it that's probably better.

>     * what does `cons` on various kinds of lists do? Does it copy?

There is no `List.cons()` method. You could write `List.insert(0, element)`, which would do whatever the underlying `insert()` method does. That method is expected to avoid copying in any persistent implementation of `List`. For `ConsList` specifically, there could be a `cons()` function that takes an element and another `ConsList`. That's what we've got today anyway.

>     * does it make sense to have hybrid representations to avoid copying?

Maybe. We've certainly got the option to try it in the future. I'd like to get more done before exploring it though, since mutable collections and non-list collections might complicate the picture significantly.

jackfirth:
This is ready for review now.

rfindler:
One question I have is about `append`: will that be part of the builder interface? (It seems like a common pattern to build up a list with appends, not that I've done any formal studies or anything!)

jackfirth:
> One question I have is about `append`: will that be part of the builder interface? (It seems like a common pattern to build up a list with appends, not that I've done any formal studies or anything!)

It will, in the form of a `ListBuilder.addAll()` method. I'm not planning on implementing it or the other batch methods until everything else is done though.

mflatt:
The proposal should make clearer that `ConsList` is intended as distinct from a Racket list  (that I learned via Discord). A `ConsList` might have Racket pairs internally, but with a wrapper around the head pair or empty list. That's different than the current Rhombus `List`, which is a Racket list (so, I did not understand that facet of this proposal, at first).

In retrospect, if `List` can refer to something with a different representation than Racket lists (e.g., using an RBB-tree), then I have to agree that a Racket list cannot be a Rhombus `List`. Trying to make that correspondence work would create more problems than it solves in terms of implementing `==`, bridging to Racket, and managing programmer expectations. A Racket list can be used as a sequence, I expect, but not as something that is `==` to any `List`.

This seems like a significant performance challenge right at the start, since not using Racket lists is likely to impose a cost (relative to Racket) on most programs. I had previously imagined that multiple `List` implementations might be handled by inlining dispatch to the common(?) case of cons cells—which would inflate code but maybe not make things much slower—but I doubt that it works out if lists have an extra wrapper.

samth:
More generally related to what @mflatt said, I think we need to be aware that most of the models we've considered for collections generally are in typed languages with compilers (often the JVM) that can eliminate more of the cost of dynamic dispatch and particularly wrappers relative to what Racket can do. We may need to consider different collections of tradeoffs than those systems ended up with (such as different choices about equality).

mflatt:

PR #352, opened 2023-09-05
scribble/rhombus: fix wrong property in detection of space suffix

### LINKED ISSUE #350: Spacing can be missing from `typeset-rhombus`
Consider the following program:

```
#lang rhombus/static
import lib("scribble/private/typeset-rhombus.rkt"):
  expose #{typeset-rhombus}

#{typeset-rhombus}('1 + 1')

#{typeset-rhombus}('(1) + 1')
```

The former is (eventually) rendered as `1 + 1`, while the latter as `(1)+ 1`.  More generally, any bracketed (`()`, `[]`, and `{}`) expression can lead to this behavior.  This causes types like `(Int, Int) -> Int` and `(Int * Int) -> Int` in [the Shplait documentation](https://docs.racket-lang.org/shplait/sec_inference.html) to be incorrectly rendered.

usaoc:
Fix #350.

### DIFF
--- a/scribble/private/typeset-rhombus.rkt
+++ b/scribble/private/typeset-rhombus.rkt
@@ -73,7 +73,7 @@
            (define (suffixed? stx)
              (not (null? (or (eq? 'space (ends (syntax-raw-suffix-property stx)))
                              (syntax-case stx ()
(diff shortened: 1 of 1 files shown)


mflatt:
Thanks!

PR #355, opened 2023-09-11
scribble/rhombus: fix styling and linking of Rhombus module names

usaoc:
Note: I’m not sure if the `index_element_with` approach is the best one, but this is what I can come up with given the current Scribble API.

### DIFF
--- a/scribble/private/docmodule.rhm
+++ b/scribble/private/docmodule.rhm
@@ -43,29 +43,33 @@ meta:
                                        ...])]
 
 decl.macro 'docmodule ($(option :: Option), ..., $mod ...)':
-  def is_lang: (match '[$option, ...]'
-                | '[$_, ..., ~lang, $_, ...]': #true
-                | ~else: #false)
-  '$(decl_meta.pack_s_exp(['defmodule',
-                           '#{#:require-form}',
-                           expr_meta.pack_expr('fun (t_mod): @rhombus(import: $mod ...)'),
-                           if is_lang
-                           | as_racket_mod_name('$mod ...')
-                           | pack_tree(ModulePath.s_exp(ModulePath('$mod ...'))),
-                           option.form, ..., ...]))'
-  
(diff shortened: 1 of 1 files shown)


mflatt:
I see what you're fixing here, but I haven't yet paged back in enough to understand the `index_element_with` part. I'll look more soon!

mflatt:
I think you're right that we don't want to rely on the details of `defmodule` like that, so I'll look at improving `defmodule` — hopefully in some way that lets the documentation still build with an older Racket version.

I'm struggling, though, to see exactly the intended effect of your change. With the change, I can search on "measure", and I get a link to `(lib "rhombus/measure.rhm")`. Without the change,  searching"measure" find only the section title. I still haven't sorted out enough to see why the search result is different, though, and it seems like the goal would have been to show `rhombus/measure`. Are you observing or aiming for a different effect here?

usaoc:
Yes, a link to `(lib "rhombus/measure.rhm")` is intended.  The search results are based on indexes, so after the change the typeset module name will be searchable in the usual Racket module path syntax.  I don’t think we should expose something like `rhombus/measure` to be searchable, at least yet, because that can lead people to think that `(lib "rhombus/measure")` is available.

Looking at `defmodule` more carefully, I’ve realized that `#:module-paths` can probably be used to override the typesetting of module names.  I’ll try to see if that works out.

usaoc:
The new commit uses `#:module-paths` to specify the Racket module path, and it seems to end up searchable.  However, the search result is like `rhombus/measure` (Rhombus module name) as opposed to `(lib "rhombus/measure.rhm")` (Racket module path).  Would you find this more preferable?

mflatt:
Yes, this looks right. I think it makes sense to index the Rhombus name `rhombus/measure` (since Rhombus search results are generally in Rhombus syntax, and search results more generally only make sense for some language). Thanks for sorting this out!

PR #366, opened 2023-09-20
repairs to composite static infos

usaoc:


mflatt:
Thanks! Can you provide an example/test that failed before each repair, so we can add that to the test suite and make sure there's no regression in the future?

usaoc:
Sure!  I also realized that I wasn’t handling all cases right when writing the tests; those cases are now fixed.

mflatt:
Perfect - thanks!

PR #371, opened 2023-09-26
Rename `List` to `ConsList`

jackfirth:
This is the first step on the road to a generic `List` interface, as outlined in #351. It's still in draft status for now until it's not completely broken.


mflatt:
Done via #455 (as `PairList` instead of `ConsList`, though).

PR #382, opened 2023-10-10
add exponentiation

rfindler:
At the risk of starting a bikeshed discussion, `^` was very tempting, but [this wikipedia page](https://en.wikipedia.org/wiki/Exponentiation#In_programming_languages) suggests that `**` is a lot more common.


rfindler:
Also, I'm not very sure I got all the places to declare precedence.

sorawee:
Perhaps PR is not the right place for this discussion, but if adding a new operator requires changing a lot of existing definitions like this, does that mean the protocol is not extensible enough?

rfindler:
@sorawee I think that this stems from the (imo, good) decision that unless precedence is declared between two operators, then there must be parens. Which means that the transitive edges in the relation aren't added. I guess you know/agree with that! So, that said, maybe a good way to go would be to add layer on top of that, where by some smaller set of operators could opt into a specific set of precedence declarations where, for that specific subset, the transitive edges were added automatically. Then that could be used locally to the arithmetic file here.

soegaard:
> At the risk of starting a bikeshed discussion, ^ was very tempting, but [this wikipedia page](https://en.wikipedia.org/wiki/Exponentiation#In_programming_languages) suggests that ** is a lot more common.

It seems there are three camps of languages: the Fortran/JavaScript/Python one, the BASIC/Mathematica/TeX one and the no-operator-for-exponentiation camp (C/Go/Rust/Swift).

Which camp to choose depends on which kind of programs is expected to most common.

Programs using a lot of bit manipulations will probably use ^ for xor.

Programs of a mathematical nature will be in the ^ camp.

The question is whether Rhombus needs a guiding principle that can lead to a consistent set of operators. Is it time to choose a camp - or is the tactic to choose a minimal default set, 
so that users can choose which operators they need?

mflatt:
I think `**` is the right choice because it's what the languages closest to Rhombus syntax use (including Python and OCaml) and because it leaves the `^` character for other purposes.

The new operator could be added without changing existing definitions (by specifying all relations in the new operator), but it can be convenient to change others when adding to the same module. Ideas for better approaches are welcome, though; I have wondered whether some way to define sets of operators like "common arithmetic operators" and "common comparison operators" might be useful.

rfindler:
I forgot docs! Added now.

I also took the opportunity to fix some docs right in the same place (but in a separate commit).

PR #385, opened 2023-10-13
fix srcloc of `!` module path

usaoc:


mflatt:
I think some larger changes are needed here, and I pushed 54e134ebe477f402fba05bf90a132ce9558724ff. I'm not sure that it covers the problem that your change was aimed at, though.

usaoc:
Yes, I think it covers the problem.  For the record, the problem was that an “unused require” could be drawn all over the place due to the wrong srcloc.

PR #387, opened 2023-10-14
fix `each` shorthand after `any`/`all`

usaoc:


mflatt:
Thanks for all your repairs!

PR #390, opened 2023-10-17
revise `pattern` shorthand

usaoc:
Now that we allow a block before alts, the `pattern` syntax class clause is redundant and arguably disfavored.  Not only that, it also ties the meaning of the corresponding (unquote) binding operator to it.  I think it’d be better that the syntax class clause simply be removed, and the (unquote) binding operator be changed into an actual shorthand for an inline syntax class matching.

My experience is that inline syntax classes are quite useful for the same reason complex patterns in `syntax/parse` are useful, but using them unfortunately requires an unnaturally large amount of rightward shift.  Moreover, supplying `open` with inline syntax classes is common but, again, looks ugly.  Therefore I believe `pattern` can serve a better purpose this way.

Interestingly, syntax class clause macros originally can supply spliced patterns through `pattern`.  This use is kept through changing the expected expansion to a block optionally followed by alts.

Two remaining problems:

1. ~~Some test cases that seem to be testing the “scopes” of attribute names don’t work anymore.  Actually, I’m not sure if they’re supposed to work at all in the first place since the corresponding syntax classes don’t work this way.  If they should work, then the fix should be generalizable to all syntax classes.~~  
   Reverted to the original behavior.

3. The handling of repetitions *above* the quasiquote doesn’t seem to cover attributes.  Although this is not directly related to `pattern` per se, the issue may be apparent when the `pattern` binding is used in a composite binding form.  Again, the fix should be generalizable.

### DIFF
--- a/rhombus/private/check.rhm
+++ b/rhombus/private/check.rhm
@@ -13,27 +13,25 @@ export:
 meta:
   syntax_class ResultMode:
     kind: ~term
-    pattern
-    | '~is'
-    | '~is_now'
-    | '~is_a'
-    | '~prints_like'
-    | '~prints'
-    | '~raises'
-    | '~matches'
+  | '~is'
+  | '~is_now'
+  | '~is_a'
+  | '~prints_like'
+  | '~prints'
+  | '~raises'
+  | '~matches'
       
   syntax_class Result:
     kind: ~sequence
-    pattern
-    | '$(mode :: ResultMode): $body':
-        field [suffix, ...]: []
-    | '$(mode :: ResultMode) $g ...':
-        field body: '$g ...'
-        field [suffix, ...]: []
-    | '~completes':
-        field mode: '~completes'
-        field body: '#void'
-        field [suffix, ...]: ['#void']
+  | '$(mode :: ResultMode): $body':
+      field [suffix, ...]: []
+  | '$(mode :: ResultMode) $g ...':
+      field body: '$g ...'
+      field [suffix, ...]: []
+  | '~completes':
+      field mode: '~completes'
+      field body: '#void'
+      field [suffix, ...]: ['#void']
 
   syntax_class EvalClause:
     fields: make_eval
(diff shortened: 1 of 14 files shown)


mflatt:
This change seems good overall to me. The approach to `syntax_class` clause macros is interesting. While this is a non-backward-compatible change, I don't find existing code that needs to change.

I'm unclear on remaining problem 1, though. All the tests seem to pass, and I don't see any removed. Is that problem referring to tests in `rhombus/tests`?

usaoc:
> I'm unclear on remaining problem 1, though. All the tests seem to pass, and I don't see any removed. Is that problem referring to tests in `rhombus/tests`?

Yes, in `rhombus/tests/syntax-class.rhm` under “check interaction of macro introduction and pattern variables,” which are commented out with a note.  For example, this case (changed slightly)

```
#lang rhombus/and_meta
expr.macro 'matcher $(ex :: Term)':
  '«match '1 2'
    | (pattern | '$('$')x $('$')$ex'): x»'
matcher x
```

fails with `syntax-parse: duplicate attribute`, which is consistent with the behavior of inline syntax classes.

mflatt:
Ah, I see. (I may have missed that by not clicking "Load diff".)

I'll need to think about this more, but my initial reaction is that the new `pattern` behavior seems wrong.

In general terms, it's a question of "internal" names, like ones normally bound by unquote escapes or by `def`, and "external" names, like syntax-class fields and package exports. Internal names are identifiers, and external names are symbols. Where there's a transition between those two, there normally should be a renaming form so that the two kinds of names can be distinct. An `open` form is a shorthand (in `::` as well as `import`) that makes the two symbolically the same, which has certain limitations, but you have to opt into that.

The new `pattern` essentially has `open` built in. For a use of `pattern` in an unquote binding position, though, it seems like all the names can just be internal/identifiers. That's the intent of the current implementation.

Along those lines, I think the following should be allowed — but isn't in the current implementation — because the two `x`s don't need to be treated as external names:

```
defn.macro 'd $name $id':
  '«
     syntax_class $name:
       fields:«»
     | '$('$')x $('$')$id'
    »'

d C x
```

So, it seems like something should be fixed there to make unquote bindings behave consistently.

mflatt:
Thanks!

PR #392, opened 2023-10-23
allow no fields in syntax classes

usaoc:
Technically, this has always been allowed through supplying an empty block after `fields`, but doing so is less convenient than supplying no block at all.


mflatt:
What do you think about allowing `fields <spec> ...` more generally? I think that was the intent of the `(_ (tag::block fld ...))` pattern which was meant to be `(_ fld ...)`.

Allow both block and non-block variants is a general convention. For examples, the `implements` class clause form allows those two shapes. Originally, I imagined that only some canonical choice would be allowed for each form, but I've been happier when writing Rhombus code to have the choice and not have to remember whether a `:` is required.

If the more flexible form seems ok, I suggest documenting as

`
fields <spec> ...
fields:
  <spec> ...
  ...
`

instead of introducing another non-terminal.

PR #393, opened 2023-10-24
fix generation of attributes in inline `pattern`s

usaoc:
I’m not sure what’s the exact purpose of `bind-counter`, as the scopes are already provided by the identifier.  It also doesn’t appear to do the right thing at all, as evidenced by the test cases.

Instead, symbolic names are kept distinct at each *depth* level of `pattern`, which is required for the proper propagation of attributes in nested `pattern`s.  Moreover, attributes in `open`ed syntax classes can get their context from the `pattern` form id, and therefore the implementation of binding `pattern` is changed to not alter its context.

Also fix a bug accidentally introduced in 053523c that `open`s named `pattern`s.

### DIFF
--- a/rhombus/private/attribute-name.rkt
+++ b/rhombus/private/attribute-name.rkt
@@ -1,10 +1,19 @@
 #lang racket/base
 
-(provide compose-attr-name)
+(provide compose-attr-name
+         inline-attr-depth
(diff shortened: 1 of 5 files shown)


mflatt:
Thanks!

PR #394, opened 2023-10-25
scribble/rhombus: refactor and add more spacers

usaoc:
The refactoring tries to make the `relocate`s more consistent and use more inline `pattern`s.  Hopefully the refactoring makes some parsing easier to get right.  For example, 65d1182 didn’t get the parsing quite right because a sequence could be matched too greedily.

The implemented spacers are far from complete, but they should cover most of what `doc`s use.

### DIFF
--- a/scribble/private/escape.rhm
+++ b/scribble/private/escape.rhm
@@ -1,0 +1,10 @@
+#lang rhombus/static/and_meta
+export:
+  meta:
+    Escape
+
+meta:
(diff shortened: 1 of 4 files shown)


mflatt:
Thanks!

Just to have it on our radar: We should figure out how to write some tests here. I haven't given it too much thought, but maybe something that has code annotated with expected spaces would be useful and easy to maintain.

PR #395, opened 2023-10-27
define built-in classes with `define-primitive-class`

usaoc:
The (ab)use of `define-primitive-class` makes the code significantly more organized.  In particular, method tables and dot providers can be automatically generated as opposed to the error-prone manual approach.  To make everything even more automated, an extension of `define/arity` to methods is implemented.  A more detailed changelog:

- Extend `define-primitive-class` to allow the definition of built-in “classes” like `List`;
- Implement `define/method` to generate method definition and dot dispatch;
- Allow `who` in `define/arity` and `define/method`;
- Try to generate error adjusting entries;
- Refactor relevant definitions.

### DIFF
--- a/rhombus/private/array.rkt
+++ b/rhombus/private/array.rkt
@@ -1,31 +1,30 @@
 #lang racket/base
 (require (for-syntax racket/base
                      syntax/parse/pre
-                     syntax/stx
                      "srcloc.rkt")
+         (only-in racket/base
+                  [vector array])
          "provide.rkt"
          "expression.rkt"
          "binding.rkt"
          (submod "annotation.rkt" for-class)
-         (submod "dot.rkt" for-dot-provider)
          "static-info.rkt"
          "define-arity.rkt"
          "index-key.rkt"
          "append-key.rkt"
          "call-result-key.rkt"
          "index-result-key.rkt"
-         "function-arity-key.rkt"
          "sequence-constructor-key.rkt"
          "composite.rkt"
          "op-literal.rkt"
-         "name-root.rkt"
-         "dot-parse.rkt"
          "reducer.rkt"
          "parens.rkt"
          "parse.rkt"
          "realm.rkt"
          "mutability.rkt"
-         "vector-append.rkt")
+         "vector-append.rkt"
+         "class-primitive.rkt"
+         "rhombus-primitive.rkt")
 
 (provide (for-spaces (rhombus/namespace
                       #f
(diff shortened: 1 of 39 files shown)


mflatt:
This is a very welcome improvement! I'd just like to look more closely before merging, and I haven't had a chance yet.

rhombus/private/array.rkt
```diff
@@ -167,33 +168,54 @@
...
-(define/arity (copy v)
+(define (check-array who v)
+  (unless (vector? v)
+    (raise-argument-error* who rhombus-realm "Array" v)))
+
+(define (check-nonneg-int who n)
+  (unless (exact-nonnegative-integer? n)
+    (raise-argument-error* who rhombus-realm "NonnegInt" n)))
+
+(define/arity (Array.get v i)
```
mflatt, line 179: A drawback of defining this `Array.get` with extra checks is that it prevents inlining of `vector-ref` to implement array access. The alternative is reflected by "runtime-config.rkt": use `vector-ref` directly, but rewrite error messages. The current "runtime-config.rkt" doesn't have a mapping from `vector-ref` to Array.get`, but the idea was to fill all those in, eventually.

rhombus/private/box.rkt
```diff
@@ -33,18 +25,42 @@
...
 
 (define/arity (Box v)
   #:static-infos ((#%call-result #,box-static-infos))
   (box v))
 
+(define (check-box who b)
+  (unless (box? b)
+    (raise-argument-error* who rhombus-realm "Box" b)))
+
+(define/arity Box.value
```
mflatt, line 55: Same issue as `Array.get`: the original intent was to let `unbox` be used directly.

rhombus/private/bytes.rkt
```diff
@@ -25,79 +29,109 @@
...
+    (raise-argument-error who rhombus-realm "maybe(Char)" c)))
+
+(define-syntax (define-string stx)
+  (syntax-parse stx
+    [(_ utf8 utf-8)
+     #:with method-name (datum->syntax #'utf8 (string->symbol (format "Bytes.~a_string" (syntax-e #'utf8))))
+     #:with fn-name (datum->syntax #'utf-8 (string->symbol (format "bytes->string/~a" (syntax-e #'utf-8))))
+     #'(define/method (method-name bs [err-char #f] [start 0] [end (and (bytes? bs) (bytes-length bs))])
+         #:static-infos ((#%call-result #,indirect-string-static-infos))
+         (check-maybe-char who err-char)
```
mflatt, line 110: This is a case where functions like `string->bytes/utf-8` cannot be used directly, but maybe the checking built into those methods should be relied on, instead of checking again here. Note that the checks here are incomplete, because there are constraints on `start` versus `end` versus the length of the string.

rhombus/private/list.rkt
```diff
@@ -289,64 +291,72 @@
...
+     (let ([ln (let loop ([ls ls])
+                 (if (pair? (cdr ls)) (loop (cdr ls)) (car ls)))])
+       (unless (list? ln)
          (for ([l (in-list ls)])
-           (unless (list? l) (raise-argument-error* 'List.append rhombus-realm "List" l))))]))
+           (check-list who l))
+         (raise-argument-error* who rhombus-realm "List" ln)))
+     (apply append ls)]))
+
+;; error adjusting in "runtime-config.rkt"
```
mflatt, line 347: For some reason, `List.get` doesn't seem to get inlined as `list-ref` by schemify, so that's worth a closer look. Also, we may want to have an exposed `List.get` operation that checks for listness (not just a sequence of pairs), and then distinction between `list-ref` for `#%index` and `List.get` may be worthwhile; I'm not sure.

mflatt (commented):
I've marked some some small issues to consider about primitives and redundant checks. The same concern has been taken into account in other places, though, so maybe I'm missing something about the places that worry me.

When I run the tests, I see a failure in "rhombus/tests/printable.rhm" where `to_string("apple", ~mode: #'expr)` produces `"apple"` instead of `"\"apple\""`. I haven't yet investigated.

usaoc:
`to_string`: That was just a silly oversight.

I’m not sure what’s the best thing to do here.  I was mostly looking for a more consistent implementation, so I did those changes, but the concerns are valid.  If the idea is to use more rewriting, I can do some changes to improve that (e.g., generating entries seems like a good idea than manually filling in).  Eventually, though, we may want to do the checks at Rhombus side (probably by using more unsafe operations and more `begin-encourage-inline`?).  We’ll definitely need to think more about this; I’ll try to stick closer to what we have now, though.

usaoc:
`List.get`: That was because `syntax-parameterize` (for `who`) was around the whole RHS of the function definition.  I changed it to be around the function bodies.

I took some freedom to add an `#:inline` directive (that adds `'compiler-hint:cross-module-inline` to the function definition) and a `#:primitive` directive (that adds entries to `primitive-who-table`, for which see the newly added `rhombus-primitive.rkt`) to `define/{arity,method}`.  I think this may be closer to what you have in mind.

mflatt:
Thanks for this really great improvement!

PR #403, opened 2023-11-15
miscellaneous cleanups

usaoc:
This is a follow-up to 74f5ae2:
- Use more `who` in `define/arity` definitions;
- Use more `case-lambda` to enable inlining;
- Finish up the to-dos that I left;
- Fix things that I noticed during the process.

### DIFF
--- a/rhombus/private/syntax-meta.rkt
+++ b/rhombus/private/syntax-meta.rkt
@@ -48,40 +48,48 @@
   (define/arity (syntax_meta.value id/op
                                    [sp expr-space-path]
                                    [fail (lambda ()
-                                           (raise-syntax-error 'syntax_meta.value "no binding" id/op))])
-    (define id (extract-name 'syntax_meta.value id/op sp))
+                                           (raise-syntax-error who "no binding" id/op))])
+    (define id (extract-name who id/op sp))
     (syntax-local-value id (if (and (procedure? fail)
                                     (procedure-arity-includes? fail 0))
                                fail
                                (lambda () fail))))
 
-  (define (extract-free-name stx sp)
-    (extract-name 'syntax_meta.equal_binding stx sp #:build-dotted? #t))
+  (define (extract-free-name who stx sp)
+    (extract-name who stx sp #:build-dotted? #t))
 
   (define/arity syntax_meta.equal_binding
     (case-lambda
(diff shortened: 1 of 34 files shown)


rhombus/private/syntax-meta.rkt
```diff
@@ -48,40 +48,39 @@
...
+                       (extract-free-name id2 sp)
+                       phase1
+                       phase2))
 
-  (define/arity syntax_meta.equal_binding
-    (case-lambda
-      [(id1 id2) (free-identifier=? (extract-free-name id1 expr-space-path) (extract-free-name id2 expr-space-path))]
-      [(id1 id2 sp) (free-identifier=? (extract-free-name id1 sp) (extract-free-name id2 sp))]
-      [(id1 id2 sp phase1) (free-identifier=? (extract-free-name id1 sp) (extract-free-name id2 sp) phase1)]
-      [(id1 id2 sp phase1 phase2) (free-identifier=? (extract-free-name id1 sp) (extract-free-name id2 sp) phase1 phase2)]))
```
mflatt, line 66: This function was written with `case-lambda` to avoid calling `(syntax-local-phase-level)`, relying instead on the default in `free-identifier=?` as a kind of shortcut.

rhombus/private/syntax-meta.rkt
```diff
@@ -48,40 +48,39 @@
...
-
-  (define/arity syntax_meta.equal_name_and_scopes
-    (case-lambda
-      [(id1 id2) (syntax_meta.equal_name_and_scopes id1 id2 (syntax-local-phase-level))]
```
mflatt, line 70: Similar intent here for using `case-lambda`.

mflatt (changes requested):
Thanks for these improvements! I notice a couple of places where `case-lambda` was used to avoid computing default arguments. The performance difference is probably difficult to measure, but I think it's worth checking and maybe keeping the `case-lambda` approach.

rhombus/private/syntax-meta.rkt
usaoc: I reverted the last one, but I think this one is fine (it already calls `syntax-local-phase-level` anyway).

usaoc:
The comment reminds me that I also should change some definitions to use `case-lambda` to enable inlining.  I should’ve done this earlier, but well.

PR #410, opened 2023-11-27
fix missing main converter in `operator`

usaoc:
Currently, `operator` is discarding the main converter (and the static infos for that matter) from the outer result annotation.  This commit fixes this issue and cleans up some of the code in `operator.rkt`.

Another bug revealed by the new test cases is that a postfix case can wrongly proceed as an “infix” case due to the presence of annotation operator.  This is also fixed.


mflatt:
Thanks!

PR #412, opened 2023-11-28
use chaperones for `{Array,Box}.later_of`

usaoc:
This commit changes `{Array,Box}.later_of` to use chaperones for the predicate case, and only use impersonators for the binding case.  The binding case is changed to *not* discard the converted value.  Moreover, the error message is improved by including the annotation and, for arrays, the position.

A bug was found that the binding case of `of` annotations did not check the predicate before continuing.  This was fixed and some test cases were added.

### DIFF
--- a/rhombus/private/annotation.rkt
+++ b/rhombus/private/annotation.rkt
@@ -1,28 +1,19 @@
 #lang racket/base
 (require (for-syntax racket/base
                      syntax/parse/pre
-                     syntax/stx
                      shrubbery/print
-                     enforest
-                     enforest/operator
-                     enforest/syntax-local
                      enforest/property
                      enforest/proc-name
                      enforest/name-parse
-                     enforest/operator
                      "srcloc.rkt"
-                     "pack.rkt"
                      "introducer.rkt"
                      "annotation-string.rkt"
                      "keyword-sort.rkt"
-                     "with-syntax.rkt"
-                     "tag.rkt"
                      "macro-result.rkt"
                      (for-syntax racket/base))
          "provide.rkt"
          "enforest.rkt"
          "annotation-operator.rkt"
-         "definition.rkt"
          "expression.rkt"
          "binding.rkt"
          "name-root.rkt"
@@ -90,6 +81,7 @@
              annotation-binding-form
 
              parse-annotation-of
+             parse-annotation-of/chaperone
 
              build-annotated-expression
              raise-unchecked-disallowed))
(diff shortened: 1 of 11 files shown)


mflatt:
Discarding the converted value is definitely wrong, but I don't understand the rationale to disallow converter annotations. Why that restriction, instead of changing the two `(lambda (v) val)`s to `(lambda (v) v)`?

To put it concretely, I would expect to be able to write
```
def b :: Box.later_of(Box.later_of(Int)) = Box(Box("apple"))
```
and get true from `b.value is_a Box` but an error from `b.b.value`.

usaoc:
For a `later_of` value, the converter has to be run for each read and write, which doesn’t sound right.  In particular, the value from each read can be different (by the loosest standard, say `is_now`) if the converter weirdly behaves.  I think this is more surprising than useful.  Even in the above innocent example, the inner box has to be repeatedly impersonated, producing non-`===` results (of course, no such guarantee is made anyway).

For a comparison, a converter in mutable variable runs for init and each write.  Reads don’t have to be converted, because there’s no way to mutate the location other than through the variable.  This is not the case for `later_of` values.

mflatt:
These trade-offs seems familiar from the contracts world, though.

Concretely, there are cases that look useful and would work fine when a `Box.later_of` content converter is applied on both "get" and "set". For example, a normalizing annotation like `ReadableString.to_string` would make sense to apply in both directions. Those are the kinds of converters that we have found useful in Rhombus, so far, and that's the sort that I originally had in mind.

In the case of class instances, I can imagine that a something like `.later_of` for classes could eventually be useful, and that it would be like `struct/c`. Specifically, I imagine that mutable fields would work in the same way as for `struct/c` — which is like `box/c` and a conversion-supporting `Box.later_of`.

The comparison to contracts suggests that we'll eventually want chaperone converters as well as impersonator converters. The existing Rhombus converters are impersonator-like, because we ran into that need first when trying out a design for drawing and GUI libraries. And probably that perspective is why I'm comfortable with a `Box.later_of` that is analogous to `box/c` in impersonator mode. (But maybe it's not quite as simple as that, because `box/c` in chaperone mode still has the potential drawbacks that you describe.)

usaoc:
There doesn’t seem to be a mode of `box/c` that corresponds to what I have in mind, but basically the commit’s version of `Box.later_of` works like a chaperone `box/c` that accepts only flat contracts (ignoring the fact that `box/c` also checks at contract-attach time).

In general, I don’t think converter annotations are quite like higher-order contracts, at least not in my mind.  I think converter annotations make sense if there’s a clear annotation boundary, but views don’t seem to create such a clear boundary because they are nonlinear (i.e., multiple views can coexist) in nature.  Existing use cases in the GUI libraries also seem to me that this restriction is fine, at least for now.  We’ll definitely have to think of how to fit this picture into other parts in the future, though.

rfindler:
I wanted to ask a clarifying question: is the idea that converters are a general mechanism that can be used to implement a variety of things (perhaps contracts, perhaps some form of normalization, perhaps some kind of backwards-compatibility convenience (eg when you generalize a function from accepting a boolean to accepting a natural number, you might still allow true and false as inputs but convert them to 0 and 1 so the function itself isn't polluted with dealing with the old inputs))? Or do we see converters as something more specific with more specific goals that might lead us to want to put some errors in various places to help detect buggy programs?

usaoc:
Yes, in my mind converters should be a rather general mechanism for pre/post-processing of values.  This means that restricting the conversion to be done *once* at boundary is a good idea, and this isn’t applicable to arrays or boxes because they don’t impose a boundary.  In contrast, specific reads/writes do impose a boundary.  This is kind of paradoxical of `later_of` annotations, because they don’t really annotate the values (or actually the references), but annotate the reads/writes!

On this track, maybe a more convenient way to create wrapper classes will be a good idea?  Just like how annotated mutable fields can guarantee the invariant, being a part of the “flat” class annotation.

rfindler:
In that case, echoing @mflatt 's comments, one of the things we've learned through our experience with contracts is that when we're dealing with values that have behavior (functions or objects), we have to delay our conversions until the moment that a value reveals its behavior (ie, later, dynamically). We also seem to have learned something similar about stateful values, namely that treating them in a similar manner seems to be useful. In the mutable case, the behavior isn't in the value itself, but the behavior is in some other place (perhaps another thread) that also has access to the mutable behavior.

usaoc:
I’m not sure I agree with the reasoning for *this case*, but I’ll write a version that accepts converter annotations.

usaoc:
This is now ready for another review.

usaoc:
A note that converter annotations are now allowed, just in case it wasn’t clear.

mflatt:
Thanks - I had lost track of this one.

PR #413, opened 2023-11-29
Add support for using ~matches with multiple values in a check test

ashton314:
Previously something like this wouldn't work:

```
check: values(1, [2, 3, 4])
       ~matches values(a, [b, c, d])
```

Now it does!

I'd be happy to add unit tests—I need someone to show me where they're at if there are any. Thanks!

### DIFF
--- a/rhombus/private/port.rkt
+++ b/rhombus/private/port.rkt
@@ -1,8 +1,10 @@
 #lang racket/base
 (require (for-syntax racket/base)
          "provide.rkt"
+         "call-result-key.rkt"
          "function-arity-key.rkt"
-         "indirect-static-info-key.rkt"
+         (submod "bytes.rkt" static-infos)
+         (submod "string.rkt" static-infos)
          "static-info.rkt"
          "define-arity.rkt"
          "class-primitive.rkt"
@@ -39,22 +41,56 @@
   #:properties ()
   #:methods ())
 
-(define/method (Output.flush [p (current-output-port)])
-  (unless (output-port? p)
-    (raise-argument-error* who rhombus-realm "Output" p))
-  (flush-output p))
-
 (define-primitive-class Output output-port
+  #:lift-declaration
   #:existing
   #:translucent
   #:fields ()
   #:namespace-fields
   ([current current-output-port]
-   [current_error current-error-port])
+   [current_error current-error-port]
+   [open_bytes Port.Output.open_bytes]
+   [open_string Port.Output.open_string])
   #:properties ()
   #:methods
-  (flush))
+  ([get_bytes Port.Output.get_bytes]
+   [get_string Port.Output.get_string]
+   [flush Port.Output.flush]))
 
 (define-static-info-syntaxes (current-input-port current-output-port current-error-port)
   (#%function-arity 3)
   (#%indirect-static-info indirect-function-static-info))
+
+;; TODO these need a more specific annotation
+(define/arity Port.Output.open_bytes
+  #:inline
+  #:primitive (open-output-bytes)
+  #:static-infos ((#%call-result #,output-port-static-infos))
+  (case-lambda
+    [() (open-output-bytes)]
+    [(name) (open-output-bytes name)]))
+
+(define/arity Port.Output.open_string
+  #:inline
+  #:primitive (open-output-string)
+  #:static-infos ((#%call-result #,output-port-static-infos))
+  (case-lambda
+    [() (open-output-string)]
+    [(name) (open-output-string name)]))
+
+(define/method (Port.Output.get_bytes port)
+  #:inline
+  #:primitive (get-output-bytes)
+  #:static-infos ((#%call-result #,bytes-static-infos))
+  (get-output-bytes port))
+
+(define/method (Port.Output.get_string port)
+  #:inline
+  #:primitive (get-output-string)
+  #:static-infos ((#%call-result #,string-static-infos))
+  (string->immutable-string (get-output-string port)))
+
+(define/method (Port.Output.flush [p (current-output-port)])
+  (unless (output-port? p)
+    (raise-argument-error* who rhombus-realm "Port.Output" p))
+  (flush-output p))
--- a/rhombus/private/string.rkt
+++ b/rhombus/private/string.rkt
@@ -1,6 +1,7 @@
 #lang racket/base
 (require (for-syntax racket/base
                      syntax/parse/pre)
+         (only-in racket/string string-contains?)
          racket/symbol
          racket/keyword
          "provide.rkt"
@@ -63,6 +64,7 @@
   #:methods
   ([length String.length]
    [get String.get]
+   [contains String.contains]
    [append String.append]
    [substring String.substring]
    [utf8_bytes String.utf8_bytes]
@@ -95,6 +97,7 @@
    [append String.append]
    [length String.length]
    [get String.get]
+   [contains String.contains]
    [substring String.substring]
    [utf8_bytes String.utf8_bytes]
    [latin1_bytes String.latin1_bytes]
@@ -183,6 +186,11 @@
 (define/method (String.to_number s)
   (check-readable-string who s)
   (string->number s))
+
+(define/method (String.contains s1 s2)
+  (check-readable-string who s1)
+  (check-readable-string who s2)
+  (string-contains? s1 s2))
 
 (define/method String.substring
   #:inline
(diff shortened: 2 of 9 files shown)


mflatt:
Tests are in `rhombus/tests` directory. There's no file or specific unit tests for the `check` form right now, though, so adding `rhombus/tests/check.rhm` would make sense.

ashton314:
Huh… this is a little tricky to test because the thing I'm testing is `check` itself. I'm having difficulty testing that a test fails in the way I want it to. For example, this isn't working:

```
check:
  ~eval
  check: values(1, [2, 3])
         ~matches a
  ~raises "matching a"
```

Can you think of a way I can test failure?

mflatt:
I think you will need to set `Print.current_error` to capture error output, but you'll also need to import from `lib("racket/base.rkt")` for now to be able to create a string port and extract the string.

My guess is that you'll need `~eval` only if you write syntax-error tests.

ashton314:
Ok; I'll give that a whirl

ashton314:
Capturing the output and testing its contents led me down a yak shave that resulted in me adding `Port.open_output_string`, `Port.get_output_string`, and `String.contains`. I've added tests for the last one; I'm not sure if tests for the `Port.*` functions are necessary, as they seem to be little more than aliases to the Racket implementations.

I'm a little worried that this PR got away from me and there's a simpler solution here. Please let me know if that's the case.

usaoc:
Not trying to discourage you, but I think `~matches` already supports multiple values by supplying multiple groups, like

```
#lang rhombus/static
check:
  values(1, 2, 3)
  ~matches: 1; 2; 3
```

mflatt:
I completely forgot that `~matches` supports multiple patterns that way. Adding `values` still seems like a good idea, since it seems like the more expected form.

rhombus/tests/check.rhm
```diff
@@ -0,0 +1,35 @@
...
+  check: values(1, [2, 3])
+         ~matches values(a, [b, c])
+  ~completes
+
+macro 'with_err_to_string:
+         $body ...':
+  'block:
+     def old_port = Port.current_error()
+     def new_port = Port.open_output_string()
+     Port.current_error(new_port)
```
mflatt, line 14: Unless `parameterize` doesn't work for some reason, using `parameterize` is better than setting and then resetting the parameter value.

rhombus/tests/check.rhm
```diff
@@ -0,0 +1,35 @@
...
+     block:
+       $body ...
+     Port.current_error(old_port)
+     Port.get_output_string(new_port)'
+
+check:
+  def the_output = with_err_to_string:
+    check: values(1, [2, 3])
+           ~matches a
```
mflatt, line 23: A nit, but I encourage `def the_output:` over `def the_output =` when the part after `=` is multiple lines.

mflatt (changes requested):
In addition to these small requests/suggestions, the new form and other additions need documentation. That's in `rhombus/scribblings`. I'm happy to help if it's not clear how to change the documentation.

rhombus/private/string.rkt
```diff
@@ -184,6 +187,11 @@
   (check-readable-string who s)
   (string->number s))
 
+(define/method (String.contains s1 s2)
+  #:inline
+  #:primitive (string-contains?)
+  (string-contains? s1 s2))
```
usaoc, line 193: `string-contains?` isn’t a “primitive”, because it’s defined in Racket.  This function will have to include two explicit checks.  (The idea of `#:inline` and `#:primitive` is that primitive functions defined in the kernel can be directly used.  The latter actually sets up the error rewriting, which only applies to actual primitive functions.)

rhombus/private/port.rkt
```diff
@@ -21,7 +21,9 @@
    Output
    [current_input current-input-port]
    [current_output current-output-port]
-   [current_error current-error-port]))
+   [current_error current-error-port]
+   [open_output_string open-output-string]
+   [get_output_string get-output-string]))
```
usaoc, line 26: These functions should ideally use `define/arity`, especially `get_output_string` because Rhombus wants immutable strings.  But I haven’t had the time to look into filling these in yet.

usaoc:
I think it would be better if you avoided adding in new functions *for now*; you could import what you need in the test file and leave a note there.  Actually, adding tests isn’t urgent at all, so don’t feel obliged to do it.

I have another question though: should we remove the multi-group form?  It’s probably better if we just keep the single-group form.  I don’t see the multi-group form used anywhere, either.

mflatt:
I agree that it would be fine to import the needed Racket functions just into the test code, but I also think it's ok to add them to Rhombus. These seem like things that will definitely be needed eventually, even if we decide to adjust names eventually.

rhombus/tests/check.rhm
ashton314: That's much better—I missed that in the docs.

rhombus/private/string.rkt
ashton314: I was following what's going on with `String.length`, since that seemed to be close in form. What should this definition look like? This?

```
(define/method (String.contains s1 s2)
  #:inline
  (string-contains? s1 s2))
```

Or should I drop the `#:inline` too?

rhombus/private/port.rkt
ashton314: I've done a bit on this—I need some help with the `#:static-infos` key though.

rhombus/private/string.rkt
usaoc: It should look more like `String.to_number`, which also doesn’t use a (kernel) primitive.  There isn’t much benefit to inline wrappers of non-primitives (trivial ones *will* get inlined anyway, so we just trust the compiler for those).  You should also include explicit checks, because we want to avoid error reporting in the underlying implementation.

rhombus/private/port.rkt
```diff
@@ -21,10 +22,22 @@
...
 (define-annotation-syntax Input (identifier-annotation #'input-port? #'()))
 (define-annotation-syntax Output (identifier-annotation #'output-port? #'()))
 
 (define-static-info-syntaxes (current-input-port current-output-port current-error-port)
   (#%function-arity 3))
+
+(define/arity Port.open_output_string
+  (case-lambda
+    [() (open-output-string)]
+    [(name) (open-output-string name)]))
```
usaoc, line 38: I think this one should also use `#:inline` and `#:primitive` for the same reason.

rhombus/scribblings/ref-string.scrbl
```diff
@@ -120,6 +121,21 @@ immutable strings.
 
 }
 
+@doc(
+  fun String.contains(str :: ReadableString, substr :: ReadableString) :: Boolean
+){
+
+ Returns whether or not @rhombus(str) contains the string @rhombus(substr)
```
usaoc, line 128: A stylistic issue: predicates are conventionally documented (as in Racket) as “Returns `#true` if \<some condition\>, `#false` otherwise.”  If you think the phrasing “whether” is better, consider changing the verb to “Checks”.  Also note the missing full stop.

rhombus/tests/check.rhm
```diff
@@ -0,0 +1,35 @@
...
+check:
+  check: values(1, [2, 3])
+         ~matches values(a, [b, c])
+  ~completes
+
+macro 'with_err_to_string:
+         $body ...':
+  'block:
+     def new_port = Port.open_output_string()
+     parameterize {Port.current_error : new_port}:
```
usaoc, line 13: (Sorry that I’m being picky here.)  I notice that Matthew favors putting spaces after opening and before ending braces in `parameterize`, so maybe sticking to that convention is better:

```
parameterize { some_param: some_value }:
  ....
````

usaoc:
I think it’ll be good if we can at least merge the main part of this PR (the change to `match`).  If you’re okay with that, I can cherry-pick that commit and merge it; other commits look more tangential, and deserve their own PRs.

ashton314:
Finals done did me in. 😅 That's all over now so I can revisit this. I'm fine if you want to go ahead and just merge it though.

usaoc:
No worries, it’s okay if you’d like to further improve it!  I just think that putting too many things in one PR makes it more difficult to go through.

usaoc:
I cleaned up and refined this PR’s commits at usaoc/rhombus-prototype@cf2e1b6 (and shamelessly added myself as the coauthor).  If you don’t mind, this PR can reset to that commit and I’ll merge them.

PR #414, opened 2023-11-29
fix `Pair.cons` and `Pair.of`

usaoc:
I accidentally exported too much for `Pair.cons`.  Also, it went unnoticed that `Pair.of` in converter mode was broken.

Also add some test cases for `of` annotations in converter mode.


mflatt:
Thanks!

PR #424, opened 2023-12-08
support multiple-result annotations in methods

usaoc:
This aligns the behavior with the documented grammar, which specifies that multiple-result annotations are allowed.


mflatt:
Thanks!

PR #426, opened 2023-12-08
rhombus/draw: fix `Printable` instance of `Color`

usaoc:
Also fix `recur` argument of the `describe` method, which should expect a keyword `~mode` argument defaulting to `#'expr`; clarify the intention in the documentation.

### DIFF
--- a/rhombus/draw/private/color.rhm
+++ b/rhombus/draw/private/color.rhm
@@ -38,13 +38,14 @@ class Color(handle):
   reconstructor(red, green, blue, alpha):
     Color(red, green, blue, alpha)
 
-  implements Printable
-  override describe(mode, recur):
-    "Color("
-      +& red +& ", "
-      +& green +& ", "
-      +& blue +& ", "
-      +& alpha +& ")"
+  private implements Printable
+  private override describe(mode, recur):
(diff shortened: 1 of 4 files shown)


mflatt:
Fixing the printing of `Color` is good, and fixing `recur` to have a `~mode` keyword is also good.

But I think the default for `~mode` should stay `#'expr`, not `#'string`, even though that's different than the default for `Printable.describe`. The reason is that a `describe` method almost always wants `#'expr` for its components, as illustrated by `Color.describe`. More generally, it's because the current Rhombus convention is to have only a few values (like strings) print specially in `#'string` mode, and most print the same in both modes — where "the same" means that after switching to `#'expr` style, the style stays that way for nested components. (This is a different choice than Racket, where the convention is to recur in the same mode.)

Meanwhile, `Printable.describe` has the `#'string` default to be consistent with `print`. I tried various combinations here, and I found that I generally wanted `print(x)` to print string content when `x` is a string, especially when `x` is constructed with `+&`.

usaoc:
I reverted the default and added some explanation in the docs.

mflatt:
Thanks!

PR #428, opened 2023-12-10
shrubbery: fix a type error of `current-classify-range`

usaoc:
Reported by John Clements.  I’m not sure how to test this.


mflatt:
I can't remember enough to suggest a test, either, but this seems clearly an improvement. Thanks!

PR #431, opened 2023-12-10
add another example of a reducer macro, more explanatory text

jbclements:
I wanted to add a more "user-friendly" reducer macro example that shows a not-trivial, not-wizard-mode reducer, and exercises a reasonable number of reducer features. It would have been nice to have an example that uses more than one accumulator variable, to show the `values` in the `step_result`, but ... I timed out, and I think this is an improvement on the existing docs.


mflatt:
Thanks!

PR #433, opened 2023-12-11
Add disappeared-use for fields.

samth:


mflatt:
I'm not clear on why the disappeared identifier here is the `new-field-id` instead of `field-id` (but with the source locations of `field-id`). Can you say more about what example this fixes?

greghendershott:
The motivation is enabling check-syntax to draw arrows.

Here's a link directly into a long thread, where @samth [first mentions this PR](https://racket.discourse.group/t/imports-as-completion-candidates-for-hash-langs/2511/20).

(Later in the thread are some examples using `#lang rhombus` per se. And @rfindler also makes a change to check-syntax. The combination seems to work.)

samth:
I think @mflatt is just wondering why the identifier in the property is not the obvious one. I think the more obvious choice doesn't work but I haven't had a chance to go back and check.

mflatt:
Thanks — After looking at the example in the disucssion, I think I understand now why `new-field-id` is needed.

mflatt:
After thinking about this more, I pushed a refinement as 03c1cb0412619b33b397c33374a914c13d634335. Using `new-field-id` directly worked for `math.max` only because there are no extra scopes on the introduced `max`, but it's really the originalness of `field-id` that we want to preserve.

For example, with the change, `f.max` gets a documentation link:

```
#lang rhombus

namespace f:
  export:
    math.max

f.max
```

The documentation link is wrong for both `math.max` and `f.max` — it goes to racket instead of Rhombus docs — but that's a different issue.

PR #438, opened 2023-12-14
allow single-group body in `meta`

usaoc:
This form is similar to `define-for-syntax` in Racket, but more general.  In particular, it avoids the extra indentation as compared to the block form.


mflatt:
Very nice.

PR #444, opened 2023-12-19
add `satisfying` annotation

usaoc:
This form can be useful for turning an existing predicate into annotation without using `annot.macro` and friends, especially when defining new annotation isn’t worth it.  I also think that it can be used to document Racket interops, because Racket modules generally export many predicates as flat contracts.

### DIFF
--- a/rhombus/private/annotation.rkt
+++ b/rhombus/private/annotation.rkt
@@ -52,6 +52,7 @@
                     False
 
                     matching
+                    satisfying
                     #%parens
                     #%literal)
          (for-spaces (rhombus/annot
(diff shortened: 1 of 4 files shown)


rhombus/private/annotation.rkt
```diff
@@ -864,6 +865,31 @@
...
+   (annot-quote satisfying)
+   '((default . stronger))
+   'macro
+   (lambda (stx)
+     (syntax-parse stx
+       [(_ (_::parens pred-g) . tail)
+        (values
+         (annotation-predicate-form
+          ;; use `((lambda ....) ....)` to avoid inferred name
+          #'((lambda (pred)
```
mflatt, line 879: I guess `(lambda (pred) ...)` is inlined here to encourage optimization when `pred` is obviously a procedure of arity 1?

mflatt (commented):
LGTM

rhombus/private/annotation.rkt
usaoc: Yes, that was the intention.

PR #447, opened 2023-12-20
more consistent `{List,(Readable){Map,Set}}.empty`

usaoc:
Change the expression forms to simple variables with appropriate static information, and associate static information in binding forms. An idiosyncrasy of `Readable{Map,Set}.empty` is that the static information is *less* than directly using the constructor (and you can’t construct a “readable {map,set}”, apparently).

### DIFF
--- a/rhombus/private/list.rkt
+++ b/rhombus/private/list.rkt
@@ -212,6 +212,9 @@
        (lambda (v) (cons v lst))
        (lambda () #f))))
 
+(define-static-info-syntax null
+  #:defined list-static-infos)
+
 (define-binding-syntax null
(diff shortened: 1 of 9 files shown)


usaoc:
I mainly want to confirm that changing the map and set variants to variables is okay.  I guess they are currently expressions macros so that they behave the same as the constructors?  But I’m not sure that something to be expected.

mflatt:
I think `Map.empty`, etc., were probably added as expression macros because it was at the same time as adding binding forms. I don't see any reason it should matter, though, and it does seem better to be consistent with `List.empty`.

PR #448, opened 2023-12-22
enforest: improve precedence error message

usaoc:


usaoc:
@mflatt Do you have any opinion on this?  I wanted to make sure that the message read well.

PR #450, opened 2023-12-23
fix `#%literal` bindings when `case` expansion applies

### LINKED ISSUE #417: Surprising behavior of literal bindings when `case` expansion applies
Racket `case` follows `equal?` semantics, while Rhombus wants `equal-always?` semantics.  This leads to some corner cases being broken:

```
#lang rhombus/static
import:
  lib("racket/base.rkt").#{string-copy}

match #{string-copy}("abc")
| "abc": "matched"
| ~else: "unmatched"
```

(And I realized there was another bug that broke the `case` expansion when all patterns are literal.  Welp, let’s fix that first.)

Should we add different versions of `case` to Racket?  I’m not sure how I feel about that.

usaoc:
This uses the change in racket/racket#4856 for `case` expansion, which requires a very recent development version of Racket.

Fix #417.

### DIFF
--- a/rhombus/private/match.rkt
+++ b/rhombus/private/match.rkt
@@ -14,7 +14,13 @@
          "realm.rkt"
          "parens.rkt"
          (submod "quasiquote.rkt" for-match)
-         (only-in "literal.rkt" literal-infoer))
+         (only-in "literal.rkt" literal-infoer)
+         "version-case.rkt")
+;; TEMP approximate `case/equal-always`
+(meta-if-version-at-least
+ "8.11.1.8"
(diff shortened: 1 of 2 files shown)


usaoc:
Opening this as a reminder, but as the commit message says, this change requires a very recent development version of Racket, so I think we should wait for a while before merging this.

mflatt:
How about using `meta-if-version-at-least` to work (as well as before) in older Racket versions? I'm not sure how long we can or should support old versions, but maybe we should keep trying as long as it's easy enough.

PR #451, opened 2023-12-28
remove key-specific static-info indirections

mflatt:
There's a generic `#%indirect-static-info` key that can be used to splice static information from another identifier. There were also use-specific keys like `#%append-indirect`. The use-specific ones were a misguided attempt to avoid collisions when merging and replacing from different sources, and they actually have the effect of breaking merges, because searching for two keys can mean that the first one you look for is found even though it appears after the second one you would look for. So, this commit removes the use-specific indirections.

For function arity information, the indirection served a second purpose: an arity adjustment to account for the implementation as a method that expects an extra argument. A lot of this commit is detangling and removing that role of the indirection.


usaoc:
LGTM.

PR #452, opened 2023-12-29
add `Array.of_length` annotation and reducer

usaoc:
Mainly to replace the `~length` form of `Array`.  My rationale is that reducer macros should consume a finite number of terms in the presence of the `each` shorthand; the existing parsing hack makes me very uncomfortable.  Having a grouping in hand also makes it possible to add a `~fill` option, as in `for/vector`.

### DIFF
--- a/rhombus/private/array.rkt
+++ b/rhombus/private/array.rkt
@@ -55,7 +55,8 @@
    [append Array.append]
    [of now_of] ;; TEMPORARY
    now_of
-   later_of)
+   later_of
+   of_length)
   #:properties
   ()
   #:methods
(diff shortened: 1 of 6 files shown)


usaoc:
I don’t know what’s the better choice—we only have so many kinds of brackets.  I find this choice fine due to the parallel to the annotation form, but maybe someone will have a better idea.

mflatt:
In working toward #455, I ran into confusion with `for Array ~length sizes.length()` when modifying with a `for` that was already written with `each`. So, I lean more in favor of this change than I did before. I still worry about the "of" convention, but having `Array.of_length` also be an annotation constructor (as in the latest version) seems like a good step.

usaoc:
Is there any further objection to this from someone else?  If not, I’ll merge this later.

mflatt:
Since this will break existing programs, let's wait just a little to combine with other breaking changes.

PR #453, opened 2023-12-30
add static class veneers

mflatt:
The `veneer` form is similar to `class`, but it uses an existing representation instead of creating a new one. Since there's no new representation, a veneer affects only static dispatch and not dynamic dispatch.

For example, suppose you have the following `Posn` definition:

```
class Posn(x, y):
  private implements: Indexable
  private override get(i):
    match i
    | 0: x
    | 1: y
```

You've been using `static` mode to ensure the best available performance, but it turns out that your application creates about a billion posns. It could be worth using raw pairs, which will take up half the space, but you don't want to change all your code that says `Posn`, `x`, and `y` and make it less clear by writing `Pair`, `first`, and `rest`.

With `veneer`, you can replace the `Posn` declaration with

```
veneer Posn(this :~ Pair):
  property x:
    this.first
  property y:
    this.rest
  private implements: Indexable
  private override get(i):
    match i
    | 0: x
    | 1: y
  expression 'Posn($x, $y)':
    'Pair($x, $y) :~ Posn'
```

and a use like

```
  def p = Posn(1, 2)
  p.x + p[1]
```

will produce the same value as before. From the outside of this new `Posn`, it's a class with `x` and `y` properties that is indexable. Inside `Posn`, `this` is a `Pair`, so `this.first` statically selects the same as `Pair.first`.

More example interactions:

```
> def p = Posn(10, 20)
> p.x
10
> p[1]
20
> p.first
first: no such field or method (based on static information)
> block:
    use_dynamic
    p.first
10
> dynamic(p).x
x: no such field or method (based on static information)
```

The actual motivating example involves interior `Node`s in an RRB tree implementation, where a `Node` class has an `Array` and `maybe(Array)`, which can be more compactly represented as `Array || Pair.of(Array, Array)`. The savings there are in both space and time, since interior nodes are created by update operations.

Issues:

 * Maybe there's a better existing technical term to use instead of "veneer".
 * Rhombus has some things that work only with static dispatch, but `veneer` seems like a significant escalation.
 * The `this` pseudo-field is a pun that makes `veneer` closer in syntax to `class` while also highlighting the static information that's on `this` inside the veneer. The `::` or `:~` after `this` provides a place to indicate whether the underlying annotation is checked.
 *  A veneer can work with a converter annotation (using `::` and not `:~`), but you have to declare that intent explicitly with a `converter` veneer clause. The conversion is currently applied on every method call, not just the initial "entry" into the veneer's annotation, mostly because it's not clear how to avoid that check and also avoid duplicating code for methods.

### DIFF
--- a/rhombus/private/annotation.rkt
+++ b/rhombus/private/annotation.rkt
@@ -335,18 +335,20 @@
                                   sub-n kws predicate-maker info-maker
                                   binding-maker-id binding-maker-data
                                   parse-annotation-of)
+    (define root
+      (annotation-prefix-operator
+       name
+       '((default . stronger))
+       'macro
+       (lambda (stx)
+         (syntax-parse stx
+           [(_ . tail)
+            (values (annotation-predicate-form predicate-stx
+                                               static-infos)
+                    #'tail)]))))
     (values
      ;; root
-     (annotation-prefix-operator
-      name
-      '((default . stronger))
-      'macro
-      (lambda (stx)
-        (syntax-parse stx
-          [(_ . tail)
-           (values (annotation-predicate-form predicate-stx
-                                              static-infos)
-                   #'tail)])))
+     root
      ;; `of`:
      (annotation-prefix-operator
       name
@@ -385,10 +387,15 @@
         (~optional (~seq #:extends name-extends)
                    #:defaults ([name-extends #'#f])))
      #:with annot-name (in-annotation-space #'name)
+     (define sub-n-val
+       (syntax-parse #'sub-n
+         [(quote sub-n) (syntax-e #'sub-n)]
+         [_ (syntax-e #'sub-n)]))
+     (define extra-names (list #'of-name))
      (define defs
        ;; usually `(define-syntaxes (annot-name of-name) ....)`:
        (build-syntax-definitions/maybe-extension
-        (list 'rhombus/annot) #'name #:extra-names (list #'of-name) #'name-extends
+        (list 'rhombus/annot) #'name #:extra-names extra-names #'name-extends
         #'(let binds
             (annotation-constructor #'annot-name predicate-stx static-infos
                                     sub-n 'kws
--- a/rhombus/private/appendable.rkt
+++ b/rhombus/private/appendable.rkt
@@ -51,22 +51,23 @@
       (bytes? v)))
 
 (define-class-desc-syntax Appendable
-  (interface-desc #'Appendable
-                  #'Appendable
-                  #'()
-                  #'prop:Appendable
-                  #'prop:Appendable
-                  #'Appendable-ref
+  (interface-desc #'()
                   '#(#&append)
                   #'#(#:abstract)
                   (hasheq 'append 0)
                   #hasheq()
-                  #t
                   '()
                   #f
                   #'()
-                  #f
-                  '(append)))
+                  '(append veneer)
+                  ;; --------------------
+                  #'Appendable
+                  #'Appendable
+                  #'prop:Appendable
+                  #'prop:Appendable
+                  #'Appendable-ref
+                  #t
+                  #f))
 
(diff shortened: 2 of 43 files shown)


samdphillips:
This seems similar in spirit to the "static class view" @jackfirth proposed.  In that case, we discussed ways to make existing Racket values have a more Rhombus interface without having a lot of boilerplate of `namespace` and `annot` macros.

samth:
There's a way to do basically this in Common Lisp, which we should probably look at to see how different it is. I think that works dynamically, but I don't know much more.

mflatt:
The CL feature (thanks @samdphillips and @samth) is `:type` for `defstruct`: https://www.cs.cmu.edu/Groups/AI/html/cltl/clm/node173.html#SECTION002350000000000000000. That's similar in that it overlaps CL's `deftype` machinery on a use of lists or vectors specifically, but it's not about adapting an arbitrary datatype.

The CL analogy does suggest making `veneer` a mode of `class` instead of a new syntactic form. In fact, I started out that way, but then there are many caveats about which class-clause forms work in veneer mode and which don't, and I found it clearer and cleaner to organize that with a separate `veneer` form and veneer clauses.

mflatt:
@samdphillips: I'm not sure which discussion you have in mind, but I agree that we've considered this kind of thing before. There are also two things we've considered that this `veneer` proposal *dosn't* address:

 * It's not a way of adding methods to an existing class. For example, if `Posn` exists and you want to add a `flip` method, we can imaging allowing `method` as a definition form as in `method Posn.flip(): ...`, which would create a static dispatch target for `(p :: Posn).flip()` when the definition is visible. That's still something we can consider doing, but it's different from `veneer`, where the intent is to add a new annotation.
 * It's not a general way of adapting existing Racket datatypes to Rhombus without a wrapper. Currently, wrapperless adaptations have to be built into the `rhombus` base language, and @usaoc has improved syntactic support to do that internally. We could imagine having something like a `dynamic` veneer clause that supports more of the same things as the internal macros; the problem is that there's ultimately a Rhombus-global effect to make the adaptation work, and that's not safe or modular. It might be possible to recover modularity through a table keyed by namespace registry and phase, but then there are still all sorts of issues with namespace crossing and effect order. After trying to think through that direction, my conclusion is still that wrapper classes are the way to go for most adaptations.

samth:
Which pieces of class don't work for veneer?

I think the biggest other question raised by the CL approach is whether to make veneer static-only as in this PR, or to support more dynamic behavior (it's not clear to me from the docs exactly how that works in CL).

samdphillips:
> I think the biggest other question raised by the CL approach is whether to make veneer static-only as in this PR, or to support more dynamic behavior (it's not clear to me from the docs exactly how that works in CL).

AIUI CL only works dynamically if the value is tagged (using the `:named` option) otherwise it looks like whatever the base type is with some accessor functions.

mflatt:
Class clauses that do not make sense for veneers: `field`, `abstract`, `reconstructor`, `opaque`, `prefab`, and `nonfinal`. Class clauses that don't currently have a `veneer` analog: `internal`, `constructor`, `binding`, and `annotation`; those could make sense, but they did not seem useful enough for a first cut. Distinguishing class and veneer clauses also reduces the burden for writing a new class clause, since it does not have to explicitly check for a veneer context (while a clause that makes sense for veneers must be explicitly written for both, similar to supporting both classes and interfaces).

I don't see that CL's `defstruct` plus `:type` says anything dynamic dispatch. CL's generics and methods are relevant, but my second bullet [above](https://github.com/racket/rhombus-prototype/pull/453#issuecomment-1872965466) is more about that.

mflatt:
We discussed this some in the 11-JAN-2023 meeting. While no one seems really sure that this is a good idea, no one seems sure that it's a bad idea, and it does seem useful. (I've encountered a second application.) I'll clean up the implementation and merge, and we can see whether we find enough uses that pan out longer-term.

PR #454, opened 2024-01-01
add support for annotations in `values` reducer

### LINKED ISSUE #449: `values` reducer incorerctly propagates initial-value static info
The following example is currently allowed statically:

```
#lang rhombus/static

class Posn(x, y)

for values(p = Posn(1, 2)) (i: 0..2):
  p.x
```

The problem is that the `values` reducer incorrectly transfers static info from `Posn(1, 2)` to the binding `p`. I introduced this bug in eeb362d4, likely in an attempt to make the loop in `gui_demo.rhm` nicer. Unfortunately, other Rhombus code likely relies on this incorrect inference.

A possible change is allow a binding form on the left-hand side of `=`, instead of allowing only an identifier, and then you can at least write `values(p :: Posn = Posn(1, 2)`. That would be more similar to `recur`. I don't have an idea that would allow a `:: Posn` to be omitted.

usaoc:
Instead of unsound propagation of static infos from the initial expressions, use static infos from inline annotations, which can be unchecked as usual (and the user is responsible for that).

Also add support for
- `_` in place of id;
- block-style initial expressions;
- propagation of static infos for results.

Fix #449.

### DIFF
--- a/rhombus/scribblings/ref-values.scrbl
+++ b/rhombus/scribblings/ref-values.scrbl
@@ -57,13 +57,24 @@
 }
 
 @doc(
-  reducer.macro 'values($id = $expr, ...)'
+  reducer.macro 'values($id $maybe_annot $init, ...)'
+
+  grammar maybe_annot:
+    #,(@rhombus(::, ~bind)) $annot
+    #,(@rhombus(:~, ~bind)) $annot
+    #,(epsilon)
+
+  grammar init:
+    = $expr
+    : $body; ...
 ){
 
  A @tech{reducer} used with @rhombus(for), expects as many results from a
- @rhombus(for) body as @rhombus(id)s. For the first iteration of
+ @rhombus(for) body as @rhombus(id)s. When @rhombus(id) is
+ @rhombus(_, ~bind), a fresh identifier is used, otherwise
+ @rhombus(id) is bound as follows. For the first iteration of
  the @rhombus(for) body, each @rhombus(id)'s value is the result
- of the corresponding @rhombus(expr). The results of a @rhombus(for) body
+ of the corresponding @rhombus(init). The results of a @rhombus(for) body
  for one iteration then serve as the values of the @rhombus(id)s
  for the next iteration. The values of the whole @rhombus(for) expression
  are the final values of the @rhombus(id)s.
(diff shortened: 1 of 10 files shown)


rhombus/scribblings/ref-values.scrbl
```diff
@@ -57,13 +57,24 @@
...
-  reducer.macro 'values($id = $expr, ...)'
+  reducer.macro 'values($id $maybe_annot $expr_or_block, ...)'
+
+  grammar maybe_annot:
+    :: $annot
+    :~ $annot
+    #,(epsilon)
+
+  grammar expr_or_block:
+    $expr
```
mflatt, line 68: Should have an `=` here? I would call this non-terminal `init`, or something like that.

mflatt:
I don't think we way of attaching static info to the accumulator identifiers is right. For example, I don't think it will behave correctly if the identifier is shadowed by an `each` clause. More generally, I think this reflects the same problem as not being able to use static information in clauses like `keep_when` or in expressions/definitions before for clauses.

It looks like the reducer protocol needs to change, instead — probably to accept forms to place at the beginning of the initial `#:splice` after `for/fold f.binds` in the expansion. I think that approach would work with the improved `for/fold` of v8.11.1.3, at least, but I might be missing something. (I think it would not have worked before the `for/fold` corrections of v8.11.1.3, but I'm not certain.)

usaoc:
I think that extension will also allow arbitrary binding forms, which is more general than what this PR does.

usaoc:
Actually, this also reminds me of another thing (more tangential to this specific change): will it make sense to extend the protocol to also support right folds (`for/foldr`)?

mflatt:
When you suggest supporting right folds, do you mean that something more is needed to allow a reduce to be used in either a left-fold or right fold, or that the reducer should pick?

A reducer should be usable in either a right or left fold, and I don't know offhand why it wouldn't already work  — but maybe you can say why.

I don't think a reducer should pick the direction itself. A reducer is intended to be used for future forms, particularly a transducer form, and I don't think it will work in other contexts for the reducer to pick.

mflatt:
Rebased in https://github.com/mflatt/rhombus-prototype/tree/values-ann. If if looks like I got that right, we can reset this PR's branch to that commit.

usaoc:
That branch’s history seemed mangled, so I did a different rebase.

PR #455, opened 2024-01-01
RRB trees as lists

mflatt:
This is an update on @OxSon's work on implementing RRB trees for Rhombus and the changes that it inspired in Racket to arrive at better performance.

With an eye toward trying RRB trees in place of pair-based `List` (and with the thought of avoiding bootstrapping complexity), I ported the Rhombus implementation to Racket. The branch for this PR has both versions: `rhombus/private/treelist.rhm` and `rhombus/private/treelist.rkt`. Comparing Racket performance to Rhombus performance exposed issues in Rhombus and Racket that are now fixed. The Racket implementation is still faster, because it has access to things like `vector*-set/copy` and `fx+`; adding that access to Rhombus is straightforward, but maybe not needed for now. Both the Racket and Rhombus implementations are measured as compiled in unsafe mode.

Measurements below use a [microbenchmark suite](https://gist.github.com/mflatt/5c1541e479550a7946b974bae197fc02) that creates lists of size 10.

Here's the output with the latest Racket, where "TreeList" is the Rhombus implementation and "treelist" is a Racket port of that code, and "add" builds by adding to the end while "cons" builds by adding to the front. The "Set" lines are included as a comparison to a HAMT, since it should have related performance characteristics in these tasks.

```
 Pair.cons       cpu time: 20 real time: 21 gc time: 3
 List.cons       cpu time: 86 real time: 89 gc time: 0
 Set.add         cpu time: 237 real time: 246 gc time: 1
 TreeList.add    cpu time: 120 real time: 124 gc time: 1
 treelist-add    cpu time: 80 real time: 82 gc time: 1
 TreeList.cons   cpu time: 122 real time: 126 gc time: 1
 treelist-cons   cpu time: 78 real time: 80 gc time: 0
 Pair.rest       cpu time: 15 real time: 17 gc time: 0
 List.rest       cpu time: 88 real time: 91 gc time: 0
 Set.remove      cpu time: 320 real time: 333 gc time: 1
 TreeList.drop   cpu time: 280 real time: 290 gc time: 0
 treelist-drop   cpu time: 93 real time: 96 gc time: 0
```

Note that RRB tree construction can now be *faster* than `List.cons`, because `List.cons` has a `list?` check on its second argument. Constructing an RRB tree is within a factor of 4 of `Pair.cons`, which I think is pretty good.

The gap widens when creating 1/10th as many lists of size 100, but it's still reasonable:

```
Pair.cons       cpu time: 14 real time: 15 gc time: 3
List.cons       cpu time: 78 real time: 81 gc time: 0
Set.add         cpu time: 463 real time: 474 gc time: 2
TreeList.add    cpu time: 304 real time: 312 gc time: 2
treelist-add    cpu time: 152 real time: 155 gc time: 2
TreeList.cons   cpu time: 314 real time: 322 gc time: 3
treelist-cons   cpu time: 180 real time: 184 gc time: 2
Pair.rest       cpu time: 10 real time: 12 gc time: 0
List.rest       cpu time: 76 real time: 80 gc time: 0
Set.remove      cpu time: 506 real time: 520 gc time: 2
TreeList.drop   cpu time: 664 real time: 695 gc time: 3
treelist-drop   cpu time: 210 real time: 215 gc time: 2
```

Keeping just the RRB lines for 10-element lists for easier comparison to older Racket:

```
 TreeList.add    cpu time: 120 real time: 124 gc time: 1
 treelist-add    cpu time: 80 real time: 82 gc time: 1
 TreeList.cons   cpu time: 122 real time: 126 gc time: 1
 treelist-cons   cpu time: 78 real time: 80 gc time: 0
 TreeList.drop   cpu time: 280 real time: 290 gc time: 0
 treelist-drop   cpu time: 93 real time: 96 gc time: 0
```

Output with v8.9 (and I expect the same for v8.11):

```
 TreeList.add    cpu time: 859 real time: 891 gc time: 2
 treelist-add    cpu time: 583 real time: 610 gc time: 2
 TreeList.cons   cpu time: 870 real time: 903 gc time: 2
 treelist-cons   cpu time: 562 real time: 589 gc time: 2
 TreeList.drop   cpu time: 899 real time: 923 gc time: 1
 treelist-drop   cpu time: 317 real time: 329 gc time: 1
```

So, Racket improvements (for functional vector operators, mostly) are worth a factor of 7 or so.

Those microbenchmarks are meant to be tasks that favor pair lists. A [random list generator](https://gist.github.com/mflatt/3d64a4bdb699273910193a4282d48b37) helped with testing, where it constructs a list of roughly 1000 elements by randomly picking where to insert into the list and occasionally appending the list to itself — and doing that 1000 times. By turning only one mode on at a time, it provides a rough performance number:

```
List       cpu time: 4028 real time: 4160 gc time: 33
TreeList   cpu time: 1507 real time: 1563 gc time: 35
treelist   cpu time: 875 real time: 925 gc time: 35
```


mflatt:
The latest commit changes `List` to the treelist implementation, while `PairList` is Racket's list datatype.

Beware that documentation doesn't build, because `scribble/rhombus` needs a lot of changes to adapt.

Some status points and thoughts on the conversion, so far:
 
 * Overall, I think this is probably a good idea. Getting fast enough performance has been challenging, and I expect there will be a long tail of performance issues going forward. But log-time complexity for length, insert (front, end, or middle), append, sublist, and random-access operations will help programmers avoid performance pitfalls on a larger scale.

 * One complication in performance comparisons is a difference in reachablity. When you use `for` to walk through a Racket list, earlier items in the list can become unreachable as later items are process. With a treelist, iteration retrieves each element using an index (roughly), so the whole list stays live. That's fine for many purposes, and it can provide a performance benefit: iterating that way (instead of a `cdr`-like operation on `treelist`) halves the performance difference compared to a pair-list traversal.

 * The treelist implementation has faster equality checking that can detect internal sharing, which means that it can be asymptotically faster. On lists of size 6 or more (roughly), treelist comparison tents to be faster than list comparison, even without sharing, possibly because it doesn't have to detect sharing at the pair level.

 * `for/treelist` can build a tree list faster than `for/fold` using `treelist-add` (and, unlike `for/list`, it doesn't involve a `reverse` operation to complete). If you already have a list, then `list->treelist` is probably even faster, and if you have a vector, then `vector->treelist` is certainly faster; the `list->treelist` and `vector->treelist` functions directly construct a tree of nodes for a treelist.

 * The treelist implementation now supports chaperones. Supporting chaperones seems to impose a cost of 10% or so on non-chaperoned operations, but I think that's probably the right tradeoff.

 * Benchmark code is now in `design/treelist`, along with the original Rhombus implementation of `TreeList`.

 * The treelist implementation should be usable in Racket without Rhombus. I imagine putting it in `racket/treelist`, partly so that `racket/pretty` and `racket/dict` can depend on it instead of the other way around, and partly because I can imagine the treelist data structure becoming even more primitive so that lower levels can use it. For now, `racket/treelist` seems like the right layer. Is "treelist" a good name for the Racket view?

 * To check the performance effect of using treelists on programs written for traditional Lisp lists, I tried running a Shplait lambda-calculus interpreter on `fib(25)` (using the Y combinator for recursion). That creates and searches a lot of environments using `cons` and `cdr`. The effect was significant, making the program take almost twice as long — but there are some complicating factors in the case of Shplait. After porting to Rhombus as `rhombus/tests/example-interp-ast.rhm`, I see a more modest 20% hit. So, probably ok. Compilation time for Shplait programs seems unaffected. Then again, to support lazy mode, it turns out that Shplait can't use Rhombus lists and needs to use pairs. Furthermore, to control printing, it turns out to need its own pairs (that adaptation is https://github.com/mflatt/shplait/tree/treelist); as a result, performance is about the same as before.

 * Rhombus currently allows either `PairList`s or `List`s in some places, such as when constructing or splicing into syntax objects. The union of those is currently called `Listable`, and I imagine making that an interface that other classes can implement it. Probably `Array` should implement `Listable`, too. The only operation I imagine on `Listable` is `to_list`.

 * I think a `MutableList` class would make sense, and I imagine implementing it as a wrapper around a treelist with a mixture of imperative and copying operations on that treelist. (There's a larger discussion about the interplay of functional and imperative data structures that I'll leave to elsewhere.)

 * The fact that `scribble/rhombus` is currently broken demonstrates how making Rhombus lists different from Racket lists will create friction. The last step in this experiment for me is to see how that conversion plays out, but this will be an issue that never goes away.

mflatt:
Here's an updated performance table — using labels like `PairList` that match the implementation, and leaving out the initial `TreeList` implementation that was written in Rhombus. The `+`, `-`, or `^` character at the front just groups related measurements.

```
+ Pair.cons     cpu time: 19 real time: 20 gc time: 3
+ PairList.for  cpu time: 47 real time: 50 gc time: 0
+ PairList.cons cpu time: 84 real time: 87 gc time: 0
+ List.for      cpu time: 78 real time: 80 gc time: 0
+ List.add      cpu time: 94 real time: 97 gc time: 0
+ List.cons     cpu time: 96 real time: 98 gc time: 0
+ Set.add       cpu time: 241 real time: 249 gc time: 1
- Pair.rest     cpu time: 14 real time: 15 gc time: 0
- PairList.rest cpu time: 86 real time: 89 gc time: 0
- List.rest     cpu time: 146 real time: 151 gc time: 0
- Set.remove    cpu time: 322 real time: 335 gc time: 1
^ PairList.iter cpu time: 10 real time: 12 gc time: 0
^ List.iter     cpu time: 38 real time: 39 gc time: 0
^ Set.iter      cpu time: 87 real time: 92 gc time: 0
```

rhombus/private/treelist.rkt
```diff
@@ -0,0 +1,1077 @@
...
+      [[(d) (_ tl-expr)]
+       #'[(d)
+          (:do-in
+           ([(tl) tl-expr])
+           (unless (variable-reference-from-unsafe? (#%variable-reference))
+             (check-treelist 'in-treelist tl))
+           ([pos 0]
+            [node empty-node]
+            [node-pos 0])
+           (pos . fx< . (treelist-size tl))
```
samth, line 190: Do we want to record the size instead of computing it every time?

rhombus/private/treelist.rkt
```diff
@@ -0,0 +1,1077 @@
...
+     (define-values (root height)
+       (cond
+         [(size . fx<= . MAX_WIDTH) (values (Node vec) 0)]
+         [else
+          (define height (fx- (fxquotient (fx+ (integer-length (fx- size 1)) BITS -1) BITS) 1))
+          (values
+           (let loop ([start 0] [height height])
+             (cond
+               [(fx= height 0)
+                (Node (vector*-copy vec start (fxmin size (fx+ start MAX_WIDTH))))]
```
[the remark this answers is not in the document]
mflatt: Yes, good point. I'll fix this.

rhombus/private/treelist.rkt
```diff
@@ -0,0 +1,1077 @@
...
+(define (leaf v) (Node (vector v)))
+
+(struct treelist (root size height)
+  #:property prop:equal+hash (list
+                              (lambda (tl other-tl recur)
+                                (treelist-equal? tl other-tl recur))
+                              (lambda (v recur)
+                                (treelist-hash-code v recur))
+                              (lambda (v recur)
+                                42))
```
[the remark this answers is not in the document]
mflatt: It's not used for CS, but I agree that using the size would be a good choice, so I'll change that.

rhombus/private/treelist.rkt
```diff
@@ -0,0 +1,1077 @@
...
+            (node*-set node branch-index (set (node*-ref node branch-index) index el (fx- height 1)))]
+           [else
+            (define-values (branch-index subindex) (step node index height))
+            (node-set node branch-index (set (node-ref node branch-index) subindex el (fx- height 1)))])))
+     (treelist new-node size height)]))
+
+;; add `el` to end of vector
+(define (treelist-add tl el)
+  (cond
+    [(impersonator? tl)
```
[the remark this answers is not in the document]
mflatt: There's a `(fx= size 0)` check as a shortcut a little later. I'm not sure it would be faster to check against `empty-node`, since `empty-node` isn't a compile-time constant, so it has to be fetched at run time.

rhombus/private/treelist.rkt
```diff
@@ -0,0 +1,1077 @@
...
+                                                    #:result (if (fx= size 0)
+                                                                 empty-treelist
+                                                                 (treelist root size height)))
+                           binds
+           pre-body ...
+           (unsafe-root-add root size height
+                            (let ()
+                              post-body ...))))]))
+
+(define-syntax (for*/treelist stx)
```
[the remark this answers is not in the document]
mflatt: It's not clear that it would be faster, and it would definitely cooperate less with `call/cc`, so I think it's probably better to not do that.

rhombus/private/treelist.rkt
mflatt: That's probably better, but I'm not certain. The `treelist-size` function is just a field accessor, and before treelists supported chaperones, it would not be obviously slower than a closure lookup or loop-carried variable (in addition to `tl`, which must still be kept). But now that treelissts support chaperones, it's more likely faster to extract the size once.

rhombus/private/treelist.rkt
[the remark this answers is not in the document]
mflatt: Checking `(eq? tl empty-node)` makes the benchmark slower on my machine.

rhombus/private/treelist.rkt
mflatt: This does not make enough of a difference to measure reliably on my machine, but since it doesn't hurt, I'll lift out the size lookup.

sorawee:
Would `ListLike` work better than `Listable`? The `able` suffix makes sense when the prefix is a verb.

mflatt:
The conversion is complete, including documentation updates.

Some notes on how lists, pair lists, and listables turned out in this proposal/implementation:

 * `List` means a list implemented with an RRB tree so that insertion, append, removal, and sublist operations are O(log N).
 * `PairList` means a Racket list. For example, `PairList[1, 2, 3]` is a Rhombus pair list and Racket list, while `[1, 2, 3]` is a Rhombus list and a Racket treelist.
 * `Listable` is something that supports a conversion to lists. Listables includes lists, pair lists, arrays, range sequences like `0..10`, and any class that implements `Listable` by overriding `to_list`.
 * Listable, indexable, and sequences are all different things and do not imply each other, although they often go together. For example, `0..10` is a listable sequence, while `0..` is a sequence that is not listable.
 * Explicit splicing positions generally allow listables instead of requiring a list. That's convenient for working with different kinds of listables, and it also turns out to be useful for converting between them. For example, `[& 0..10]` creates a list from 0 to 9.
 * A list pattern does *not* match a listable. A list pattern matches a list, a pair list pattern matches a pair list, and so on. The intent is to avoid accidental expensive conversions. Meanwhile, splicing operations seem explicit enough to allow the convenience of conversions. Where you do want conversions, we can have converting annotations.
 * Converting annotations are used pervasively to implement `scribble/rhombus` on top of `scribble`. Scribble "content" is defined to allow nested lists, and so content built with `List`s need to be converted to content built with `PairList`s. This worked out nicely, and most of the effort was building up `scribble/rhombus` as a proper Rhombus layer of `scribble`, instead of the previous hacky mixture of adapted and unadapted functions.

samth:
I think the one thing I'm not sure of is where we left `List` genericity wrt this change. `Listable` is the common interface but it's very thin intentionally and I don't remember if that's something we decided on or something to revisit later or a consequence of these changes that you and @OxSon learned about while building this.

design/treelist/bm.rhm
```diff
@@ -0,0 +1,133 @@
...
+  'block:
+     print($who +& "                ".substring(0, math.max(0, 16 - ($who).length())))
+     measure.time:
+       ~gc
+       $body
+     #void' 
+
+bm "+ Pair.cons":
+  for values(r = #false) (i: 0..M):
+    for values (l = []) (i: 0..N):
```
usaoc, line 21: Style: space after `values`

design/treelist/chaperone.rhm
```diff
@@ -0,0 +1,40 @@
+#lang rhombus
+
+import:
+  lib("rhombus/private/treelist.rkt") open
+
+def orig = #{treelist}(1, 2, 3, -4)
+def tl = #{chaperone-treelist}(orig,
```
usaoc, line 7: Can you reformat this call a little bit?

design/treelist/treelist.rhm
```diff
@@ -0,0 +1,1006 @@
...
+def MASK = MAX_WIDTH - 1
+def MAX_ERROR = 2
+
+      
+macro 'radix($index, $height)':
+  '((($index) bits.(>>) (BITS * ($height))) bits.and MASK)'
+
+module test:
+  for (i: 0 .. MAX_WIDTH):
+    check: radix(i, 0)
```
usaoc, line 33: Style: this should use the single-line form?

rhombus/draw/private/font.rhm
```diff
@@ -110,6 +110,6 @@ class Font(private _handle):
       #false
 
   fun norm_hash(map :: Map):
-    racket.hash(& (for values(list = []):
+    racket.hash(& (for values(list = PairList[]):
                      each values(k, v): map
-                     [k, v, &list]))
+                     PairList[k, v, &list]))
```
usaoc, line 115: Can you add a space after this and other occurrences of `&`, just for a more consistent style?

rhombus/private/annotation.rkt
```diff
@@ -791,6 +792,7 @@
 (define-annotation-syntax NonnegReal (identifier-annotation #'nonnegative-real? #'()))
 (define-annotation-syntax Void (identifier-annotation #'void? #'()))
 (define-annotation-syntax False (identifier-annotation #'not #'()))
+(define-annotation-syntax True (identifier-annotation #'(lambda (x) x) #'()))
```
usaoc, line 795: I think using `(lambda (x) (and x #t))` is slightly better so that it’s really a predicate.

rhombus/private/entry-point-adjustment.rkt
```diff
@@ -41,8 +43,8 @@
          entry-point-adjustment-method-table
          #hasheq())
   #:guard (lambda (args wrap is-method? info)
-            (unless (and (list? args) (andmap identifier? args))
-              (raise-argument-error* 'entry_point_meta.Adjustment rhombus-realm "List.of(Identifier)" args))
+            (unless (and (listable? args) (andmap identifier? (to-list #f args)))
```
usaoc, line 46: `(to-list #f args)` is computed twice.  Actually, will it be better to use `to-treelist` here (and have a version of `andmap` that works with that)?

rhombus/private/function-parse.rkt
```diff
@@ -566,19 +573,24 @@
...
                              [(maybe-kwrest-tmp ...) (if kws?
                                                          #'(#:kwrest kwrest-tmp)
                                                          #'())]
                              [maybe-kwrest-tmp-use (if kws?
                                                        #'kwrest-tmp
                                                        #''#hashalw())])
-                 #`[(#,@(entry-point-adjustment-prefix-arguments adjustments) pos-arg-id ...)
+                 #`[(#,@(treelist->list (entry-point-adjustment-prefix-arguments adjustments)) pos-arg-id ...)
                     maybe-rest-tmp ... maybe-kwrest-tmp ...
+                    rest-def ... ;; possible improvement: convert to treelist in indivual try instead of for
```
usaoc, line 592: Typo: “indivual”

rhombus/private/function.rkt
```diff
@@ -57,12 +58,30 @@
...
     (raise-argument-error* who rhombus-realm "List" l)))
 
+(define (check-proc-arity who fn n)
+  (unless (procedure-arity-includes? fn n)
+    (raise-arguments-error* who rhombus-realm
+                            (string-append
+                             "map: argument mismatch;\n"
+                             " the given procedure's expected number of arguments does not match the given number of lists\n")
+                            "given procedure" fn
+                            "expected" (procedure-arity fn)
```
usaoc, line 71: These numbers should be converted to strings and then wrapped as `unquoted-printing-string`s.

rhombus/private/list.rkt
```diff
@@ -31,81 +34,157 @@
...
 
+(define-name-root NonemptyPairList
+  #:fields
+  ([of NonemptyPairList.of]))
+
 (define-binding-syntax List.cons
   (binding-transformer
    (let ([composite (make-composite-binding-transformer
-                     "List.cons" #'nonempty-list? (list #'car) (list #'())
+                     "PairList.cons" #'nonempty-treelist? (list #'(lambda (l) (treelist-ref l 0))) (list #'())
```
usaoc, line 170: Should be `"List.cons"`.

rhombus/private/list.rkt
```diff
@@ -351,108 +651,195 @@
...
+  #:static-infos ((#%call-result #,treelist-static-infos))
+  (treelist-drop-right l n))
+
+(define (check-list-count who what l len n)
+  (when (n . > . len)
+    (raise-arguments-error* who rhombus-realm
+                            (string-append "list is shorter than the number of elements to "
+                                           what)
+                            "list length" len
+                            (string-append "number to " what n))))
```
usaoc, line 691: Wrong parenthesization.

rhombus/private/list.rkt
```diff
@@ -351,108 +651,195 @@
...
+   #`(#,form-id (parens . #,args) . #,tail)
+   rest-arg))
+
+(define-for-syntax (generate-list-binding form-id len or-more? args tail [rest-arg #f] [rest-selector #f]
+                                          [rest-repetition? #t])
+  (define pred #`(lambda (v)
+                   (and (list? v)
+                        #,(if or-more?
+                              #`(length-at-least? v #,len)
+                              #`(= (length v) #,len)))))
```
usaoc, line 809: (Something missed during conflict resolution?)  I changed this to `maybe-list-tail` in dd02e51; see the commit message for the reason.

rhombus/private/listable.rkt
```diff
@@ -0,0 +1,62 @@
...
+    (lambda (field-sym field-proc ary nary fail-k)
+      (case field-sym
+        [(to_list) (nary 1 #'to_list #'make_to_list)]
+        [else (fail-k)])))))
+
+(define-name-root Listable
+  #:fields
+  (to_list))
+
+(define/arity (to_list v)
```
usaoc, line 56: This can use `define/method`, which generates a `<name>/method`.  Also `who` can be used in the body if we change the name to `Listable.to_list`.

rhombus/private/function.rkt
```diff
@@ -57,12 +58,30 @@
...
-  (unless (list? l)
+  (unless (treelist? l)
     (raise-argument-error* who rhombus-realm "List" l)))
 
+(define (check-proc-arity who fn n)
+  (unless (procedure-arity-includes? fn n)
+    (raise-arguments-error* who rhombus-realm
+                            (string-append
+                             "map: argument mismatch;\n"
+                             " the given procedure's expected number of arguments does not match the given number of lists\n")
```
usaoc, line 69: Trailing new line.

rhombus/private/list.rkt
```diff
@@ -150,47 +270,105 @@
...
-        #:when (normal-call? #'tag)
-        (parse-list-form stx #:repetition? #f #:span-form-name? #t)]
-       [(_ . tail)
-        (values #'list #'tail)]))))
+;; used to define `List` and `PairList` further below:
+(define-for-syntax (make-constructor proc-stx build-form static-infos wrap-static-infos
+                                     #:repetition? [repetition? #f]
+                                     #:convert-rep [convert-rep #f])
+  ;; special cases optimize for `...` and `&`; letting it expand
+  ;; instead to `(apply list ....)` is not so bad, but but we can
```
usaoc, line 311: I just noticed the “but but” here, haha

rhombus/private/list.rkt
```diff
@@ -514,21 +909,19 @@
            (lambda new-content
              (let ([content (for/list ([e (in-list content)]
                                        [new-e (in-list new-content)])
-                              (cond
-                                [(list-rest-splice? e) (list-rest-splice new-e)]
-                                [(list-rest-rep? e) (list-rest-rep new-e)]
-                                [else new-e]))])
-               (values (tag-props (build-list-form content))
-                       list-static-infos))))]))
+                              (if (pair? e) (list (car e) new-e) new-e))])
```
usaoc, line 912: Something missed during conflict resolution.

rhombus/private/map.rkt
```diff
@@ -179,8 +181,15 @@
...
+         [(and (treelist? arg) (= 2 (treelist-length arg)))
+          (loop (hash-set ht (treelist-ref arg 0) (treelist-ref arg 1)) (cdr args))]
+         [(and (listable? arg) (let ([arg (to-list #f arg)])
+                                 (and (pair? arg) (pair? (cdr arg)) (null? (cddr arg))
+                                      arg)))
+          => (lambda (arg)
+               (loop (hash-set ht (car arg) (cadr arg)) (cdr args)))]
          [else
-          (raise-argument-error* who rhombus-realm "[_, _]" arg)])])))
+          (raise-argument-error* who rhombus-realm "Listable[_, _]" arg)])])))
```
usaoc, line 192: This annotation is very wrong; it should be something like `Listable.to_list && matching([_, _])` (again, we should probably add a `Listable.to_list` annotation for this purpose).

rhombus/tests/annotation.rhm
```diff
@@ -38,3 +38,13 @@ check:
...
   ~throws values("contract violation", "expected: Function.of_arity(1)")
+
+check:
+  "x" :: (False || ((converting(fun (v): [v]) && True) || False))
+  ~is ["x"]
+
+check:
+  #false :: False ~is #false
+  "x" :: True ~is "x"
+  "x" :: False ~raises "does not satisfy"
```
usaoc, line 49: `~raises` → `~throws`

rhombus/scribblings/ref-stxobj.scrbl
```diff
@@ -760,7 +758,7 @@ Metadata for a syntax object can include a source location and the raw
 
 
 @doc(
-  fun Syntax.unwrap_group(stx :: Group) :: List.of(Syntax)
+  fun Syntax.unwrap_group(stx :: Group) :: Listable.of(Syntax)
```
usaoc, line 761: These *result* annotations should simply say `List.of`.

rhombus/private/pack-s-exp.rkt
```diff
@@ -14,10 +15,10 @@
...
          [_ stx])]
-      [(list? s)
-       (for/list ([e (in-list s)])
+      [(listable? s)
+       (for/list ([e (in-list (to-list #f s))])
          (loop e))]
       [else
        (raise-arguments-error* who rhombus-realm
-                               "not a list nesting of syntax objects"
+                               "not a Liistable nesting of syntax objects"
```
usaoc, line 23: Typo: “Liistable”

rhombus/private/treelist.rkt
```diff
@@ -0,0 +1,1102 @@
...
+           #t
+           #t
+           ((fx+ pos 1) next-node next-node-pos))]]
+      [_ #f])))
+
+(define-syntax (for/treelist stx)
+  (syntax-case stx ()
+    [(_ binds body0 body ...)
+     (with-syntax ([((pre-body ...) (post-body ...)) (split-for-body stx #'(body0 body ...))])
+       #'(for/fold/derived #,stx ([root empty-node] [size 0] [height 0]
```
usaoc, line 225: Should use `` #` ``.

rhombus/private/treelist.rkt
```diff
@@ -0,0 +1,1102 @@
...
+           pre-body ...
+           (unsafe-root-add root size height
+                            (let ()
+                              post-body ...))))]))
+
+(define-syntax (for*/treelist stx)
+  (syntax-case stx ()
+    [(_ binds body0 body ...)
+     (with-syntax ([((pre-body ...) (post-body ...)) (split-for-body stx #'(body0 body ...))])
+       #'(for*/fold/derived #,stx ([root empty-node] [size 0] [height 0]
```
usaoc, line 239: Ditto.

rhombus/private/syntax-object.rkt
```diff
@@ -317,12 +326,13 @@
 
 (define (check-nonempty-list who l)
   (unless (and (pair? l) (list? l))
-    (raise-argument-error* who rhombus-realm "NonemptyList" l)))
+    (raise-argument-error* who rhombus-realm "NonemptyListable" l)))
```
usaoc, line 329: Same problem as elsewhere.

rhombus/scribblings/list.scrbl
```diff
@@ -189,5 +184,49 @@ elements:
...
+@rhombus(List.delete), or dropping or taking a list prefix or suffix
+with functions like @rhombus(List.drop_left). Internally, lists are
+implemented as relaxed radix balanced (RRB) trees.
+
+As an alternative to @rhombus(List), @rhombus(PairList) constructs a
+@tech{pair list}, which is implemented as a singly linked list. As the
+name suggests, a pair list uses a @tech{pair} to add each element to the
+front of an existing list. The @rhombus(PairList.cons) operation
+performs that addition, and it takes @math{O(1)} time, as do its
+@rhombus(PairList.first) and @rhombus(PairList.rest) operationsto access
```
usaoc, line 201: Typo: “operationsto”

usaoc:
> Would `ListLike` work better than `Listable`?

For comparison, Haskell uses `IsList` for a similar interface, which also includes `fromList` and `fromListN`.

> `Listable` is the common interface but it's very thin intentionally […]

I think a “thin” interface like this is fine, and doesn’t interfere with our grand scheme.

rhombus/scribblings/map.scrbl
```diff
@@ -12,10 +12,11 @@
...
 The @rhombus(Map) constructor creates an immutable mapping of arbitrary
-keys to values. The term @deftech{map} is meant to be generic, and
-@rhombus(Map) as a constructor just uses a default implementation of
-maps. The @rhombus(Map) constructor can be used like a function, in
-which case it accepts keys paired with values in two-item lists:
+keys to values. A map is @tech{indexable} using @brackets with a key,
+and the result is the corresponding value.
+
+The @rhombus(Map) constructor can be used like a function, in which case
+it accepts keys paired with values in two-item lists to ceate a map:
```
usaoc, line 19: Typo: “ceate”

rhombus/scribblings/ref-list.scrbl
```diff
@@ -233,6 +275,20 @@ it supplies its elements in order.
 }
 
 
+@doc(
+  fun List.delete(lst :: List, n :: NonnegInt) :: List
+){
+
+ Creates a list like @rhombus(lst), but with out the @rhombus(n)th
```
usaoc, line 282: Typo: “with out”

rhombus/scribblings/ref-list.scrbl
```diff
@@ -297,6 +358,30 @@ it supplies its elements in order.
...
 
 
+@doc(
+  fun List.take_left(lst :: List, n :: NonnegInt) :: List
+  fun List.take_right(lst :: List, n :: NonnegInt) :: List
+){
+
+ Returns a list like @rhombus(lst), but with only the first @rhombus(n)
+ elements in the case of @rhombus(List.take_left), or only the last
+ @rhombus(n) elements in the case of @rhombus(List.dtake_right). The given
```
usaoc, line 368: Typo: `@rhombus(List.dtake_right)`

rhombus/scribblings/ref-listable.scrbl
```diff
@@ -0,0 +1,73 @@
...
+@title{Listables}
+
+A @deftech{listable} value can be converted to a @tech{list} using a
+@rhombus(to_list) operation such as @rhombus(Listable.to_list). Lists
+are themselves listable, where @rhombus(List.to_list) just returns the
+same list. @tech{Pair lists}, @tech{array}s, and instances of classes
+that implemement @rhombus(Listable, ~class) are also listable.
+
+List-splicing contexts generally allow listables, including in the
+construction of a list. As a result, of @nontermref(expr) produces a
```
usaoc, line 18: Typo: “of”

rhombus/scribblings/ref-pair.scrbl
```diff
@@ -103,3 +127,399 @@ its ``rest'' is a list.
...
+ @rhombus(PairList.of, ~annot) makes an expression acceptable as a sequence
+ to @rhombus(for) in static mode.
+
+}
+
+@doc(
+  ~nonterminal:
+    listable_expr: block expr
+  fun PairList(v :: Any, ...) :: List
+  expr.macro 'PairList [$expr_or_splice, ...]'
```
usaoc, line 150: This should exclude the space to be consistent with `List`.

rhombus/scribblings/ref-pair.scrbl
```diff
@@ -103,3 +127,399 @@ its ``rest'' is a list.
...
+ Static information associated by @rhombus(PairList, ~annot) or
+ @rhombus(PairList.of, ~annot) makes an expression acceptable as a sequence
+ to @rhombus(for) in static mode.
+
+}
+
+@doc(
+  ~nonterminal:
+    listable_expr: block expr
+  fun PairList(v :: Any, ...) :: List
```
usaoc, line 149: Result annotation.

design/treelist/bm.rhm
mflatt: Thanks for the working through this very large set of changes!

This one is hopefully fixed (HF) — that is, fixed unless I still got it wrong in the latest commit.

design/treelist/chaperone.rhm
mflatt: I think this start at a test suite is destined for Racket, so I didn't change this, so far.

rhombus/draw/private/font.rhm
mflatt: HF

rhombus/private/annotation.rkt
mflatt: HF

rhombus/private/entry-point-adjustment.rkt
mflatt: I agree about using `to-treelist` and I made that change. I'm not sure about `Listable.of` versus `Listable.to_list && ....` here, so that's pending.

rhombus/private/function-parse.rkt
mflatt: HF

rhombus/private/function.rkt
mflatt: HF

rhombus/private/list.rkt
mflatt: HF

rhombus/private/list.rkt
mflatt: HF

rhombus/private/function.rkt
mflatt: HF

rhombus/private/list.rkt
mflatt: HF

rhombus/tests/annotation.rhm
mflatt: HF

rhombus/scribblings/ref-stxobj.scrbl
mflatt: HF

rhombus/private/pack-s-exp.rkt
mflatt: HF

rhombus/private/treelist.rkt
mflatt: HF

rhombus/private/treelist.rkt
mflatt: HF

rhombus/scribblings/list.scrbl
mflatt: HF

rhombus/scribblings/map.scrbl
mflatt: HF

rhombus/scribblings/ref-list.scrbl
mflatt: HF

rhombus/scribblings/ref-list.scrbl
mflatt: HF

rhombus/scribblings/ref-listable.scrbl
mflatt: HF

rhombus/scribblings/ref-pair.scrbl
mflatt: HF

rhombus/scribblings/ref-pair.scrbl
mflatt: HF

rhombus/scribblings/ref-pair.scrbl
```diff
@@ -103,3 +127,399 @@ its ``rest'' is a list.
...
+
+@doc(
+  fun PairList.take_left(lst :: PairList, n :: NonnegInt) :: PairList
+  fun PairList.take_right(lst :: PairList, n :: NonnegInt) :: PairList
+){
+
+ Like @rhombus(List.take_left) and @rhombus(List.take_right), but for
+ @tech{pair lists}. Creating the new list takes @math{O(log N)} time,
+ where @math{N} is the @rhombus(n) argument for taking from the left or
+ the length of the pair list for taking from the right.
```
usaoc, line 405: This description seems wrong; `take_left` creates a new list and takes linear time w.r.t. the number of elements to take (something to note: the implementation doesn’t comply with this!); `take_right` returns a certain tail and takes linear time w.r.t. the length of the list.

rhombus/scribblings/ref-pair.scrbl
```diff
@@ -103,3 +127,399 @@ its ``rest'' is a list.
...
+
+@doc(
+  fun PairList.drop_left(lst :: PairList, n :: NonnegInt) :: PairList
+  fun PairList.drop_right(lst :: PairList, n :: NonnegInt) :: PairList
+){
+
+ Like @rhombus(List.drop_left) and @rhombus(List.drop_right), but for
+ @tech{pair lists}. Creating the new list takes @math{O(log N)} time,
+ where @math{N} is the @rhombus(n) argument for dropping from the left or
+ the length of the pair list for dropping from the right.
```
usaoc, line 425: Ditto.

rhombus/scribblings/ref-reducer-macro.scrbl
```diff
@@ -120,7 +120,7 @@
     expr.macro 'build_finish [$count, $_, $_, $_, $_, $finish, $data]':
       'values($finish $data, $count + 1)'
   ~repl:
-    for counted(List) (i: 0..3): i
+    for counted(PairList) (i: 0..3): i
```
usaoc, line 123: This one doesn’t need to change, I think.

rhombus/scribblings/ref-list.scrbl
```diff
@@ -54,15 +60,17 @@ it supplies its elements in order.
...
 @doc(
   ~nonterminal:
-    list_expr: block expr
+    listable_expr: block expr
   fun List(v :: Any, ...) :: List
   expr.macro '#%brackets [$expr_or_splice, ...]'
   repet.macro '#%brackets [$repet_or_splice, ...]'
-
+  expr.macro 'List [$expr_or_splice, ...]'
+  repet.macro 'List [$repet_or_splice, ...]'
```
usaoc, line 68: I missed this one; the space should be removed.

usaoc:
I looked at the comments I left and added some new comments.

> I'm not sure about `Listable.of` versus `Listable.to_list && ....` here […]

The biggest problem is that we don’t have a `Listable.of` (or `NonemptyListable` or `Listable[....]`), and I’m not sure adding those is a good idea.  Using an “adapter” annotation like `Listable.to_list` and chaining to the old annotations seems to be the more appropriate thing to do here, because that’s essentially what we’re doing manually.

More generally, I think we should consistently provide annotations parallel to `to`/`from` functions at some point, and maybe we should add some forms to more conveniently do that (internally or in Rhombus).

scribble/private/base.rhm
```diff
@@ -0,0 +1 @@
+#lang rhombus
```
usaoc, line 1: What is this file supposed to do?

mflatt:
> Would ListLike work better than Listable? The able suffix makes sense when the prefix is a verb.

"List" is a verb. It's different from some other verbs that we've used for interfaces in that "list" is also used as a noun, but the same is true of "sequence" with `Sequence` and `Sequenceable`.

I considered "enumerate" and `Enumerable` in place of "list" and `Listable`, but to me it seemed long and needlessly indirect for an interface that is specifically related to `List`.

> Edit: I totally forgot this—we already have {Point,Size,Rect}Like in rhombus/draw, so ListLike is a more consistent choice.

Well, `PointLike`, etc., are not interfaces that can be implemented by a new class. If we wanted to support that, maybe they should have "able" names (although I don't think that's worthwhile).

Or we could always revisit the use of "able" convention.

mflatt:
> Using an “adapter” annotation like Listable.to_list and chaining to the old annotations seems to be the more appropriate thing to do here, because that’s essentially what we’re doing manually.

Ok, I 'm starting to see how that does fit together well. Although including `Listable.to_list` in an API feels to me too much like talking about implementation instead of interface, using a `Listable.to_list` annotation would ensure a single early conversion, so maybe it's a good compromise/default that avoids needing other things.

rhombus/private/list.rkt
mflatt: HF

rhombus/private/list.rkt
mflatt: HF

rhombus/scribblings/ref-pair.scrbl
mflatt: HF

rhombus/scribblings/ref-pair.scrbl
mflatt: HF

rhombus/scribblings/ref-reducer-macro.scrbl
mflatt: Right, the change to `List` exposed a limitation/bug in the example. Not fixed, yet.

rhombus/scribblings/ref-list.scrbl
mflatt: HF

scribble/private/rhombus-spacer.rhm
```diff
@@ -125,7 +125,7 @@ meta:
           '$new_tag $new_p'
       | ~else: spacer.adjust_sequence(res_ann, #'~annot, esc)
     '$new_op $new_res_ann'.relocate(stx)
-
+
```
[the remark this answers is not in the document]
mflatt: HF

scribble/private/base.rhm
mflatt: Accidentally added, now removed.

rhombus/private/entry-point-adjustment.rkt
mflatt: HF, with `Listable.to_list` added as a converter annotation.

rhombus/private/listable.rkt
mflatt: HF

rhombus/private/map.rkt
mflatt: HF

rhombus/scribblings/ref-stxobj.scrbl
```diff
@@ -640,7 +638,7 @@ Metadata for a syntax object can include a source location and the raw
 }
 
 @doc(
-  fun Syntax.make_group(terms :: List,
+  fun Syntax.make_group(terms :: Listable,
```
usaoc, line 641: Also note that this should to be `.... && NonemptyList`.

usaoc:
This is mostly good now, except the few unresolved conversations; also, a few occurrences of `Listable.of`/`NonemptyListable` are missed.  I’ll see what I can do with the reducer macro example, but that’s not urgent.

mflatt:
Rebased on #452 and #454, intended to be merged after those.

rhombus/private/list.rkt
```diff
@@ -311,24 +567,68 @@
                            proc)))
 
 (define/method (List.map lst proc)
+  #:static-infos ((#%call-result #,treelist-static-infos))
+  (check-treelist who lst)
+  (check-function-of-arity 1 who proc)
+  (for/fold ([accum empty-treelist]) ([e (in-treelist lst)])
```
usaoc, line 573: I think this can use `for/treelist`?

rhombus/private/list.rkt
```diff
@@ -202,6 +380,15 @@
     #`((#%index-result #,(car static-infoss))))
   #'list-build-convert #'())
 
+(define-syntax (treelist-build-convert arg-id build-convert-stxs kws data)
+  #`(for/fold ([lst empty-treelist])
```
usaoc, line 384: Will it be worth it to use unsafe operations here?

rhombus/private/list.rkt
mflatt: HF

rhombus/private/list.rkt
mflatt: It's difficult to be sure, but my best guess is that it won't matter.

rhombus/private/entry-point-adjustment.rkt
mflatt: HF^2

rhombus/private/syntax-object.rkt
mflatt: HF

rhombus/scribblings/ref-pair.scrbl
mflatt: HF^2

rhombus/scribblings/ref-reducer-macro.scrbl
mflatt: HF, since you fixed the example.

rhombus/scribblings/ref-stxobj.scrbl
mflatt: HF

PR #456, opened 2024-01-03
use more namespaces to group bindings

usaoc:
Try to group some bindings relevant to namespaced annotations under corresponding namespaced namespaces.

### DIFF
--- a/rhombus/private/control.rkt
+++ b/rhombus/private/control.rkt
@@ -32,15 +32,26 @@
   #:fields
   (Marks
    PromptTag
-   current_marks
+   ;; TEMP see `Marks`
+   [current_marks Continuation.Marks.current]
    capture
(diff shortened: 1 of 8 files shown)


usaoc:
Opening to ask if people feel this is a good direction.

mflatt:
This seems good to me. Not sure whether we should merge now or wait for a large group of backward-incompatible changes in the near future.

usaoc:
Merging now is fine if no further objection arises.  This one comes with temporary aliases for backward compatibility, so it should be “less backward-incompatible.”

mflatt:
I keep forgetting to reply here, but I agree that merging now is fine. (I had noticed compatibility bindings the first time around, but somehow that didn't sink in.)

PR #463, opened 2024-01-12
allow `~open` in `pattern` shorthands

usaoc:
This form is useful when you need to name a match, but the match’s attributes are already being used.


mflatt:
LGTM

PR #466, opened 2024-01-15
refactor generation of functions with keyword rest arguments

usaoc:
This also fixes some bugs in arity extraction, which are currently missing test cases.  I think the test cases will need to test on entry-point forms.


mflatt:
I have not looked at this commit in detail, but overall it seems fine, so LGTM.

PR #470, opened 2024-01-23
more consistently expose methods related to implicit forms

usaoc:
This also makes it so that the methods can be properly documented.

### DIFF
--- a/rhombus/scribblings/ref-array.scrbl
+++ b/rhombus/scribblings/ref-array.scrbl
@@ -22,6 +22,9 @@ mutable and immutable arrays, while @rhombus(MutableArray, ~annot) and
   "array",
   @rhombus(Array),
   [arr.length(), Array.length(arr)]
+  [arr.get(n), Array.get(arr, n)]
+  [arr.set(n, v), Array.get(arr, n, v)]
+  [arr.append(arr2, ...), Array.append(arr, arr2, ...)]
   [arr.copy(arg, ...), Array.copy(arr, arg, ...)]
   [arr.copy_from(arg, ...), Array.copy_from(arr, arg, ...)]
   [arr.take_left(n), Array.take_left(arr, n)]
@@ -29,6 +32,7 @@ mutable and immutable arrays, while @rhombus(MutableArray, ~annot) and
   [arr.drop_left(n), Array.drop_left(arr, n)]
   [arr.drop_right(n), Array.drop_right(arr, n)]
   [arr.set_in_copy(i, v), Array.set_in_copy(arr, i, v)]
+  [arr.to_list(), Array.to_list(arr)]
 )
 
 @doc(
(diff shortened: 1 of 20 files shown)


rhombus/scribblings/ref-map.scrbl
```diff
@@ -361,6 +363,23 @@ in an unspecified order.
...
 
 
+@doc(
+  fun Map.append(map :: Map, ...) :: Map
+){
+
+ Functionally appends @rhombus(map)s, like the @rhombus(++) operator
+ (but without the special optimization). In case mappings of a key
+ exist in multiple maps, the rightmost mapping is used, following the
+ precedence of @rhombus(++).
```
mflatt, line 373: precedent

rhombus/scribblings/ref-map.scrbl
```diff
@@ -361,6 +363,23 @@ in an unspecified order.
 }
 
 
+@doc(
+  fun Map.append(map :: Map, ...) :: Map
+){
+
+ Functionally appends @rhombus(map)s, like the @rhombus(++) operator
+ (but without the special optimization). In case mappings of a key
```
mflatt, line 371: FWIW, I had trouble parsing `In case mappings of a key`. Maybe `When a key has a value in multiple given @rhombus(map)s,`?

rhombus/scribblings/ref-array.scrbl
```diff
@@ -192,10 +196,61 @@ mutable and immutable arrays, while @rhombus(MutableArray, ~annot) and
 }
 
 
+@doc(
+  fun Array.get(arr :: Array, n :: NonnegInt) :: Any
+){
+
+ Returns the @rhombus(n)th element of @rhombus(arr) (starting from
+ @rhombus(0)), like the implicit @rhombus(#%index) form (usually
+ written with @brackets).
```
mflatt, line 205: This is precise, but I think it would help to show `arr[n]` at the start of the explanation (in addition to the examples), since someone who jumps to this documentation probably won't know about the weird name `#%index` and how it it written with `[]`, but they'll probably instantly recognize `arr[n]`.

rhombus/scribblings/ref-array.scrbl
```diff
@@ -192,10 +196,61 @@ mutable and immutable arrays, while @rhombus(MutableArray, ~annot) and
...
+
+
+@doc(
+  fun Array.set(arr :: MutableArray, n :: NonnegInt, val :: Any)
+    :: Void
+){
+
+ Updates the @rhombus(n)th position of @rhombus(arr) to @rhombus(val),
+ like the @rhombus(:=) operator along with the implicit
+ @rhombus(#%index) form.
```
mflatt, line 222: Similarly, I'd show `arr[n] := val` explicitly here.

rhombus/scribblings/ref-bytes.scrbl
```diff
@@ -70,6 +73,56 @@ and @rhombus(ImmutableBytes, ~annot) require one or the other.
 }
 
 
+@doc(
+  fun Bytes.get(bstr :: Bytes, n :: NonnegInt) :: Byte
+){
+
+ Returns the @rhombus(n)th byte of @rhombus(bstr) (starting from
+ @rhombus(0)), like the implicit @rhombus(#%index) form (usually
+ written with @brackets).
```
mflatt, line 82: Same for this and other `[]` forms...

mflatt (approved):
Thanks very much for this clean-up! I have small editing suggestions for the docs.

rhombus/scribblings/ref-map.scrbl
usaoc: I think I meant “(left) associativity” here.

PR #471, opened 2024-01-28
use internal-definition contexts for `let` scopes

### LINKED ISSUE #467: Scope pruning in Rhombus blocks
This Racket program

```racket
#lang racket/base
(let-values ([(id1 id2)
              (let ([id1 (quote-syntax id)])
                (values id1 (quote-syntax id)))])
  (bound-identifier=? id1 id2))
```

leads to `#t`, due to “scope pruning” in `quote-syntax`.  However, this Rhombus program

```
#lang rhombus/static
import:
  meta -1:
    rhombus/meta.syntax_meta

block:
  let (id1, id2):
    let id1 = Syntax.literal 'id'
    values(id1, Syntax.literal 'id')
  syntax_meta.equal_name_and_scopes(id1, id2)
```

leads to `#false`.  This can make the use of `let` in macros rather awkward, so we should probably try to align with what Racket is doing.

usaoc:
*(Is this a reasonable solution?)*

This allows the scope-pruning behavior of `quote-syntax` to apply.

Fix #467.

### DIFF
--- a/rhombus/private/forwarding-sequence.rkt
+++ b/rhombus/private/forwarding-sequence.rkt
@@ -149,7 +149,7 @@
          #:literals (begin define-values define-syntaxes rhombus-forward pop-forward #%require provide #%provide quote-syntax
                            define-syntax-parameter)
          [(rhombus-forward sub-form ...)
-          (define introducer (make-syntax-introducer #t))
+          (define introducer (make-syntax-introducer/intdef))
           #`(begin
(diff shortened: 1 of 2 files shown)


mflatt:
This works as far as I can tell. I think we probably want to add a `syntax-local-make-definition-context-introducer`, or something like that, to create just one scope without the definition-context overhead, but this patch's approach would be the backward-compatible implementation.

Issue #474, opened 2024-02-07
Use-site binder hygiene bug for `let`


michaelballantyne:
I've previously discussed this with @mflatt , but I figure I should make a ticket.

Consider this program:

```
#lang rhombus/and_meta

block:
  let x = "outer"
  defn.macro 'm $id':
    'let $id = "inner"
     x'
  m x
```

It returns "inner", but I think it should return "outer" like the similar Racket program does:

```
(define x "outer")
(define-syntax-rule (m id)
 (let ([id "inner"])
   x))
(m x)
```

I want to point this out now because I saw https://github.com/racket/racket/pull/4929, which proposes to add something new to the `syntax-local-` zoo to support the implementation of Rhombus `let`. I'm a little skeptical of the current implementation strategy for `let` given that it gets this use-site binder hygiene wrong, and it feels like it'd be unfortunate to expose more low level `syntax-local-` operations that could point users towards implementations of scoping structures that don't get hygiene right.

In contrast, an implementation of `let` using definition contexts and local-expand should be able to avoid this problem because one of the things definition contexts encapsulate is a set of local use-site scopes. I don't know enough about Rhombus to know what other implementation constraints there are, though.

usaoc:
Given the current way Rhombus blocks are expanded, I think the alternative strategy you suggest will require much more significant change, because the expansion uses a trampolining macro, which can’t thread through an opaque internal-definition context value.  I’ll do some investigation to see what is exactly happening, but I don’t think we can change the strategy very easily.

Another thing that’s curious about `let` is that a `def` after `let` works like “splicing” in the sense that they should be visible to forms *before* the `let`.  There isn’t much precedent to this, except the historical `define*` form, which works like Rhombus `let`:

```racket
#lang racket/base
(require compatibility/package)

(define-package test ()
  (define* x "outer")
  (define-syntax-rule (m id)
    (begin
      (define* id "inner")
      (println x)))
  (m x))
;; prints "inner"
```

usaoc:
[...]

PR #476, opened 2024-02-13
use `syntax-local-apply-transformer`

mflatt:
Switch enforestation-level expansion away from a custom expansion step and scope management, and use `syntax-local-apply-transformer`, instead. Thanks to @michaelballantyne for reminding me how this is supposed to work.

This is step 1 of 2 toward correcting enforestation-level expansion to have use-site scopes. Related to the discussion in #474.

### DIFF
--- a/enforest/operator.rkt
+++ b/enforest/operator.rkt
@@ -136,8 +136,8 @@
        [else dir])]))
 
 (define (lookup-prefix-implicit alone-name adj-context adj-form in-space operator-ref operator-kind form-kind)
-  (define op-stx (datum->syntax adj-context alone-name))
-  (define op (syntax-local-value* (in-space op-stx) operator-ref))
+  (define op-stx (in-space (datum->syntax adj-context alone-name)))
+  (define op (syntax-local-value* op-stx operator-ref))
   (unless op
     (raise-syntax-error #f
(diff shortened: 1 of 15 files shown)


enforest/operator.rkt
```diff
@@ -215,39 +215,48 @@
...
 
 (define (apply-infix-direct-operator op form1 form2 stx track-origin checker)
-  (call-as-transformer
-   stx
-   track-origin
-   (lambda (in out)
-     (define proc (operator-proc op))
-     (checker (out (proc (in form1) (in form2) stx)) proc))))
+  (define proc (operator-proc op))
+  (checker (call-as-transformer
```
usaoc, line 230: Nit: the similar function call above is formatted differently.

rhombus/private/import.rkt
```diff
@@ -240,7 +240,7 @@
 
 (define-defn-syntax import
   (definition-transformer
-   (lambda (stx)
+    (lambda (stx)
```
usaoc, line 243: This whitespace change probably shouldn’t be included?

enforest/operator.rkt
mflatt: Thanks - made consistent.

PR #477, opened 2024-02-15
add use-site scopes for definition and expression expansion

mflatt:
Part 2 of 2, following up on #476.

Use-site scope tracking is enabled only for expression, definition, and declaration contexts. I'm pretty sure they're not needed anywhere else, although I worry that I might be overlooking something. Disabling use-site-scope tracking for other contexts avoids having to figure out where to remove them.

It's not obvious from the documentation for `local-expand` that an internal-definition-signalling list must be provided as the context argument is needed to enable use-site scopes. Inventing a new context each time in `call-as-transformer` is a little awkward, but it's ok as far as I can reason though it. I'm inclined to blame `local-expand` for having an awkward interface, plan to improve the docs, and leave it at that.

There's a little clean-up in this commit to remove `rhombus-mixed-forwarding-sequence`, which is unused, even though it's not really about support for use-site scopes.

### DIFF
--- a/enforest/main.rkt
+++ b/enforest/main.rkt
@@ -108,6 +108,8 @@
                          #:defaults ([check-result #'check-is-syntax]))
               (~optional (~seq #:track-origin track-origin)
                          #:defaults ([track-origin #'syntax-track-origin]))
+              (~optional (~seq #:use-site-scopes? use-site-scopes?)
+                         #:defaults ([use-site-scopes? #'#f]))
               (~optional (~seq #:make-identifier-form make-identifier-form)
                          #:defaults ([make-identifier-form #'values]))
               (~optional (~seq #:make-operator-form make-operator-form)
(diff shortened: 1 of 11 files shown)


usaoc:
I pushed a commit (usaoc/rhombus-prototype@ebd1160) that adds a more complete test suite (and the PR’s commit is amended to revert the test-related change).  If you’re okay with that, this PR can reset to that branch.

Another thing I noticed that needs discussion/documentation is that `let` “scopes over” `def`:

```
#lang rhombus

block:
  let x = "outer"
  def x = "inner"
  // refers to the `let` binding
  x
```

this is more tangential to the PR, but it’s something that I didn’t notice until the writing of that test suite.

mflatt:
Thanks! My initial guess is that the `def` and `let` issue might be addressed with a check at the point where `define` or `define-values` is discovered in `forwarding-sequence-step`, but I'm not sure.

PR #478, opened 2024-02-16
add `matches` infix operator

usaoc:
This is like `matches` macro in Rust, which can serve as a shorthand and more discoverable alternative to
`<expr> is_a matching(<bind>)`.


mflatt:
LGTM

PR #479, opened 2024-02-16
add `pict` and `slideshow` libraries

mflatt:
This is most of the way to Rhombus `pict` and `slideshow` libraries. These libraries are built on the Racket implementations, of course, but not as thin wrappers. Most significantly, the Rhombus `Pict` datatype incorporates a new abstraction for animated picts, and the Slideshow variant cooperates to create better paths between animated picts and the `slide` sublanguage for staging and aligning slide content.

I've put rendered documentation here, but beware that cross links between the documents don't work:

https://users.cs.utah.edu/~mflatt/tmp/rhombus-slide/

For `pict`, function names and arguments are based on the Racket `pict` and `2htdp/image` libraries, but changed further in an attempt to make things even more consistent. Function conventions rely heavily on keywords and defaults to pack in a lot of functionality. For example, `rectangle` has an `~around` argument that lets it do the job of `frame` as well as `rectangle` and `filled-rectangle`.

The notion of animation built into `Pict` is motivated by the problems of slide rendering, but it's not inherently tied to Slideshow. Then again, the details are slide-oriented in much the same way that pict "baseline" and "topline" concepts are oriented dealing with horizontal text (and there's no built-in support for a horizontal analog of baseline, for example). To a first approximation, the animation idea is to introduce a "time box" property for a pict that is like a bounding box, but in the time dimension. See the overview in the `pict` docs for more information.

The `slide` function in the Rhombus `slideshow` library takes picts, as usual, but those can be animated picts, in which case multiple slides are registered. The `slide` function here has a cleaned-up variant of `'next` and `alts`, but those can be viewed (and are implemented as) a conversion to a single animated pict. Along those lines, the `slide_pict` function that takes the same arguments as `slide`, but returns an animated pict instead of registering a slide. An entire Slideshow talk can be represented as single animated pict — or not, because the more traditional Racket Slideshow view is often more convenient, and the two views mesh well through the animated-pict abstraction.

I've put my current slides that use these libraries at https://github.com/mflatt/netsec-slides. Like a lot of slide code, it's not great code (just a notch above throw-away), but maybe useful to demonstrate what it looks like to build slides with Rhombus and these libraries.

There are still many things I'd like to add or improve. A notable omission is any support for rendering Rhombus code as a pict, since I don't need that for my current class. And `standard-fish`.

Thanks to @oflatt and @rfindler for their discussions and contributions to the design.


samth:
One thing I note here is that relative to `pict` and `slideshow`, this is a much more keyword-heavy API (eg, with `circle`). Maybe this sort of stylistic question should be a topic for a Rhombus meeting?

PR #481, opened 2024-02-20
docs: use `Parameter.def` to document parameters

usaoc:
This PR also includes a commit that adds `Any.to_boolean`, mainly for documenting “Boolean parameters” here.


mflatt:
This is great - thanks!

PR #484, opened 2024-03-06
add `as` support in `expose`

usaoc:
This works like a combination of `rename` and `expose`, which is not very common, but does appear to be used in the Rhombus codebase.

Also change the grammar to disallow multiple names in a group in the multi-group form, to avoid confusion with the `as` form.

### DIFF
--- a/rhombus/draw/private/rkt.rhm
+++ b/rhombus/draw/private/rkt.rhm
@@ -2,9 +2,6 @@
 import:
   lib("racket/class.rkt"):
     expose:
-      send
-      #{is-a?}
-    rename:
       send as rkt_send
       #{is-a?} as rkt_is_a
   lib("racket/draw.rkt")
--- a/rhombus/measure.rhm
+++ b/rhombus/measure.rhm
@@ -1,8 +1,8 @@
 #lang rhombus/and_meta
 import:
   lib("racket/base.rkt"):
-    expose: time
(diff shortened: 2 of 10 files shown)


mflatt:
This is ok with me.

An observation, and probably something you've already considered: If a macro implementation includes `expose: $id ...`, then there's a chance that the macro could be made to do something strange with three identifies for `id` where the middle one is `as`. The right implementation of the macro would be to use `expose: $id; ...`, I guess. This kind of subtlety shows up in a few places in Rhombus, and I don't know whether it's avoidable. At a minimum, we need to make sue that there's some way of reliably expanding, and separate groups provide that mechanisms for `expose`.

usaoc:
I considered changing the grammar to `expose: $id_or_rename_as; ...`, but I worried about the backward compatibility.  I can apply the change if that wouldn’t be a big concern.

usaoc:
One more thing I considered and wasn’t so sure was whether the same support should be added to `only`.  I reckoned it wouldn’t be as useful, but it would be more consistent.

usaoc:
I applied the change and found three places where the `expose: $id ...` form is used (none uses more than one group), among which one place was in the import-macro docs.

mflatt:
The change to `expose` caused no problems with any of my other existing code, and this seems ready to merge.

Issue #480, opened 2024-02-17
A `let` before another `def` in the same block


usaoc:
*(Found during the writing of e7dc0d9; opening for public attention)*

Currently, a `let` can “scope over” another `def` in the same block:

```
#lang rhombus

block:
  let x = "outer"
  def x = "inner"
  // refers to the `let` binding, i.e., "outer"
  x
```

This is because the `let` binding has an extra scope (in order to achieve “forward definition”) compared to the `def` binding.  Is this a reasonable behavior?  And in general, how should `let` and `def` interact within the same block?

samth:
I think either an error or "inner" are acceptable behaviors here.

mflatt:
Made an error in bfbb0975096f34a8bbe9e566ce912555427a114a

PR #485, opened 2024-03-07
disallow `let` and `def` on the same identifier in the same definition context

mflatt:
The fist commit in this PR is intended to address #480 by making the combination of `def` and `let` for the same identifier an error in a given binding context. Otherwise, the commit preserves the current behavior and intent of `let`.

The second commit addresses the longstanding problem that `export`s by a namespace of an unbounded identifier didn't trigger an error until an attempt to use the exported binding. That repair is part of this PR because the new checking has particular support for `let`: it detects the case that the identifier is unbound because it is defined later with `let` and ties to provide a better error message in that case. (I have run into that situation and been confused a handful of times, and I think the improved error message would have helped.)

In the February 22 Rhombus meeting, there was a brief discussion about changing the meaning of `let` to create different definition contexts  before and after the `let`. After looking into that possibility, though, I concluded that it wouldn't help enough to detangle competing `let`s and `def`s within a block, and it wouldn't interfere with the intent of `let` and the way that I have been using `let` in practice. So, there's no change along those lines that here.

### DIFF
--- a/rhombus/private/class-dot.rkt
+++ b/rhombus/private/class-dot.rkt
@@ -11,6 +11,7 @@
          "class-method-result.rkt"
          (submod "dot.rkt" for-dot-provider)
          "dotted-sequence-parse.rkt"
+         "export-check.rkt"
          "parens.rkt"
          "static-info.rkt"
          "expression.rkt"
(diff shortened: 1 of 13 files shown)


usaoc:
One thing that stands out to me is that some test cases in `rhombus/tests/use-site-scope.rhm` involving `let` after `def` still pass; I would expect those to fail with “duplicate definition” instead.

mflatt:
Thanks for noticing that! I've pushed a repair as a new commit.

PR #486, opened 2024-03-08
implement keyword punning in function calls

usaoc:
This is most similar to label punning in OCaml.


usaoc:
This was proposed in the 8th March Rhombus meeting, so I rolled out a quick implementation for people to try around.

usaoc:
I think we can merge this and try it out if there’s no objection from anyone.

PR #487, opened 2024-03-14
shrubbery: exact rationals as literals

mflatt:
Allow `1/2` as a literal, for example. To go along with this, lift the precedence restriction in Rhombus against `*` after `/`.


rhombus/scribblings/ref-number.scrbl
```diff
@@ -210,8 +208,7 @@
   8/2
```
usaoc, line 208: This and the lower example are not exemplifying the `/` operator.

mflatt: Changed - thanks!

PR #488, opened 2024-03-18
add `..=` for inclusive finite ranges

usaoc:
The `..=` operator has a precedent in Rust.

This PR also includes a commit that fixes suffix `..`.


mflatt:
I like the idea of having a way to make an inclusive range, and the Rust choice of `..=` (which I didn't know about) seems fine.

I had previously considered trying `<expr> .. <expr> ~inclusive` (and allowing `<expr> .. <expr> ~exclusive` to mean the same thing as `<expr> .. <expr>`), but I never got around investigating how well that would work with expression parsing. If it does work, I think it would be clearer to someone reading code — but I also expect that opinions will vary.

soegaard:
For comparison Swift uses `..<` for the half-open interval and `...` for closed ranges.

https://developer.apple.com/documentation/swift/comparable/'.._(_:_:)
https://developer.apple.com/documentation/swift/comparable/'...(_:_:)

Issue #489, opened 2024-03-19
Giant green blobs in DrRacket


rfindler:
With this program:

```
#lang rhombus/and_meta

'abc(fun (v):
       if 1 = 2
       | 3
       | 4)'
```

it seems there are some implicitly inserted identifiers that have very large spans. I'm not sure if a change to Check Syntax is in order or to Rhombus so I wanted to open the issue to ask what folks think. The largest one in this program is a `parens` that starts at position 29 and goes to position 77, but this one doesn't seem to trigger big green blob. The second largest one is a `block` that goes from position 37 to position 76 which does seem to trigger a big green blob.

rfindler:
It looks like this is because of the documentation link that `block` gets. If that's the case, then maybe `block` should have a syntax location that's just the colon. That'd match what shows up in the docs, as following the link to the docs ends up [here](https://docs.racket-lang.org/rhombus/Block.html#%28def._%28%28lib._rhombus%2Fprivate%2Fcore..rkt%29._block%29%29).

mflatt:
Yes, the `block` identifier in the S-expression encoding of Shrubbery does get a source location that spans the block. I experimented some time back with making the source location just the `:`, but that didn't work out well for other reasons — like error reporting and `@rhombus` typesetting, if I remember correctly.

The underlying difference between source locations in plain S-expressions (i.e., syntax objects that start out as S-expressions) versus the S-expression encoding of shrubbery is that the shrubbery encoding avoids attaching information to pairs. Attaching information to pairs is always a little fragile, and more so with the shrubbery encoding because pairs don't correspond to parentheses in the source.

I'm not sure of the right choices here. If may be that `block` needs to have different source locations associated with it for different purposes, but it would be nice to avoid that kind of complexity.

rfindler:
It does seem like a few things are true:
- the current green blob isn't what we want
- the source locations produced by the reader should somehow match the actual shrubbery that the programmer wrote

What other degrees of freedom does that leave? 

Maybe there is a distinction that we should be drawing between identifiers that the programmer wrote and implicit ones (which would include this and also `#%app` on the racket side). If Check Syntax knew that it was looking at an identifier that has some different read-level manifestation, then it would know to treat it differently. Does that seem like a plausible direction to look for a solution?

mflatt:
> Maybe there is a distinction that we should be drawing between identifiers that the programmer wrote and implicit ones 

[...]

PR #491, opened 2024-03-21
identify some identifiers as to-be-ignored by check syntax

rfindler:
This makes some of the identifiers that were making the giant green blobs invisible to Check Syntax and seems to avoid the ones show in #489. I'm not sure if this is right, however. It also seems to get `#%parens` sometimes and maybe some others.

If you want to try it out, you'll need [this recent commit to DrRacket](https://github.com/racket/drracket/commit/150f105882fc64c484eb2664925215bbc4a7e128).

In this example:

```
def op = 5
'+'
```

we still get the arrow.

In this example:

```
def parens = 5
'(abc,
  
  def,

  ghi,

  jkl)'
```

we don't anymore.

I also notice when playing around, that there are some `#%body` that are being ignored, but the ones I saw didn't have source locations.

Overall, I'm very not sure this is the right change!

related to #489

### DIFF
--- a/shrubbery/lex.rkt
+++ b/shrubbery/lex.rkt
@@ -41,6 +41,7 @@
 
          syntax->token
          stx-for-original-property
+         stx-for-identifier-as-keyword
 
          current-lexer-source
 
@@ -213,6 +214,8 @@
           pending-backup))
 
 (define stx-for-original-property (read-syntax #f (open-input-string "original")))
+(define stx-for-identifier-as-keyword (syntax-property stx-for-original-property 'identifier-as-keyword #t))
+
 (define current-lexer-source (make-parameter "input"))
 
 (define (make-token name e start-pos end-pos [raw #f])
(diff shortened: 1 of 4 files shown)


rfindler:
Also, in this example:

```
#lang rhombus/and_meta

'abc(fun (v):
       if 1 = 2
       | 3
       | 4)'
```

we don't get the big blob anymore (I think all of the `block`s are hidden) and the purple arrows we get from the lang line to the syntax constant are to `fun`, `if`, and `=`.

mflatt:
Thanks! I do see that this change only covers some keyword-like identifiers. I'm happy to work on filling this out to take advantage of the new property.

rfindler:
Thanks! I wasn't sure if I got them all and also worried that I got too many! I wondered about [this spot](https://github.com/racket/rhombus-prototype/blob/master/shrubbery/lex.rkt#L215) in the lexer, whether it should also get this treatment. But I noticed that it goes on strings too, which probably isn't what we want.

mflatt:
Updated to complete the change.

PR #492, opened 2024-03-22
Provide key-equality variants of maps via `Map.by`

mflatt:
In the same way that the `Map` constructor creates a map that uses `==` and compatible hashing for keys, a new `Map.by(<key_comp>)` constructor creates a map that uses a comparison operation and compatible hashing function as bound to `<key_comp>`. Operators `<key_comp>` are bound in a new space, and the predefined `<key_comps>`s have the same names as comparison operators: `==` (which is the default), `===`, `is_now`, and
`is_same_number_or_object` (which is `eqv?`).

Examples:
```
  Map.by(===){ make(): "a", make(): "b" }
  Set.by(===){ make(), make() }
```

The `Map.by` form can be used in the same places as `Map`, including in binding patterns and as an annotation. A value constructed with `Map.by(<key_comp>)` satisifies the `Map` annotation as well as `Map.by(<key_comp>)`.

Ditto for `MutableMap`, `Set`, and `MutableSet`.

New key comparisons are bound with `key_comp.def` to implement custom hasing for a map. That functionality requires a small addition to Racket in the form of `unsafe-impersonate-hash`.


mflatt:
The relevant Racket change is racket/racket#4956. I'm ambivalent about that change, and just not having `key_comp.def` seems equally ok to me, so far.

rhombus/private/map.rkt
```diff
@@ -245,71 +312,130 @@
...
+          k
+          (lambda (k)
+            (#,(cadr build-convert-stxs)
+             v
+             (lambda (v)
+               (hash-set map k v))
+             (lambda () #f)))
+          (lambda () #f)))]))
 
 (define-static-info-syntax Map-pair-build
```
usaoc, line 440: The new variants should redirect to this function for static infos, like the set counterparts.

rhombus/private/map.rkt
```diff
@@ -364,33 +531,51 @@
...
+(define-repetition-syntax MutableMap.by
+  (repetition-transformer
+   (lambda (stx)
+     (parse-key-comp stx
+                     (lambda (stx arg-stxes str mapper)
+                       (parse-mutable-map stx #t
+                                          (key-comp-map-build-id mapper)
+                                          (key-comp-mutable-map-build-id mapper)))))))
 
 (define-static-info-syntax MutableMap-build
```
usaoc, line 573: Ditto.

rhombus/tests/map.rhm
usaoc, line 1: Correspondingly, the new variants should be tested for static arity, I think.

rhombus/tests/set.rhm
usaoc, line 1: Ditto.

usaoc:
It occurs to me that `.by(....)` variants are missing `.empty`, which makes it impossible to match an empty map/set when using `.by(....)` variants.

mflatt:
Thanks for the review, @usaoc! I think I've addressed your points except for `Map.by(...).empty`.

It would make sense to add `Map.by(...).empty`, as far as I can tell. But it's at least possible to match a specific empty list already with `Map.empty :: Map.by(...)` or `Map.empty && Map.by(...){}`.

usaoc:
Ah yes, I seem to have forgotten that nothing in `Map.empty` binding depends on the comparison mode.  This can be confusing, because `Map.empty` value is a specific map, while `Map.empty` binding is general.  Anyway, this can be clarified in the docs at some point.

PR #495, opened 2024-03-30
add some second-arguments to `rhombus`

rfindler:
This doesn't seem to actually work, however. There aren't any errors (unlike when I misspell `~annot` which does result in a compile-time error) but there are no links in the rendered documentation. I do see that, when I search for "::", there are results as both an expression and as an annotation operator. So probably this isn't quite complete somehow.

### DIFF
--- a/rhombus/scribblings/annotation.scrbl
+++ b/rhombus/scribblings/annotation.scrbl
@@ -26,7 +26,7 @@ the left-hand expression must produce a value that satisfies the
 right-hand annotation, otherwise a run-time exception is thrown. The
 @rhombus(is_a) operator takes an annotation like @rhombus(::), but it
 produces a boolean result indicating whether the result of the left-hand
-expression matches the annotation.
+expression satisfies the annotation.
 
 @examples(
   ~eval: ann_eval
@@ -39,7 +39,7 @@ expression matches the annotation.
 
 When @rhombus(class) defines a new class, an annotation can be
 associated with each field. When the annotation is written with
-@rhombus(::), then the annotation is checked when an instance is
+@rhombus(::, ~bind), then the annotation is checked when an instance is
 created.
 
 @examples(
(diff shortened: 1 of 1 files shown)


usaoc:
The links aren’t there because `::`/`:~` aren’t annotation operators, but are binding operators.  I don’t think the change to dotted names is quite right, because the dot is actually recognized symbolically (and, currently, documented dotted names are treated as a whole, along with the prefixes).  The *import* operator `.` is supposed to be used in, well, `import`, but you can’t change how dotted names are parsed, afaik.

rfindler:
Thanks!

rhombus/scribblings/annotation.scrbl
```diff
@@ -65,21 +65,21 @@ the @rhombus(.) operator can be chained for efficient access:
...
 
 More generally, @rhombus(.) access is efficient when the left-hand side
 of @rhombus(.) is an @rhombus(import) prefix, a @tech{namespace}, or
 an expression that can act as a @deftech{dot provider}. A class name
 acts as a namespace to provides access to
-field-accessor functions, as in @rhombus(Posn.x) (which doesn’t get a
-specific @rhombus(x), but produces a function that can be called on a
-@rhombus(Posn) instance to extract its @rhombus(x) field). An identifier
-that is bound using @rhombus(:~) or @rhombus(::) and a class name is
+field-accessor functions, as in @rhombus(#,(@rhombus(Posn, ~datum)).#,(@rhombus(x, ~datum))) (which doesn’t get a
```
usaoc, line 76: This is also a dotted name, not a field access.

rfindler: Oh geez.

rfindler:
Third try!

usaoc:
I amended the commit to add more revision to the subsection.  The part about `.` access is confusing, because `import`ed or namespaced names are *not* relevant; the part about `use_{static,dynamic}` is outdated.

PR #497, opened 2024-04-07
add `enum`

mflatt:
As an example,

```
HorizAlignment:
  left
  center
  right
```

defines `HorizAlignment` as an annotaton that is satisfied by `#'left`, `#'center`, and `#'right`, and it also defines `HorizAlignment.left`, `HorizAlignment.center`, and `HorizAlignment.right` as expressions bound to those symbols and as binding forms that match those symbols.

Sometimes, enumerated symbols are combined with other datatypes. Although a union of an enumerated datatype with other annotations could cover that case, `enum` provides direct support through `~is_a` clauses that add other annotations directly to the enumeration. Here's an example:

```
enum LineWidth:
  ~is_a Real
  inherit
```

The `pict` and `slideshow` libraries now use `enum` for various enumerations instead of a hand-rolled annotation.

### DIFF
--- a/rhombus/scribblings/ref-enum.scrbl
+++ b/rhombus/scribblings/ref-enum.scrbl
@@ -1,0 +1,48 @@
+#lang scribble/rhombus/manual
+@(import:
+    "common.rhm" open
+    "nonterminal.rhm" open)
+
+@title{Enumerations}
+
+@doc(
+  defn.macro 'enum $id_name:
+                $enum_clause
+                ...'
+  grammar enum_clause:
+    $id ...
+    ~is_a: $annot; ...
+    ~is_a $annot
+){
+
+ Defines @rhombus(id_name) as a @tech{predicate annotation} and as a
+ namespace, where the namespace exports each @rhombus(id) as the symbol
+ @rhombus(#'id) and as a binding form that matches the symbol
+ @rhombus(#'id). The annotation is satisfied by each symbol
+ @rhombus(#'id) and by values satisfying any of the @rhombus(annot)s
+ specified with @rhombus(~is_a).
+
+@examples(
+  ~defn:
+    enum Mouse:
+      itchy
+      mickey minnie
+      jerry
+  ~repl:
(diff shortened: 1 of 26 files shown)


rhombus/scribblings/ref-enum.scrbl
```diff
@@ -0,0 +1,48 @@
...
+  defn.macro 'enum $id_name:
+                $enum_clause
+                ...'
+  grammar enum_clause:
+    $id ...
+    ~is_a: $annot; ...
+    ~is_a $annot
+){
+
+ Defines @rhombus(id_name) as a @defterm{predicate annotation} and as a
```
usaoc, line 18: Should be `@term`.

rhombus/scribblings/ref-stxobj-meta.scrbl
```diff
@@ -14,8 +14,8 @@
                         at_stx :: Syntax)
     :: None
   fun syntax_meta.error(message :: ReadableString,
-                        at_stx :: Syntax,
-                        in_stx :: Syntax || List.of(Syntax))
+                        in_stx :: Syntax,
+                        at_stx :: Syntax || List.of(Syntax))
```
usaoc, line 18: This shouldn’t be included, I think.  (I see that there’s some problem in this doc, but I’ll fix it in a separate commit.)

usaoc:
I pushed a commit that adds some simple test cases for the new `enum` spacer.

rhombus/scribblings/ref-enum.scrbl
mflatt: Fixed.

rhombus/scribblings/ref-stxobj-meta.scrbl
mflatt: Removed.

mflatt:
> I pushed a commit that adds some simple test cases for the new `enum` spacer.

Thanks!

PR #498, opened 2024-04-08
fill in some primitive `Sequenceable.to_sequence` methods

usaoc:


rhombus/private/array.rkt
```diff
@@ -305,6 +306,21 @@
...
 
+(define-sequence-syntax Array.to_sequence/optimize
+  (lambda () #'Array.to_sequence)
+  (lambda (stx)
+    (syntax-parse stx
+      [[(id) (_ arr-expr)]
+       #'[(id) (in-vector (let ([arr arr-expr])
+                            (check-array 'Array.to_sequence arr)
+                            arr))]]
+      [_ #f])))
```
mflatt, line 317: Does this generate the same code as before, or does it add an extra layer of checking? How about in unsafe mode? I wonder whether this should instead be handled by rewriting the message from a existing check (although that might require updating `in-list` to compare as being in the `racket/primitive` realm instead of `racket`).

usaoc: It doesn’t generate the same code.  Rewriting requires some cooperation at Racket side, changing relevant checks in `racket/private/for` (to be more consistent with `in-treelist`, for that matter).

mflatt: I pushed a change `racket/private/for`.

usaoc:
Updated to use message rewriting.

mflatt:
Looks good to merge - thanks!

PR #499, opened 2024-04-10
list method adjustments, including adding `List.last` and renaming `List.take_left`

mflatt:
Changes:

* Add `List.last` and `PairList.last` as properties. Although `PairList.last` is not a constant-time operation, it's added as a property for consistency with `PairList.first` and so that when `xs` has an annotation `List.of(X)`, then `x.last` has the static info of `X`. (The static-information reason is why I didn't instead change `first` and `rest` to methods.)

* Rename `List.take_left` to `List.take`, rename `List.take_right` to `List.take_last`, and so on for `drop` and `PairList`. Racket has `take` and `take-right`, and I always have to think a little about which of those to use. Previously for Rhombus, I tried `take_left` instead of just `take`, because I thought that it was the asymmetry that made me have to think. After experience with that choice, I've concluded that the problem is was actually "left" and "right", which are disconnected from the words we usually use for lists. So far, the combination `take` and `take_last` has been the most comfortable for me.

* Add `List.find` and `PairList.find`, which were considered useful enough to be in `racket/treeilst`.

* Add an equality-function argument to `List.has_element` and `PairList.has_element`, like in `member` for Racket and `racket/treelist`.

* Use `racket/treelist` when available, otherwise use a copy of the current `racket/treelist` implementation.


rhombus/private/list.rkt
```diff
@@ -252,19 +264,31 @@
...
-  (check-nonempty-treelist who l)
-  (treelist-ref l 0))
+  #:primitive (treelist-first)
+  (treelist-first l))
+
+(define/method (List.last l)
+  #:primitive (treelist-last)
+  (treelist-last l))
 
 (define/method (List.rest l)
```
usaoc, line 274: `List.{first,last,rest}` can use `define/arity` instead, unless I’m missing anything.

rhombus/scribblings/ref-list.scrbl
```diff
@@ -481,6 +504,23 @@ it supplies its elements in order.
 }
 
 
+@doc(
+  fun List.find(lst :: List, pred :: Function.of_arity(1)) :: Any
+){
+
+ Returns the first element of @rhombus(lst) for which @rhombus(pred)
+ return true, @rhombus(#false) otherwise. Searching the list
```
usaoc, line 512: “returns”

rhombus/private/list.rkt
mflatt: Right - thanks!

rhombus/scribblings/ref-list.scrbl
mflatt: Fixed.

PR #502, opened 2024-04-27
experiment: `_` shorthand for anonymous functions

mflatt:
This is an implementation experiment (no docs, only a few illustrative tests) for an anonymous-function shorthand. I'm not especially attached to this proposal, but it seems like it could work, and maybe it will provoke better ideas.

In this experiment, two forms expand to a function:

* When the terms enclosed by an expression `()` include an immediate `_`, the `()` form (more precisely, the `#%parens` form) expands to an anonymous function, and each `_` is a distinct argument. So, for example `(_ + _)` is a function that takes two arguments and adds them.
* When a function-call expression form has an argument that is just `_`, then the function call (more precisely, the `#%call` form) expands to an anonymous function that contains the call, and each argument that is just `_` becomes an argument to that enclosing function. So, `List.cons(_, [1, 2, 3])` is a function equivalent to `fun (x): List.cons(x, [1, 2, 3])`. This conversion does not apply to a function call that has spliced arguments or `...`s.

The conversion in either case here works at the level of shrubbery, not parsed content. For example, `(1 2 % _)` will turn into something like `fun (x): (1 2 % x)` independent of whether a variable `x` makes sense after the `%` operator. However, that `x` to replace `_` is injected as an already parsed expression, so it won't be misused as a plain identifier or symbol. For example, `(_ . _)` or `(#' _)` expands into a syntax error, and not a function that looks up an field with an unspecified name or a quote form that returns a symbol with a name derived from a generated argument.

I think the function-call approach here is similar to mflatt/shrubbery-rhombus-0#4, but the operator approach is different.

There is no guarantee that the `_` conversion is unambiguous. For example, to get the same behavior from `(#' _)` before this patch, you might resort to `values(#' _)` to produce the symbol whose name is an underscore. As another potential example, suppose there's a `with` form using the syntax `with <bind> = <expr>: <body>`; in that case, `(with _ = ignored(): "result")` would not parse as intended. I don't think there are any forms that have a shape like `with` right now, and so we could conceivably just avoid creating forms that risk ambiguity, but I might have overlooked more interesting problems that exists already.

More examples:

 * `(1 + 2 * _)` is like `fun (x): 1 + 2 * x`
 * `((_ + 1) * 2)` would try to multiply a function by 2, so the shorthand just doesn't work
 * `(if _ | 1 | 2)` is like `fun (x): if x | 1 | 2`
 * `(if _ | _ | _)` is a syntax error, because the second and third `_` are inside blocks and not immediate in the shrubbery structure within the parentheses
 
The patch here also adds `~>` as a threading operator, since `_` shorthands are meant to play well with threading.
```
expr.macro '$arg ~> $func':
  '(fun (x): $func(x))($arg)'
```

So, `1 ~> (_ + 2)` produces 3, for example.

### DIFF
--- a/rhombus/private/class-dot.rkt
+++ b/rhombus/private/class-dot.rkt
@@ -307,14 +307,15 @@
         (define rator #`(method-ref #,name-ref-id #,obj-id #,idx))
         (define r (and ret-info-id
                        (syntax-local-method-result ret-info-id)))
-        (define-values (call new-tail)
+        (define-values (call new-tail to-anon-function?)
           (parse-function-call rator (list obj-id) #`(#,obj-id (tag arg ...))
                                #:static? (is-static-context? #'tag)
                                #:rator-stx (datum->syntax #f name #'rator-id)
-                               #:rator-arity (and r (method-result-arity r))))
+                               #:rator-arity (and r (method-result-arity r))
+                               #:can-anon-function? #t))
         (values (let ([call #`(let ([#,obj-id (rhombus-expression self)])
                                 #,call)])
-                  (if r
+                  (if (and r (not to-anon-function?))
                       (wrap-static-info* call (method-result-static-infos r))
                       call))
                 #'tail)]
@@ -334,7 +335,7 @@
        [(rator-id (~and args (tag::parens self arg ...)) . tail)
         #:when (not (syntax-parse #'self [(_ kw:keyword . _) #t] [_ #f]))
         (define obj-id #'this)
-        (define-values (call new-tail)
+        (define-values (call new-tail to-anon-function?)
           (parse-function-call proc-id (list obj-id) #'(#,obj-id (tag arg ...))
                                #:static? (is-static-context? #'tag)
                                #:rator-stx (datum->syntax #f name #'rator-id)))
@@ -462,7 +463,7 @@
             (cond
               [add-check
                (define obj-id #'obj)
-               (values pos/id obj-id #f (lambda (e)
+               (values pos/id obj-id #f (lambda (e to-anon-function?)
                                           (define static-infos (extract-static-infos e))
                                           (wrap-static-info*
                                            (relocate+reraw form1
@@ -471,7 +472,7 @@
                                                                #,(unwrap-static-infos e)))
                                            static-infos)))]
               [else
-               (values pos/id form1 #f (lambda (e) e))])]
+               (values pos/id form1 #f (lambda (e to-anon-function?) e))])]
            [else
             (define obj-id #'obj)
             (define r (and ret-info-id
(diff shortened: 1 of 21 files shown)


samth:
A couple examples that were tricky for me were `_(1)` and `_[1]`, how do those go here?

I would suggest `|>` for threading, since that has become pretty common in other languages such as F#, OCaml, and JavaScript. 

There's more discussion at https://github.com/racket/rhombus-prototype/discussions/203 from which I haven't drawn any immediate conclusions but maybe there's more to find there.

sorawee:
I support `|>` as well.

mflatt:
> A couple examples that were tricky for me were `_(1)` and `_[1]`, how do those go here?

`(_(1))` is `fun (x): x(1)` and `(_[1])` is `fun (x): x[1]` here, while `_(1)` by itself is a syntax error. I don't know whether that's what you have in mind.

The latest push changes `~>` to `|>`.

samth:
That seems consistent with everything else so it sounds like it works right.

usaoc:
Bikeshedding before I actually look into the impl: the term “pipeline” is more common and, potentially, less confusing than “threading”.  (This is also what’s usually associated with `|>`.)

mflatt:
I adjusted the implementation name to `pipeline.rhm`, and also filled in pieces needed to propagate static information and static-versus-dynamic mode for checking call arities.

benknoble:
- Should there be a way to disambiguate `(_ / _)` to mean `fun (x y): y / x`? In Clojure's `#()`, for example, `%` is roughly like `_` (but only for the first argument), but `%<n>` with `<n>` a number refers to a (numbered) argument, and `%&` designates a rest arg. Their docs explicitly call out

    > This is not a replacement for [fn](https://clojure.github.io/clojure/clojure.core-api.html#clojure.core/fn) - [sic] idiomatic use would be for very short one-off mapping/filter fns and the like. #() forms cannot be nested.

    Arguably permitting disambiguation like this encourages overly complex and confusing shorthands.

- The example `(_ / _)` reminds me of [Scala's anonymous functions](https://scala-lang.org/files/archive/spec/3.4/06-expressions.html#placeholder-syntax-for-anonymous-functions), where subsequent placeholders bind subsequent parameters. I find this at least somewhat confusing coming from fancy-app via Qi, where a single placeholder means "the lone argument" and `__` means "all the arguments."

- Scala is probably a better example to learn from than Clojure given that Scala is marginally closer in syntax than Clojure?

- Clojure avoids the `with` problem partly by marking anonymous `fn`s specially (`#()` instead or ordinary `()`). Scala does not do so and so probably has a similar problem. Since we're already using `#%parens`, what about something like `(# )` specifically marking places where `_` becomes a placeholder?

mflatt:
@benknoble Others should weigh in, because I don't have a lot of experience with this kind of shorthand myself, but for what it's worth:

I started to favor multiple `_`s meaning different arguments (that are used in order) when I saw that it provides a compact notation to turn a binary operator into a function. I agree that adding annotations to do other things (like use arguments in reverse order) would probably strain the idea of a shorthand.

I don't have a strong opinion about whether an extra character inside or outside `()`s to trigger `_` handling would be a good idea. But attaching the behavior to `#%parens` and requiring `()` around `_` is already heavier than I think people had in mind; if we go with something that requires `()`, I lean toward trying just `()` to see whether it causes problems.

mflatt:
Now with documentation, including a guide section. Based on the May 2 meeting discussion, I plan to merge this tomorrow (Utah time), and we can see how it works out.

samth:
Looking at the docs now, I see that `[1, 2, 3].map(_ + 1)` doesn't work and an extra pair of parens is needed. Is that fundamental to the approach, or would it be easy to change if desired?

usaoc:
> Looking at the docs now, I see that `[1, 2, 3].map(_ + 1)` doesn't work and an extra pair of parens is needed. Is that fundamental to the approach, or would it be easy to change if desired?

We discussed this in the last (2nd May) meeting, but the idea was to avoid ambiguity.  Indeed, you can imagine `[1, 2, 3].map(_ + 1)` meaning either `fun (x): [1, 2, 3].map(x + 1)` or `[1, 2, 3].map(fun (x): x + 1)`, and either interpretation would make sense if we were to allow this kind of “nested” notation.

PR #503, opened 2024-04-28
Add `WeakMutableMap` and `WeakMutableSet`; also add `later_of` and `now_of`

mflatt:
A weak mutable map is an ephemeron map at the Racket level. Although there are uses that need a non-epheneron weak hash table, they're pretty low-level and related to memory management. Meanwhile, Racket CS's ephemeron-hash performance is about the same as for a weak hash table. So, there's doesn't appear to be a need for non-ephemeron weak maps.
    
A weak set uses a non-ephemeron hash table, though, since it's equivalent to an ephemeron hash table for boolean values, and it's a little simpler for the GC to manage.

### DIFF
--- a/rhombus/private/annotation.rkt
+++ b/rhombus/private/annotation.rkt
@@ -320,6 +320,11 @@
          #`(#,@(info-maker c-static-infoss)
             . #,static-infos))]
        [(c::annotation-binding-form ...)
+        (unless (identifier? binding-maker-id)
+          (raise-syntax-error #f
+                              (or binding-maker-id
+                                  "argument converter annotations are not supported")
+                              new-stx))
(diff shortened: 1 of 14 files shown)


mflatt:
A second comment here adds `later_of` and `now_of` annotation constructors.

The `later_of` annotation constructor works for both immutable and mutable collection variants, but requires predicate annotations for keys and values in the cas of immutable collections. The `now_of` annotation constructor is analogous to `of`, but for mutable collections. For example, `List.later_of(String)` as statisfied by any list, but converts it to one where operations to access or add a value are checked to ensure that the value is a string.

The two commits now in this PR are nearly independent, but the implementation and documentation end up being adjacent and a little mixed, so I'm just combining them here.

PR #504, opened 2024-05-02
Try an implementation of integer range objects

usaoc:
(*See also [the proposal](<https://github.com/usaoc/rhombus-prototype/blob/2ad95b1d5457045ef095edd1acb4a4ce17339359/design/range/range.md>).*)

Range objects can be constructed and matched by nine constructors (but no separate annotation is given for each constructor, to avoid the combinatorial explosion).  ~~The constructors are named with `from`–`to` vs. `above`–`below`, which is taken from Common Lisp `loop` for inclusive vs. exclusive bounds.~~  The name `Range.from_to` is chosen for the common case `..`, and other names build on it.  Four operators `..`, `..=`, `<..<` (this is unfortunate, because operators cannot end in `.`), and `<..=` are available as shorthand notations for the nine constructors.

Range objects are currently restricted to integers, because it’s not apparent how to make them generic while still efficient in the context of Rhombus.

~~`Range.from_below`, `Range.from_to`~~ `Range.from_to`, `Range_from_to_inclusive`, and `Range.from` can be used as sequences, and a naïve optimization strategy is applied to ensure that they work as efficiently as before.  The strategy simply looks for a direct application of the three constructors and dispatch to special cases that work like `in-range`, `in-inclusive-range`, and `in-naturals`.  The same strategy is also found in `Range.step_by`, but only when it’s invoked as a dot method (over a direct application).

~~`Range.from_below` and `Range.from_to`~~ `Range.from_to` and `Range.from_to_inclusive` are `Listable`, which is the same as before.

Accordingly, ranges are separated into three subtypes: `Range` (all ranges), `SequenceRange` (ranges that can be used as sequences), and `ListRange` (ranges that are `Listable`).

A list of methods is implemented, which is essentially borrowed from Rebellion.  In addition, ranges can be deconstructed through `Range.start`, `Range.end`, `Range.includes_start`, and `Range.includes_end`, if one wants to.

### DIFF
--- a/design/range/range.md
+++ b/design/range/range.md
@@ -1,0 +1,279 @@
+# Ranges
+
+## Summary
+This proposal discusses ranges, discrete or continuous, with or
+without inclusive or exclusive upper/lower bounds.  Specifically, when
+we speak of ranges, we refer to ranges as *objects* with a reasonable
+interface.
+
+## Motivation
+A *range*, or an *interval*, represents a contiguous sequence of
+values.  The use of ranges is widely common in programming due to its
+very general idea—whenever a (partial) order is available, or in
+Rhombus terms, a class implements `Comparable`, we can speak of
+ranges.  As an example, the sub-list/string/… operation takes an
+(optional) lower bound and an (optional) upper bound, which can as
+well be represented by a range of non-negative integers.  This is not
+only intentionally better than two ad hoc parameters `start` and
+`end`, but also practically better as one no longer has to explicitly
+verify that `start <= end`.
+
+We can also speak of *discrete* ranges, which can be enumerated,
+stepped, etc.  This is the notion of ranges that is more familiar to
+most people: discrete ranges can be used as sequences by stepping from
+the starting point to the ending point.  The above example, in fact,
+is better expressed with discrete ranges, because the bounds must be
+indexes, i.e., interpreted as natural numbers.
+
+## Current Status
+
+### Racket
+Racket has no ranges in our sense.  The sequence constructor
+`in-range` actually produces a sequence of (real!) numbers, with an
+optional step.  In other words, Racket
+
+```racket
+(in-range start end step)
+```
+
+is somewhat like Haskell
+
+```haskell
+unfoldr
+  (\x -> if (x < end) then Just (x, x + step) else Nothing)
+  start
+```
+
+or, somewhat like C (except the lack of generic numerics)
+
+```c
+for (int i = start; i < end; i += step) {
+    ....
+}
+```
+
+*when `step` is non-negative* (use `>` instead when `step` is
+negative).  The function `range` is just like `in-range`, but produces
+a list.
+
+There are also sequence constructors `in-inclusive-range`, which
+includes the `end` (by using `<=`/`>=`), and `in-naturals`, which is
+infinite, but curiously only applies to non-negative integers, i.e.,
+naturals.
+
+### Rhombus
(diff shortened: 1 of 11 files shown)


design/range/range.md
```diff
@@ -0,0 +1,248 @@
...
+
+Ranges will have nine types, with the following notations:
+
+- `Range.from_below(start, end)`, which is closed–open `start .. end`;
+- `Range.from_to(start, end)`, which is closed–closed `start ..= end`;
+- `Range.from(start)`, which is closed–unbounded `start ..`;
+- `Range.above_below(start, end)`, which is open–open `start <..<
+  end`;
+- `Range.above_to(start, end)`, which is open–closed `start <..= end`;
+- `Range.above(start)`, which is closed–unbounded `start <..<`;
```
bennn, line 183: open-unbounded

usaoc:
The last commit extends some operations to *n*-ary cases where it makes sense.

Matthew suggested changing the names of `Range` constructors, specifically with more descriptive words like `exclusive`, `inclusive`, and `unbounded`, but I found the names to be way too long (and, in fact, not very easy to tell apart at a glance) after the change.  I agree that the `from`–`to` vs. `above`–`below` convention can be confusing, but I’m not sure what’s the correct change here.

In the proposal, I mentioned that the sub-list/string/… operations can use ranges.  Currently, there are built-in operations that take starting/ending positions in some weird fashion, say `String.utf8_bytes` and friends, or `Array.copy_from`.  I don’t know how many or how much we should change built-in operations.

usaoc:
The last commit changes some method names to be more consistent with existing built-in methods, and the proposal is updated to reflect the changes.

I think an indexing syntax for sub-list/string/… is more expected, as in other languages, but doing so requires changing built-in `Indexable.get` methods.  Currently, these methods are inlined to primitive operations, so changing them will incur a performance penalty.  I’m not sure how much we want that.

mflatt:
This looks very good overall.

The names I had in mind are
* `Range.from_to` = `..`
* `Range.from` = suffix `..`
* `Range.to` = prefix `..`
* `Range.from_exclusive_to` = `<..<`
* `Range.from_to_inclusive` = `..=`
* `Range.from_exclusive_to_inclusive` = `<..=`

The idea that you generally want a range starts as inclusive and range ends as exclusive, so only call out the unusual cases. Along those lines, I especially think that the most obvious infix operator `..` should be equivalent to the most obvious name `from_to`. The long names don't bother me, since the operator forms are the short ones.

The current choice is inconsistent with `Int.in`, but I expect that should just be changed to use ranges. But what notation should `Real.in` use?

`String.utf8_bytes` seems fine to change to accept a range. I'm inclined to have `String.substring` accept either one `ListRange` argument or two index arguments. I'd just leave the `Bytes.copy_from` and `Array.copy_from` methods alone, since those are imperative and performance-oriented, anyway.

I notice that if I give bad arguments to the `..` operator, it complains as `Range.from_below`. I think it it's worth some extra plumbing to make the error say `..` when `..` is used and (using `from_to` instead of `from_below`) `Range.from_to` when `Range.from_to` is used. It seems fine to me that the result of `Range.from_to(1, 2)` would print as `1 .. 2`, though.

usaoc:
The PR is now ready for merge.

I didn’t have any good alternative, so I used the suggested names.  I’d rather leave the `String.substring` change to another PR, since this PR is already quite large.

mflatt:
I agree — it makes sense to merge this and take things from there. Thanks!

PR #508, opened 2024-05-23
add `immediate_callee` space and use for pipelines

mflatt:
Also, generatize `space.enforest` and `space.transformer`, and add functions like `expr_meta.relative_precedence` and `expr_meta.ends_parse`.

Few Rhombus users will care about these mechansisms, but the upshot is that the `|>` pipeline operator parses it's right-hand side as an "immediate callee" so that it can pass along static information from it's left-hand side. This enables more propagation of static information than would normally happen with a function call. For example,

```
 "apple" |> fun (x): x[0]
```

and

```
 "apple" |> (_[0])
```

and

```
 "apple" |> (((fun (x): x[0])))
```

all work with `use_static`, because `fun` and `#%parens` (both in functon-shorthand mode and in grouping mode) are bound as immediate callee forms that end up communicating to `x` that it's a string.

Much of this commit is about generalizing the enforestation and transformer pipelines to support syntax-class arguments and deliver those arguments to macros. That generalization turned out not to be necessary for immediate callees, since they fit into the transformer protocol, amd enough plumbing was in place already for transformers. Nevertheless, the changes and exploration toward immediate callees suggested the generalization, and it's bound to be useful in the near future.

The protocol for parsing immediate callees and falling back to expression mode does not fit nicely into the Rhombus parsing framework. For example, in `x |> (_ + 1)(2)`, the `(_ + 1)` is just to the right of `|>`, but it's not the immediate callee. To make these kinds of forms work out, `#%parens` and similar need `expr_meta.ends_parse` to "look ahead" for parsing. The functionality seems likely to be useful in other contexts.


rhombus/private/immediate-callee-macro.rkt
```diff
@@ -0,0 +1,132 @@
...
+
+(begin-for-syntax
+  (define-name-root immediate_callee_meta
+    #:fields
+    (Parsed)))
+
+(define-identifier-syntax-definition-transformer macro
+  rhombus/immediate_callee
+  #:extra ([#:static_infos #`((#%index-result-key #,(get-syntax-static-infos))
+                              #,(get-treelist-static-infos))
```
usaoc, line 41: Should be a dotted tail.

PR #514, opened 2024-06-20
use generic `is_less` argument in `{,Mutable,Pair}List.sort` methods

usaoc:
Change the default `is_less` argument to the generic `Comparable` version.  This makes a plain `.sort` slower, but probably a performance-aware `.sort` should use a specialized `is_less` anyway.


mflatt:
LGTM

PR #515, opened 2024-06-21
shrubbery: improve navigation commands on S-exprs and `@`-notations

usaoc:
Treat `#{}` S-exprs as a whole, but treat `@{}` opener–closer like regular opener–closer pairs.


mflatt:
Sounds good to me — thanks!

PR #537, opened 2024-08-09
Implement `Comparable` for `Path`

samdphillips:


samdphillips:
I'm going to add Equatable to this PR.

rhombus/rhombus/scribblings/ref-comparable.scrbl
```diff
@@ -11,9 +11,9 @@
 A @deftech{comparable} value is one that supports @rhombus(<),
 @rhombus(<=), @rhombus(>=),@rhombus(>=), @rhombus(compares_equal), and
 @rhombus(compares_unequal). Real numbers, @tech{characters},
-@tech{strings}, @tech{byte strings}, @tech{symbols}, and @tech{keywords}
-are all comparable, as are instances of classes that implement
-@rhombus(Comparable, ~class).
+@tech{strings}, @tech{byte strings}, @tech{symbols}, @tech{keywords},
+and paths are all comparable, as are instances of classes that
```
usaoc, line 15: Could you add a simple description (we can expand it later) to “Paths” along with a `deftech` so that this link would work?  Something like `A @deftech{path} value represents a filesystem path`.

rhombus/rhombus/tests/path.rhm
```diff
@@ -17,3 +17,40 @@ check:
...
+  check p != p ~is #false
+  check p != q ~is #true
+  check p compares_equal p ~is #true
+  check p compares_equal q ~is #false
+  check p compares_unequal p ~is #false
+  check p compares_unequal q ~is #true
+  check p <= p ~is #true
+  check p <= q ~is #true
+  check p >= p ~is #true
+  check p >= q ~is #false
```
usaoc, line 56: Nit: I think the diff is showing a missing newline here.

usaoc:
> Or rather, I'm working out how `==` would apply here and I'm thinking it does not because Paths are not `immutable?`.

I’d be surprised if it didn’t apply, because `equal-always?` applies to paths just fine, and currently `==` also works fine.  The `immutable?` predicate isn’t general, despite what its name might suggest).  Paths aren’t mutable by any means, even in Racket.

samdphillips:
> I’d be surprised if it didn’t apply, because `equal-always?` applies to paths just fine, and currently `==` also works fine. The `immutable?` predicate isn’t general, despite what its name might suggest. Paths aren’t mutable by any means, even in Racket.

I'm not seeing `equal-always?` working in the way I would imagine (I was only guessing it had to do with `immutable?`)
```
> (equal-always? (string->path "/") (string->path "/"))
#f
```

LiberalArtist:
> > I’d be surprised if it didn’t apply, because `equal-always?` applies to paths just fine, and currently `==` also works fine. The `immutable?` predicate isn’t general, despite what its name might suggest. Paths aren’t mutable by any means, even in Racket.
> 
> 
> 
> I'm not seeing `equal-always?` working in the way I would imagine (I was only guessing it had to do with `immutable?`)
> 
> ```
> 
> Welcome to Racket v8.13 [cs].
> 
> > (equal-always? (string->path "/") (string->path "/"))
> 
> #f
> 
> ```

That was a bug fixed in https://github.com/racket/racket/commit/5396f5ce604e5f22bf5b5991fd468085a4ffd3b4

samdphillips:
> That was a bug fixed in [racket/racket@5396f5c](https://github.com/racket/racket/commit/5396f5ce604e5f22bf5b5991fd468085a4ffd3b4)

Thanks Philip.  Not running snapshot strikes again.

usaoc:
If some tests are failing on older versions of Racket and workarounds aren’t worth it to implement, they should be put under a version guard instead.  That’s what we currently do to partially support older versions.

samdphillips:
> If some tests are failing on older versions of Racket and workarounds aren’t worth it to implement, they should be put under a version guard instead. That’s what we currently do to partially support older versions.

Sounds good.  I'll get that set up.

mflatt:
Thanks!

PR #544, opened 2024-08-15
more consistently handle modes in printing functions

usaoc:
- `Port.Input.read_line` is changed to take a keyword `~mode` argument, more akin to other printing functions.

- The `enum` `PrintDesc.SpecialMode` is defined for use in `PrintDesc.special`.


mflatt:
LGTM - thanks!

PR #549, opened 2024-08-24
regular expressions

mflatt:
This regular expression sublanguage is based on @CooperCorad's BS thesis.

Rendered documentation: https://users.cs.utah.edu/~mflatt/tmp/rhombus-rx/rhombus/regexp.html

Different from Racket:

* The `rx` sublanguage is encoded in shrubbery notation, not a string. There's also a subsublanguage for character sets (i.e., the subsublanguage within `[]`).

* Matching is whole-input by default. To select partial-input matching, use `rx_in` to construct a regexp or use a method like `RX.match_in` instead of `RX.match`.

* The primary syntax for a capture group gives it a name, instead of just a position. These names are bound as variables when `rx` or `rx_in` is used as a binding form, and they serve as symbol keys in a map (within an object) produced by methods like `RX.match`.

* The `rx` form is not provided by just `#lang racket`. Instead, use `import rhombus/rx open`, which introduces only a few bindings in the expression and binding spaces but many bindings in the regexp-pattern and character-class spaces.

Inherited from Racket:

* A regexp matches either at the byte or character level. Both kinds can be used on bytes, byte strings, and input ports.

This draft includes some changes relative to the version shown in at the August 22 meeting (#180) based on feedback in that meeting — so, thanks to Robby, Alex, and Ben.

The documentation has lots of examples, and you find more in the test suite: https://github.com/mflatt/rhombus/blob/rx/rhombus/rhombus/tests/rx.rhm


rhombus/rhombus/scribblings/ref-rx-pattern.scrbl
```diff
@@ -0,0 +1,664 @@
...
+  rx'#"a"'.match(#"a")
+)
+
+}
+
+
+@doc(
+  rx.macro '$pat #%juxtapose $pat'
+  rx.macro '$pat ++ $pat'
+  rx.macro '$pat #%call ($pat)'
```
sorawee, line 47: This seems like implementation details that readers shouldn't need to know. See also my comment below about "organizing via surface constructs"

rhombus/rhombus/scribblings/ref-rx-charset.scrbl
```diff
@@ -0,0 +1,378 @@
...
+  rx'[("a" - "z") -- ("m" - "p")]'.is_match("n")
+  rx'[("a" - "z") -- ("m" - "p")]'.is_match("a")
+)
+
+}
+
+
+@doc(
+  rx_charset.macro '^ $charset'
+  rx_charset.macro '! $charset'
```
sorawee, line 141: Why are there multiple characters used for the same purpose? I think it's a waste of characters. And they are not syntactically meaningfully different.

rhombus/rhombus/scribblings/ref-rx-object.scrbl
```diff
@@ -0,0 +1,241 @@
...
+  method (regexp :: RX).match_range_in(....) :: maybe(RXMatch)
+  method (regexp :: RX).is_match(....) :: Boolean
+  method (regexp :: RX).is_match_in(....) :: Boolean
+){
+
+ Attempts to match a regular expression to @rhombus(input). For a regexp
+ created with @rhombus(rx), the entire content (between @rhombus(start)
+ and @rhombus(end)) must match for @rhombus(RX.match), while
+ @rhombus(RX.match_in) can match a portion of the input. For a regexp
+ created with @rhombus(rx), both @rhombus(RX.match) and
```
sorawee, line 59: This doesn't sound right to me...

rhombus/rhombus/scribblings/ref-rx-object.scrbl
```diff
@@ -0,0 +1,241 @@
...
+}
+
+
+@doc(
+  method (regexp :: RX).replace(
+    input :: String || Bytes,
+    insert :: (String || Bytes || Function.of_arity(1+num_captures)),
+    ~input_prefix: input_prefix :: Bytes = #""
+  ) :: String || Bytes
+  method (regexp :: RX).replace_all(....)
```
sorawee, line 196: I think some languages also allow users to provide multiple patterns and replacers in one query, since that's different than sequentially doing `replace_all` multiple times. Do we want to provide a similar operation?

sorawee (commented):
I only reviewed the documentation part (and it's mostly about the design, not documentation per se).

rhombus/rhombus/scribblings/ref-rx.scrbl
```diff
@@ -0,0 +1,117 @@
...
+ @rhombus(RX.match) only when the whole input string, byte string, or
+ port content matches the pattern. An @rhombus(rx_in) regexp matches
+ with @rhombus(RX.match) the same as with @rhombus(RX.match_in), which
+ means that it always can match against a portion of the input.
+
+@examples(
+  ~eval: rx_eval
+  ~repl:
+    rx'any*'
+    rx'any*'.match("abc")
```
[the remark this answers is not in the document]
mflatt: I'll figure this out eventually.

rhombus/rhombus/scribblings/ref-rx-pattern.scrbl
```diff
@@ -0,0 +1,664 @@
...
+    rx'any?'.match("abc")
+)
+
+}
+
+
+@doc(
+  rx.macro '$pat #%comp {$count}'
+  rx.macro '$pat #%comp {$min ..}'
+  rx.macro '$pat #%comp {$min .. $max}'
```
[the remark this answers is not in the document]
mflatt: I added a margin note to point out that `.. max` as an expression creates a range that starts at `#neginf`, and the intent is to avoid suggesting negative counts are possible.  We could say that `.. max` means something different here or that negative values are treated as 0, but I thought it would be better to just avoid it.

rhombus/rhombus/scribblings/ref-rx-charset.scrbl
```diff
@@ -0,0 +1,378 @@
...
+
+ The @rhombus(graph, ~at rhombus/rx_charset) character set has all ASCII
+ characters that print with ink. The
+ @rhombus(print, ~at rhombus/rx_charset) character set adds space
+ (@rhombus(Char.to_int) value 32) and tab (@rhombus(Char.to_int).
+}
+
+
+@doc(
+  rx_charset.macro 'cntrl'
```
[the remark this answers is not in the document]
mflatt: The names here all match the "Posix" ones in Racket, so maybe that's the original source. There are several names I don't like, but we should change morel or none of them.

rhombus/rhombus/scribblings/ref-rx-pattern.scrbl
```diff
@@ -0,0 +1,664 @@
...
+    rx'any{2..3}'.match("aaa")
+    rx'any{2..3}'.match("aaaa")
+)
+
+}
+
+
+@doc(
+  rx.macro '.'
+  rx.macro 'any'
```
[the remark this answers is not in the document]
mflatt: That could be the right choice. I've been torn between having `any` mean "any" and having it be the thing I use by default that is readable but does what I meant. Probably I should just accept using `.`.

rhombus/rhombus/scribblings/ref-rx-pattern.scrbl
```diff
@@ -0,0 +1,664 @@
...
+  expression is formed dynamically by merging the referenced regexp into
+  the enclosing pattern. (A limitation: both the merged regexp and
+  enclosing pattern must be free of backreferences, because backreferences
+  need to be converted from names to absolute positions eagerly.)
+
+  @examples(
+  ~eval: rx_eval
+  ~defn:
+    fun labeled(key) :: RX:
+      rx'$key ": " $name: .*'
```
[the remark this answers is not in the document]
mflatt: Adding parentheses to the example would make sense. Otherwise, I think our experience has been that it's better to write things more readably by convention than to impose constraints that are ad hoc in that way. Also, parentheses would be inconvenient when you want to bind a match result with `rx_in` as a binding form, such as writing `def rx'$m: pattern' = input` to bind `m` to the full match.

rhombus/rhombus/scribblings/ref-rx-pattern.scrbl
```diff
@@ -0,0 +1,664 @@
...
+
+ @item{When followed by an expression other than an identifier or
+  literal integer, then @rhombus($, ~at rhombus/rx) always forms a
+  @tech{splice}.}
+
+)
+
+}
+
+@doc(
```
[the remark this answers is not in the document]
mflatt: I don't have a strong opinion here, but unnamed capture groups seem to me a lot like the `_` shorthand: not what you want usually, but convenient in some contexts. And having both names and indices seems to hang together well overall, including the `RXMatch` level.

rhombus/rhombus/scribblings/ref-rx-pattern.scrbl
```diff
@@ -0,0 +1,664 @@
...
+  rx.macro '$ $int'
+  rx.macro '$ $expr'
+){
+
+ The @rhombus($, ~at rhombus/rx) operator is overloaded for multiple
+ uses:
+
+@itemlist(
+
+ @item{When not followed by anything, @rhombus($, ~at rhombus/rx)
```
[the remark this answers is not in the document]
mflatt: I haven't been entirely comfortable with `enable_newline`, and so I like this idea of adding `bol` and `eol`, getting rid of `^` and `$` as anchor patterns, and also getting rid of `enable_newline` and `disable_newline`.

rhombus/rhombus/scribblings/ref-rx-charset.scrbl
```diff
@@ -0,0 +1,378 @@
...
+  ~eval: rx_eval
+  rx'["a" "b" "c"]'.is_match("a")
+  rx'[("a" "b") "c"]'.is_match("a")
+)
+
+}
+
+
+@doc(
+  rx_charset.macro '$charset - $charset'
```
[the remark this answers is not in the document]
mflatt: I don't know how to make the constraint to a single character work. By the time the `-` operator has been found, it's left-hand side has been parsed as a charset.

rhombus/rhombus/scribblings/ref-rx-charset.scrbl
```diff
@@ -0,0 +1,378 @@
...
+){
+
+ Assuming that each @rhombus(charset) contains a single character,
+ creates a charset that has those two characters and all characters in
+ between (based on @rhombus(Char.to_int) values). An error is reported if
+ either @rhombus(charset) has zero or multiple characters.
+
```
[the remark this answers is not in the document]
mflatt: It is.

rhombus/rhombus/scribblings/ref-rx-charset.scrbl
```diff
@@ -0,0 +1,378 @@
...
+)
+
+}
+
+
+@doc(
+  rx_charset.macro '$charset && $charset'
+){
+
+ Creates a character set that has each character in both the first
```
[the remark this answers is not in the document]
mflatt: Yes, an empty charset always fails to match. There's the usual trade-off here between disallowing something that's surely an error when written directly versus allows macros that expand in a simple and general way.

rhombus/rhombus/scribblings/ref-rx-charset.scrbl
mflatt: I'd be in favor of picking one, but I wasn't sure whether `^` or `!` is better. I lean toward `!`, and I lean a little more in that direction if we also drop `^` as a anchor pattern.

rhombus/rhombus/scribblings/ref-rx-object.scrbl
mflatt: Sounds nice, if someone wants to implement that.

rhombus-lib/rhombus/private/amalgam/rx_match.rhm
```diff
@@ -0,0 +1,25 @@
+#lang rhombus/private/amalgam/core
+import:
+  "when_unless.rhm" open
+
+export:
+  RXMatch
+
+class RXMatch(whole, captures :: List, capture_names :: Map):
+  implements Listable
+  private implements Indexable
```
usaoc, line 10: Can we switch this implementation to public and put an annotation like `NonnegInt || Symbol` on `index`?  In general, I think `Indexable` should be publicly implemented so that the `.get` method can be properly documented.

rhombus-lib/rhombus/private/amalgam/rx_object.rhm
```diff
@@ -0,0 +1,189 @@
...
+         private _str):
+  internal _RX
+  constructor (): error("don't instantiate")
+
+  private implements Printable
+  private override describe(mode, recur) :: String:
+    _str
+
+  property handle: _rkt_rx
+  property in_handle: _rkt_partial_rx
```
usaoc, line 32: Are these properties meant to be public?  If so, they should be documented.

rhombus-lib/rhombus/private/amalgam/rx_match.rhm
mflatt: I agree about making the implementation public, and an annotation on `index` also sounds good.

rhombus-lib/rhombus/private/amalgam/rx_object.rhm
mflatt: Yes, I forgot to document these. I also meant to add `RX.from_handles`.

rhombus-lib/rhombus/private/amalgam/rx_match.rhm
mflatt: I made the `indexable` implementation public. I didn't add the annotation after all, because I'm not sure `capture_names` should be restricted to symbol keys. If it is, then it makes sense to use `Map.of` for `capture_names`, but that will add an extra cost to constructing an `RXMatch`. I'm undecided on whether this is a place to be more precise and have more checks versus a place to leave flexible with few checks.

PR #550, opened 2024-08-31
`error` and `error.message`

mflatt:
This change is backward-incompatible, because it makes the `who` argument to `error` a keyword argument with `~who`, and it renames `Exn.Fail.Contract` to `Exn.Fail.Annot`.

The main change is to add support for message details and clauses. Clause constructors like `error.val` and `error.annot` build the label–value pairs that go at the end of an error message. The `error.message` function can be used to just construct the message text that `error` puts into an exception.

This commit does not change the non-message content of exceptions, but the clause protocol is meant to accommodate and adapt to future changes.

Examples from the docs:

```
> error("oops")
oops
> error(~who: #'me, "oops")
me: oops
> error(~who: #'me,
        "oops",
        ~details: ["something has gone wrong;",
                   "see the manual for more information"])
me: oops;
 something has gone wrong;
 see the manual for more information
> error(~who: #'me,
        ~exn: Exn.Fail.Annot,
        error.annot_msg("fruit"),
        error.annot("Tropical"),
        error.val(~label: "fruit", #'Apple))
me: fruit does not satisfy annotation
  annotation: Tropical
  fruit: #’Apple
> error(~who: #'me,
        "mismatch between fruit and vegetable",
        error.val(~label: "fruit", [#'Apple, 0]),
        error.val(~label: "vegetable", [#'Lettuce, -1]))
me: mismatch between fruit and vegetable
  fruit: [#’Apple, 0]
  vegetable: [#’Lettuce, -1]
```

I expected to have various helper functions like `annot_error` to raise a standard "does not satisfy annotation" error, for example. But in changing existing error messages, I concluded that it's better to have functions for constructing arguments to `error`, such as the `error.annot_msg()` for constructing the main "does not satisfy annotation" message text. A drawback of this approach is that `error.annot_msg()` as an argument to `error` needs to be paired with `~exn: Exn.Fail.Annot` as well as (usually) `error.annot` and `error.val` clause constructions. But this approach still ended up being more comfortable than an `annot_error` function, because explicitly raised annotation errors often include small variations on the normal message (otherwise `::` would be used in the first place), and it seemed easiest to just remember and use the `error` protocol.

### DIFF
--- a/rhombus-lib/rhombus/private/amalgam/error.rhm
+++ b/rhombus-lib/rhombus/private/amalgam/error.rhm
@@ -1,0 +1,169 @@
+#lang rhombus/private/amalgam/core
+import:
+  "core-meta.rkt" open
+  "maybe.rhm" open
+  lib("racket/base.rkt")
+
+use_static
+
+export:
+  error
+
+namespace error:
+  export:
+    Who
+
+    // constructs full message without raising it
+    message
+
+    // builders of main message test
+    annot_msg
+
+    Clause
+
+    // clause builders
+    text
+    val
+    vals
+    annot
+
+    // helper
+    reindent
+
+annot.macro 'Who':
+  'ReadableString || Symbol || Name'
+
+class Clause(msg :: String):
+  nonfinal
+
+// primary `error` function
+// (not backward-comptible with current `error` due to `~who` keyword argument)
+fun error(~srcloc: srcloc :: maybe(Srcloc) = #false,
+          ~who: who :: maybe(error.Who) = #false,
+          ~realm: realm :: Symbol = #'rhombus,
+          // make things like `~exn: Exn.Fail.User` easy; for more complex
+          // exception constructions, use `error.message`, instead, to build
+          // just the message text
+          ~exn: exn :: (Function.of_arity(2)
+                          // also gets srcloc list
+                          || Function.of_arity(3)
+                          // also gets clause list
+                          || Function.of_arity(4)) = Exn.Fail,
+          // intended to be a single short line
+          msg :: ReadableString,
+          // if non-empty, ";" is added to the end of `message`, and
+          // each `detail` is shown on its own line with a 1-space prefix
+          ~details: [detail :: ReadableString, ...] = [],
+          // intended to be constructed by `error.val`, `error.annot`,
+          // or `error.text`; each `clause` starts its own line with
+          // a 2-space prefix:
+          clause :: error.Clause,
+          ...):
+  let msg = message(~srcloc: srcloc,
+                    ~who: who,
+                    ~realm: realm,
+                    msg,
+                    ~detail: [detail, ...],
+                    clause, ...)
+  let marks = Continuation.Marks.current()
+  let e:
+    match exn
+    | exn :: Function.of_arity(4): exn(msg, marks, [srcloc], [clause, ...])
+    | exn :: Function.of_arity(3): exn(msg, marks, [srcloc])
+    | exn :: Function.of_arity(2): exn(msg, marks)
+  throw e
+
+// exported as `error.message`, same arguments as `error` except for `~exn`
+fun message(~srcloc: srcloc :: maybe(Srcloc) = #false,
+            ~who: who :: maybe(error.Who) = #false,
+            ~realm: realm :: Symbol = #'rhombus,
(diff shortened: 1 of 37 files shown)


rhombus-lib/rhombus/private/amalgam/error.rhm
```diff
@@ -0,0 +1,164 @@
...
+
+// exported as `error.message`, same arguments as `error` except for `~exn`
+fun message(~srcloc: srcloc :: maybe(Srcloc) = #false,
+            ~who: who :: maybe(error.Who) = #false,
+            ~realm: realm :: Symbol = #'rhombus,
+            message :: ReadableString,
+            ~detail: [detail :: ReadableString, ...] = [],
+            clause :: error.Clause,
+            ...) :~ String:
+   (if srcloc | Srcloc.to_report_string(srcloc) ++ ": " | "")
```
usaoc, line 84: This should be prepended to the string below, if I understand correctly.

rhombus-lib/rhombus/private/amalgam/error.rhm
```diff
@@ -0,0 +1,164 @@
...
+   base.#{error-message->adjusted-string}(
+     (who && Symbol.from_string(to_string(who))),
```
usaoc, line 86: I’ve been wanting to ask this, but should we try to unify what is conventionally accepted as name argument at some point?  Now, `error` accepts a very lenient `ReadableString || Symbol || Identifier || Operator`, while some accept `ReadableString || Symbol` or just `Symbol`, and yet some accept just `String`.  Why can’t we just accept `String` and let user does their own conversion?  And why specifically `Identifier || Operator`, but not `Name`?

rhombus-lib/rhombus/private/amalgam/exn-object.rkt
```diff
@@ -74,27 +74,27 @@
...
 
 (define-exn Continuation exn:fail:contract:continuation
-  #:parent Contract exn:fail:contract
+  #:parent Annot exn:fail:contract
   #:fields ()
   #:children ())
 
 (define-exn Variable exn:fail:contract:variable
-  #:parent Contract exn:fail:contract
+  #:parent Annot exn:fail:contract
```
usaoc, line 97: It seems to me that these exceptions make less sense as “sub-exceptions” of `Exn.Annot`.

rhombus-lib/rhombus/private/amalgam/error.rhm
mflatt: Right - thanks!

rhombus-lib/rhombus/private/amalgam/error.rhm
mflatt: I agree about unifying and about using `Name` if `Identifier || Operator` remains relevant. I still think it will be handy to be flexible here to avoid boilerplate conversions in common cases. Maybe we should allow any value for the name, since `to_string` should generally do the right thing?

rhombus-lib/rhombus/private/amalgam/exn-object.rkt
mflatt: I see what you mean, and perhaps `Contract` is the right word in the sense that it's the run-time component of an annotation. It seemed awkward to have `~exn: Exn.Fail.Contract` but `annot` in all other parts, though. Maybe `Exn.Fail.Annot` should be a constructor that creates an `Exn.Fail.Contract` — that is, an alias for construction purposes, but not for  purposes like `is_a` (to avoid suggesting that an `Exn.Fail.Contract.Variable` instance, say, is a `Exn.Fail.Annot`)?

rhombus-lib/rhombus/private/amalgam/error.rhm
usaoc: An annotation like `Any.to_string` (akin to `Any.to_boolean`) can be useful if we do allow any value.

mflatt: I remember now why `Any` (or `Any.to_string`) is probably not a good choice in this case: because `#false` s supposed to mean that nothing should be prefixed, not that `#false: ` should be a message prefix.

mflatt:
Merging because this seems like progress, at least — but as usual, that doesn't mean it's the final word.

PR #553, opened 2024-09-24
Print immutable arrays/boxes as `.snapshot`s

usaoc:
This PR changes the printing of immutable arrays/boxes to the combination of mutable constructors plus `.snapshot` methods.  `Box.{copy,snapshot}` methods are added both to align with other primitive data types and to allow the printing.


mflatt:
LGTM

PR #555, opened 2024-09-25
`rhombus/runtime` -> `rhombus/memory`

mflatt:
Rename the `racket/runtime` module to `racket/memory` and adjust some exported names to not repeat `memory` in the name.


sorawee (commented):
There's a `base.#{collect-garbage}()` left in `measure.rhm`. Should that be changed to `memory.gc()`?

Also, s/racket/rhombus in the PR description/commit message.

mflatt:
Fixed - thanks!

PR #558, opened 2024-10-07
add `Char` and `Byte` expressions/bindings

usaoc:
Mainly, this change aims to provide a "literal"(-like) form to produce or match a character.  Previously, characters can be produced in pure shrubbery notation by using string indexing, but bindings don't work that way.  Alternatively, characters can be produced or matched in S-expression notation, by using `#{}` escape.  This is not very ideal.

Instead, this change provides a `Char` expression/binding form that expects a single-character string.  There are several benefits of this approach:

- it's not too much longer than using string indexing (just one character longer);
- it's in pure shrubbery notation, without needing to extend the lexical syntax;
- it statically verifies that the string *does* have exactly one character (because Unicode is confusing!).

Accordingly, printing of characters now uses `Char`.

A `Byte` expression/binding form is also provided, just to parallel `Char`, but it can be handy when dealing with byte strings.

A large part of this commit also changes existing code to use the new forms where it makes sense.

### DIFF
--- a/gui_demo.rhm
+++ b/gui_demo.rhm
@@ -123,7 +123,7 @@ def tabs:
   fun accum_key(ev :: gui.KeyEvent, area):
     keyed.update(fun (s :: String):
                    match ev.code
-                   | #{#\backspace} || #{#\rubout}:
+                   | Char"\b" || Char"\x7F" /* ASCII delete */:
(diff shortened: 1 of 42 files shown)


mflatt:
This looks like a good idea to me.

PR #559, opened 2024-10-08
add `->` and `Function.all_of` annotation forms

mflatt:
The `->` annotation operator constructs a converting annotation that expects a function of suitable arity and wraps it to check argument and result annotations. The `Function.all_of` annotation form takes any number of annotations constructed by `->` and creates a annotation that is satified by a function that handles all of `->` cases, similar to `case->` in Racket's contract system.

The expressiveness of `->` itself is similar to `->i` in Racket's contract system: it supports optional and keyword arguments, and arguments can be named to create dependent checks.

Unlike `case->`, which selects a case based only on the arity of a call, `Function.all_of` selects a case based on the first clause that matches the arguments of a given call. This makes `Function.all_of` more analogous to a multi-case `fun`, and it means that `Function.all_of` can express some dependencies even with simple `->` forms, as in

```
  Function.all_of(Int -> Int,
                  String -> String)
```

To get error reprting to work well overall, this commit makes a general change to the annotation-macro layer. Each annotation form is now responsible for attaching its rendered form to its expansion, like expression macros. So, there are more internal uses of `relocate+reraw`, and there's more need for `Syntax.relocate_span`s at the Rhombus level.

The `Any` annotation is like `any/c` in Racket, which means that `() -> Any` will wrap a function with a non-tail call to ensure that it produces a single result. Use `() -> (Any, ...)` or `() -> (& Any)` to avoid constraining the number of results and produce a tail-calling wrapper.

The `->` operator cannot by itself handle multiple arguments in parentheses on the left of the operator, since an infix operator takes a parsed left-hand side, and a parenthesized sequence of annotation is not an annotation. As a workaround, the `#%parens` form exported by `rhombus` cooperates with the `->` form.

### DIFF
--- a/rhombus-lib/rhombus/private/amalgam/annot-macro.rkt
+++ b/rhombus-lib/rhombus/private/amalgam/annot-macro.rkt
@@ -14,7 +14,8 @@
                      (submod "syntax-object.rkt" for-quasiquote)
                      "call-result-key.rkt"
                      "values-key.rkt"
-                     (for-syntax racket/base))
+                     (for-syntax racket/base)
+                     "srcloc.rkt")
          (only-in "space.rkt" space-syntax)
          "space-provide.rkt"
          (only-in "binding.rkt" :binding-form)
@@ -78,11 +79,11 @@
 (define-for-syntax (wrap-parsed stx)
   #`(parsed #:rhombus/annot #,stx))
 
-(define-for-syntax (parse-annotation-macro-result form proc)
+(define-for-syntax (parse-annotation-macro-result form proc #:srcloc [loc (maybe-respan form)])
   (unless (syntax? form)
     (raise-bad-macro-result (proc-name proc) "annotation" form))
   (syntax-parse (unpack-group form proc #f)
-    [c::annotation #'c.parsed]))
+    [c::annotation (relocate+reraw loc #'c.parsed)]))
 
 (define-for-syntax (make-annotation-infix-operator prec protocol proc assc)
   (annotation-infix-operator
@@ -100,7 +101,8 @@
                                    proc))
        (lambda (form1 form2 stx)
          (parse-annotation-macro-result (proc (wrap-parsed form1) (wrap-parsed form2) stx)
-                                        proc)))
+                                        proc
+                                        #:srcloc (datum->syntax #f (list form1 stx form2)))))
    assc))
 
 (define-for-syntax (make-annotation-prefix-operator prec protocol proc)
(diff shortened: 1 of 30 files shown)


mflatt:
Updated to support `~any` in place of a result annotation, which is like `any` (instead of `any/c`) in Racket contracts. Also, the argument side can be `(~any)`, which is like `unconstrained-domain->` in Racket.

samth:
Does this have any of the `racket/contract` function optimizations, or are those limited to uses of `fun` with domain and result annotations?

mflatt:
> Does this have any of the `racket/contract` function optimizations, or are those limited to uses of `fun` with domain and result annotations?

Racket contract optimizations are about avoiding indirections related to blame tracking and chaperoning to preserve equality. The `->` here for Rhombus doesn't do either of those, so no extra optimization is needed.

Examples below illustrate, with these results (CPU msec) on my machine:

| program | `f` | `f1` | `f2` | `f3` |
----------|-----|----|-----|------|
| `use-f.rkt` | 3 | 19 | 140 |  386 |
| `use_f.rhm` | 19 | 19 | 19 | 19 |

Looks like Rhombus is doing something that prevents inlining of the original `f`, so that's something to look into. But `->` doesn't create any measurable overhead in this case.

`f.rkt`:

```
#lang racket

(define int->int (-> exact-integer? exact-integer?))

(provide
 f
 f1 ;; can't inline
 (contract-out (rename f f2 ;; optimized contract
                       (-> exact-integer? exact-integer?)))
 (contract-out (rename f f3 ;; general contract
                       int->int)))

(define (f x)
  (+ x 1))

(define f1 f)
(set! f1 f1)
```

`use-f.rkt`:

```
#lang racket
(require "f.rkt")

(define-syntax-rule (time* f)
  (begin
    (collect-garbage)
    (time
     (for/fold ([v #f]) ([i (in-range 10000000)])
       (f 0)))))

(time* f)
(time* f1)
(time* f2)
(time* f3)
```

`f.rhm`:

```
#lang rhombus

export:
  f
  f1
  f2
  f3

fun f(x):
  x + 1

def mutable f1 = f
f1 := f1

fun f2(x :: Int) :: Int:
  f(x)

def f3 :: (Int -> Int) = f
```

`use_f.rhm`:

```
#lang rhombus
import:
  "f.rhm" open
  rhombus/measure

macro 'time $f':
  'measure.time:
     ~gc
     for values(v = #false) (i: 0..10000000):
       $f(i)'

time f
time f1
time f2
time f3
```

usaoc:
> Looks like Rhombus is doing something that prevents inlining of the original `f`, so that's something to look into.

The linklet form looks like this:

```racket
(define-values (f)
  (#%name f
    (lambda (x1_1)
      (let-values ([(x1_2) x1_1])
        (let-values ()
          (let-values ([(x_3) x1_2])
            (let-values ()
              (begin 'syntax (+ x_3 1)))))))))
```

So maybe we need to simplify the expansion more.

samth:
I see less slowdown than @mflatt but even this definition provokes it, so I don't think just simplifying the expansion will be enough:

```
(define-values (f**)
  (lambda (x1_1)
    (let-values ([(x) x1_1])
      (+ x 1))))
```

PR #562, opened 2024-10-14
Initial Path integration

samdphillips:
Some initial Path integration that I've been poking at for a couple of weeks now.

Initially, I used a veneer for `Path.read_with` and `Path.write_with` that were syntax and would take a block that would become the read (write) procedure.  Once I started thinking about adding the other flags for reading and writing the syntax seemed too messy so I went back to a simple wrapper around `call-with-input-file`/`call-with-output-file`.

### DIFF
--- a/rhombus-lib/rhombus/private/amalgam/annotation.rkt
+++ b/rhombus-lib/rhombus/private/amalgam/annotation.rkt
@@ -442,11 +442,12 @@
    (lambda (form tail)
      (syntax-parse tail
        [(op::name . (~var t (:annotation-seq #'::)))
+        (define op-name (string->symbol (shrubbery-syntax->string #'op.name)))
         (values
          (build-annotated-expression #'op.name #'t
                                      checked? form #'t.parsed
                                      (lambda (tmp-id)
-                                       #`(raise-::-annotation-failure 'op.name #,tmp-id '#,(shrubbery-syntax->string #'t.parsed)))
+                                       #`(raise-::-annotation-failure '#,op-name #,tmp-id '#,(shrubbery-syntax->string #'t.parsed)))
                                      wrap-static-info*)
          #'t.tail)]))
    'none))
--- a/rhombus-lib/rhombus/private/amalgam/closeable.rhm
+++ b/rhombus-lib/rhombus/private/amalgam/closeable.rhm
@@ -8,30 +8,33 @@ export:
   Closeable
 
 namespace Closeable:
-  export rename cdef as def
+  export:
+    close
+    rename cdef as def  
+
+  fun close(v :: Closeable):
+    v.close()
 
   defn.sequence_macro 'cdef $(bind :: Identifier) = $(rhs :: expr_meta.Parsed)
                        $body
                        ...':
+    ~op_stx: self
     def rhs_statinfo = statinfo_meta.gather(rhs)
-    def resource = Syntax.make_temp_id()
+    def resource = Syntax.make_temp_id("resource")
     def resource_with_statinfo = statinfo_meta.wrap(resource, rhs_statinfo)
-    def close_stmt:
-      '$(if syntax_meta.is_static(bind) | 'use_static' | 'use_dynamic')
-       $(resource_with_statinfo) . $('close'.relocate(rhs))()'
     values(
       '
         block:
           def mutable $resource = #false
           try:
             ~initially:
-              $resource := ($rhs :: Closeable)
+              $resource := ($rhs $('::'.relocate(self)) Closeable)
             block:
               let $bind = $resource_with_statinfo
               #void
               $body
               ...
             ~finally:
-              $close_stmt
+              ($resource :~ Closeable).close()
       ',
       '')
--- a/rhombus-lib/rhombus/private/amalgam/closeable.rhm
+++ b/rhombus-lib/rhombus/private/amalgam/closeable.rhm
@@ -1,0 +1,37 @@
+#lang rhombus/private/amalgam/core
+
+import:
+  "core-meta.rkt" open
+  "closeable.rkt" open
+
+export:
+  Closeable
+
+namespace Closeable:
+  export rename cdef as def
+
+  defn.sequence_macro 'cdef $(bind :: Identifier) = $(rhs :: expr_meta.Parsed)
+                       $body
+                       ...':
+    def rhs_statinfo = statinfo_meta.gather(rhs)
+    def resource = Syntax.make_temp_id()
+    def resource_with_statinfo = statinfo_meta.wrap(resource, rhs_statinfo)
+    def close_stmt:
+      '$(if syntax_meta.is_static(bind) | 'use_static' | 'use_dynamic')
+       $(resource_with_statinfo) . $('close'.relocate(rhs))()'
+    values(
+      '
+        block:
+          def mutable $resource = #false
+          try:
+            ~initially:
+              $resource := ($rhs :: Closeable)
+            block:
+              let $bind = $resource_with_statinfo
+              #void
+              $body
+              ...
+            ~finally:
+              $close_stmt
+      ',
+      '')
--- a/rhombus-lib/rhombus/private/amalgam/closeable.rhm
+++ b/rhombus-lib/rhombus/private/amalgam/closeable.rhm
@@ -10,12 +10,12 @@ export:
 namespace Closeable:
   export:
     close
-    rename cdef as def  
+    rename clet as let
 
   fun close(v :: Closeable):
     v.close()
 
-  defn.sequence_macro 'cdef $(bind :: Identifier) = $(rhs :: expr_meta.Parsed)
+  defn.sequence_macro 'clet $(bind :: Identifier) = $(rhs :: expr_meta.Parsed)
                        $body
                        ...':
     ~op_stx: self
@@ -31,7 +31,6 @@ namespace Closeable:
               $resource := ($rhs $('::'.relocate(self)) Closeable)
             block:
               let $bind = $resource_with_statinfo
-              #void
               $body
               ...
             ~finally:
--- a/rhombus-lib/rhombus/private/amalgam/closeable.rhm
+++ b/rhombus-lib/rhombus/private/amalgam/closeable.rhm
@@ -14,8 +14,10 @@ namespace Closeable:
 
   fun close(v :: Closeable):
     v.close()
+    #void
 
-  defn.sequence_macro 'clet $(bind :: Identifier) = $(rhs :: expr_meta.Parsed)
+  defn.sequence_macro 'clet $bind ... = $(rhs :: expr_meta.Parsed)
+                       $body0
                        $body
                        ...':
     ~op_stx: self
@@ -29,10 +31,12 @@ namespace Closeable:
           try:
             ~initially:
               $resource := ($rhs $('::'.relocate(self)) Closeable)
-            block:
-              let $bind = $resource_with_statinfo
-              $body
-              ...
+            Continuation.barrier:
+              let $bind ... = $resource_with_statinfo
+              block:
+                $body0
+                $body
+                ...
             ~finally:
               ($resource :~ Closeable).close()
       ',
--- a/rhombus-lib/rhombus/private/amalgam/path-object.rkt
+++ b/rhombus-lib/rhombus/private/amalgam/path-object.rkt
@@ -6,6 +6,7 @@
          "call-result-key.rkt"
          "define-arity.rkt"
          "compare-key.rkt"
+         (submod "annotation.rkt" for-class)
          (submod "string.rkt" static-infos)
          (submod "bytes.rkt" static-infos))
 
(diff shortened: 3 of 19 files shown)


mflatt:
Crazy idea: what if we want RAAI-style with a `Closeable.def` form?

```
block:
  Closeable.def in = Port.Input.open_file("x")
  ....
  // `in.close()` is added to the end here
```

The `Closeable.def` form could be implemented roughly like this:

```
namespace Closeable:
  export:
    rename cdef as def
  defn.sequence_macro 'cdef $bind ... = $rhs
                         $body
                         ...':
    values(
      '
        block:
          def mutable resource = #false
          try:
            ~initially:
              resource := $rhs
            block:
              let $bind ... = resource
              $body
              ...
              #void
            ~finally:
              resource.close
      ',
      '')
```

samdphillips:
RAII isn't that crazy and I intended to have a more general form.  I still forget sometimes that Rhombus macros can see subsequent forms in the same context.  I'll take a look at implementing this.

I'm thinking that next we may want to define a `Closeable` interface that classes can opt into. 

I agree with Ben that it might be better to return the last value instead of `#void`.  Also, would using `Closeable.def`cause the final expression of the block to not be a tail call?

mflatt:
Yes, it should definitely return the value. I needed `#void` to cover the empty-remainder case, but didn't think through that it should be at the beginning, not the end.

Another issue with this macro implementation is that static info from `$rhs` should flow to `resource`. Maybe statiic info will have to be propagated  explicitly by `Closeable.def`, forcing expansion of `$rhs` and using `statinfo_meta.gather`.

Note that as written, `$rhs` will be evaluated with breaks disabled. I think that's ok — anything not meant to be in non-breaking mode can be lifted out — but users of `Closeable.def` would need to be aware. Disabling breaks was intentional here to ensure that either the resource is closed on escape or it was not allocated (as long as the running thread isn't killed).

I agree about having a `Closeable` interface that other classes can implement, and maybe `Closeable.def` should require a `Closeable` object. The fact that `close` isn't called as a method above is a bug, of course.

samth:
I've been writing some path-related code recently, so this is helpful. How much of the `racket/path` library should be exposed as `Path` methods? Do we want something minimal here, or something that exposes basically all of `racket/path` (and maybe much of `racket/file`)?

samdphillips:
Added the `Closeable` interface and `Closeable.def` syntax.  Feedback appreciated.

samdphillips:
@samth wrote:
> I've been writing some path-related code recently, so this is helpful. How much of the `racket/path` library should be exposed as `Path` methods? Do we want something minimal here, or something that exposes basically all of `racket/path` (and maybe much of `racket/file`)?

I've looked at what other languages (specifically the Python pathlib) have and slowly added those as I've needed them.

samdphillips:
Is there value in keeping `Path.read_with` and `Path.write_with`?

mflatt:
I pushed some adjustments to `Closeable`, since the interplay between Racket and Rhombus layers there is tricky.

mflatt:
> Is there value in keeping `Path.read_with` and `Path.write_with`?

I'm in favor of dropping those.

mflatt:
Besides`Closeable`, it looks like `to_complete_path` is not yet documented, and probably that should be `to_absolute_path`. 

Should `AbolutePath` and `RelativePath` be annotations (which are easier to use in patterns)? And then drop `Path.is_absolute`?

samth:
I added one more operation (`Path.only`)

rhombus-lib/rhombus/private/amalgam/closeable.rhm
```diff
@@ -0,0 +1,40 @@
...
+          try:
+            ~initially:
+              $resource := ($rhs $('::'.relocate(self)) Closeable)
+            block:
+              let $bind = $resource_with_statinfo
+              #void
+              $body
+              ...
+            ~finally:
+              ($resource :~ Closeable).close()
```
LiberalArtist, line 38: IIUC, with this implementation, control is allowed to escape from and reenter the `body`, and, if so, the `rhs` is evaluated again to re-allocate the resource. Is this the semantics you want? I think a more useful default might be to wrap body in a continuation barrier to forbid reentry. (This would be stricter than `call-with-{in,out}put-file*`, which let you reenter, you just find your port closed when you do.) Reentrancy can be very useful, but, if a `Closeable` implementer isn't thinking about it, unfortunate consequences seem likely: for example, truncating the output you wrote to the file before the suspend/resume.

samth:
Incidentally, here are the current path/file-related racket imports in the file I'm working on:

```
import lib("racket/base.rkt") as r:
  expose #{link-exists?} as link_exists
  expose #{directory-list} as directory_list
  expose #{rename-file-or-directory} as rename_file_or_directory
import lib("racket/file.rkt") open:
  expose #{write-to-file} as write_to_file
  expose #{make-temporary-file} as make_temporary_file
  expose #{delete-directory/files} as delete_directory_files
  expose #{make-directory*} as make_directory_alt
```

samdphillips:
> I added one more operation (`Path.only`)

It was initially unclear to me what that method is supposed to do.  Could we use a different name for that?  In the shell that would be similar to the `dirname` operation.  In Python pathlib it's close to `Path.parent`.

samdphillips:
> Should `AbolutePath` and `RelativePath` be annotations (which are easier to use in patterns)? And then drop `Path.is_absolute`?

At first I had made those as annotations.  But they didn't seem as useful since they were just predicate annotations and I could just use the predicate directly.  I could see them being useful in `match` and with `is_a` instead of calling the predicate methods.

samdphillips:
> expose #{make-directory*} as make_directory_alt

`Path.make_directory_tree` sounds better to me.

samdphillips:
> I pushed some adjustments to `Closeable`, since the interplay between Racket and Rhombus layers there is tricky.

I'm not sure if this is good or bad, but in the previous iteration this was a static error

```
class Bar()

block:
  use_static
  Closeable.def y = Bar()
  #void
```

rhombus-lib/rhombus/private/amalgam/closeable.rhm
mflatt: Allowing jumps out and in with reopening was the intent, at least. I'm not certain that's the best choice, but I'm having trouble thinking of a situation where continuation jumps might happen and this boundary is the right place to block them.

mflatt:
> I'm not sure if this is good or bad, but in the previous iteration this was a static error

Interesting, but I currently lean toward "bad" on the grounds that it's indirect (checking for `close` as opposed to `Closeable`) and more ad hoc than most rules.

samth:
I added `add_suffix`. I'm happy with either `dirname` or `basename` or `parent` or something else for `path-only`.

benknoble:
> I added `add_suffix`. I'm happy with either `dirname` or `basename` or `parent` or something else for `path-only`.

An idea based on Vim (and Zsh) path modifiers, which uses `:h` and `:t` for the equivalent of dirname and basename: `head` and `tail`. Probably too ML-ish/not clear enough without the connection to other tools, but a neat connection nonetheless.

rhombus-lib/rhombus/private/amalgam/closeable.rhm
```diff
@@ -0,0 +1,40 @@
...
+    values(
+      '
+        block:
+          def mutable $resource = #false
+          try:
+            ~initially:
+              $resource := ($rhs $('::'.relocate(self)) Closeable)
+            block:
+              let $bind = $resource_with_statinfo
+              #void
```
usaoc, line 34: 1. Should `Closeable.def` *not* add the `#void`, so that empty body results in an error?
2. I think `Closeable.let` may be a better name here to indicate the scope.

rhombus-lib/rhombus/private/amalgam/port.rkt
```diff
@@ -153,6 +173,11 @@
     [() (open-output-bytes)]
     [(name) (open-output-bytes name)]))
 
+(define/arity (Port.Output.open_file path #:exists [exists 'error])
+  #:primitive (open-output-file)
+  #:static-infos ((#%call-result #,(get-output-port-static-infos)))
+  (open-output-file path #:exists (->ExistsFlag exists)))
```
usaoc, line 179: This needs to check that the result of `->ExistsFlag` is non-`#f` and raise an annotation error otherwise (but also check `path` before raising the annotation error).

rhombus-lib/rhombus/private/amalgam/path-object.rkt
```diff
@@ -36,7 +45,8 @@
 (define-primitive-class Path path
   #:lift-declaration
   #:no-constructor-static-info
-  #:instance-static-info ((#%compare ((< path<?)
+  #:instance-static-info ((#%append Path.extend)
```
usaoc, line 48: I think `Path.extend` doesn’t behave exactly like an `append` implementation and thus `++` shouldn’t work.  In general, an `append` method receives arguments that inherit the same `append` implementation, but `Path.extend` can also receive path strings and the two special symbols `'up` and `'same`.  Maybe a separate `Path.append` that only works with path values could be defined, if that seems useful (but `Appendable` then needs to be updated to handle path values).

rhombus-lib/rhombus/private/amalgam/annotation.rkt
```diff
@@ -442,11 +442,12 @@
    (lambda (form tail)
      (syntax-parse tail
        [(op::name . (~var t (:annotation-seq #'::)))
+        (define op-name (string->symbol (shrubbery-syntax->string #'op.name)))
```
usaoc, line 445: Is the purpose of this change to use the raw text as the `who`?

rhombus-lib/rhombus/private/amalgam/path-object.rkt
benknoble: If we don't like `++`, I favor Python's trick of `dir / file` (overloading division) as a natural way to write paths.

samdphillips: Initially I wanted to use `/` or `//` as a dedicated operator (like Python.). The problem with `/` is that it's complicated and probably the wrong choice since there really isn't any existing overloading of the arithmetic operators.  `//` is right out since it is a comment token at the shrubbery level.

I agree `++` is probably not the right place to do this.  I do think it would be nice if there was an infix operator to do this.  Maybe adding `Path./` operator is a good idea?

rhombus-lib/rhombus/private/amalgam/port.rkt
samdphillips: Fixed, thanks!

rhombus-lib/rhombus/private/amalgam/annotation.rkt
mflatt: Yes, that's right.

rhombus-lib/rhombus/private/amalgam/closeable.rhm
mflatt: I agree about the name `Closeable.let`. The original motivation for adding `#void` was to allow just `Closeable.def` (or `Closeable.let`) by itself in a module body, but maybe that wasn't a useful goal.

samdphillips: Both sound good.  I've changed it.

rhombus-lib/rhombus/private/amalgam/path-object.rkt
mflatt: Would `+/` work well if it were allowed as an operator? I think the prohibition against `/` at the end of an operator could be lifted by adding the narrower constraint that an operator ending in `/` cannot be followed by `/` or `*`.

samdphillips:
> Besides`Closeable`, it looks like `to_complete_path` is not yet documented, and probably that should be `to_absolute_path`.
> 
> Should `AbolutePath` and `RelativePath` be annotations (which are easier to use in patterns)? And then drop `Path.is_absolute`?

Both of these are now implemented.

rhombus-lib/rhombus/private/amalgam/path-object.rkt
samdphillips: Removed support for `Appendable` and add the `+/` operator on `PathString`s

mflatt:
This seems ready to merge. I still have some small thoughts, but I don't want to create too much delay with more questions and further discussion (because we can always change more after), so let's time out these suggestions if they don't seem immediately like a good idea:

* Change `Path.extend` to `Path.add`? That seems easier to remember in combination with `+/`.
* Change `Path.parts` to `Path.split`? The terminology "elements" is more established than "parts" within Racket, but both "parts" and "elements" seem generic. The verb "split" is more compelling to me, makes clearer that it's a method instead of a field/property (although that's a weak benefit), and fits with a likely `String.split` to appear.

FWIW, I think some of the path tests will fail on Widows, but we can leave that to later (along with path-kind additions).

rhombus-lib/rhombus/private/amalgam/closeable.rhm
mflatt: I had always wondered why `call-with-input-file` and similar don't support continuation jumps in by reopening a file. Now that I try it, the reason seems clear: `call-with-input-file` could reopen a file, but it won't change the port that the argument to `call-with-input-file` is using. So, I concede that a barrier here is a good idea.

PR #564, opened 2024-10-23
shrubbery: fix handling of sign before `0x`/`0o`/`0b` integers

usaoc:
Allow an optional sign before `0x`/`0o`/`0b` integers, like normal decimal integers.  Previously, the specification didn't allow an optional sign before `0x`/`0o`/`0b` integers, but the presence of sign would make the integer be parsed into `#false` instead of an error.

### DIFF
--- a/shrubbery-lib/shrubbery/lex.rkt
+++ b/shrubbery-lib/shrubbery/lex.rkt
@@ -55,7 +55,7 @@
          lex-dont-stop-status?)
 
 (define-lex-abbrevs
-  
+
   ;; For case insensitivity
   [e (char-set "eE")]
 
@@ -73,7 +73,7 @@
   ;; does not constrain to avoid surrogates:
   [unicode  (:or (:: "u" (:** 1 4 digit16))
                  (:: "U" (:** 1 6 digit16)))]
-  
+
   [str (:: "\"" (:* string-element ) "\"")]
 
   [string-element (:or (:~ "\"" "\\" "\n" "\r")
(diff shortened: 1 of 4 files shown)


shrubbery-lib/shrubbery/lex.rkt
```diff
@@ -1055,29 +1055,34 @@
...
-      (case (string-ref s 1)
-        [(#\x) (string->number (regexp-replace* #rx"_" (substring s 2) "") 16)]
-        [(#\o) (string->number (regexp-replace* #rx"_" (substring s 2) "") 8)]
-        [(#\b) (string->number (regexp-replace* #rx"_" (substring s 2) "") 2)]
-        [else
-          (string->number (regexp-replace* #rx"_" s ""))])
-      (string->number (regexp-replace* #rx"_" s ""))))
+  (define-values (negative? radix num)
+    (cond
+      [(regexp-match #rx"^([+-])?(0[xob])?(.+)$" s)
```
mflatt, line 1060: I worry a little about the cost of the redundant parsing here, since an extra layer of regexp matching and allocating string matches is a lot more expensive than the old character tests. To check whether it could possibly matter, I tried the program below before an after pulling the commit. It reports 25ms before and 29ms after on my machine, which seems significant given everything else that happens for parsing.

```
#lang rhombus
import:
  rhombus/shrubbery
  rhombus/measure

def s = String.append(& for List (i: 0..10000):
                        " " +& i)
def p = Port.Input.open_string(s)

measure.time:
  ~gc
  shrubbery.read(p)
  #void
```

usaoc: Updated to address the concern, hopefully.

mflatt:
LGTM - thanks!

PR #566, opened 2024-11-02
fill out `Path` and add support for path conventions

mflatt:
This update is intended to cover all of the syntactic operations on paths. There's a new `filesystem` namespace to house filesystem-sensitive operations, mostly to innustrate the intent of putting filesystem-sensitive operations there instead of methods of `Path`.

The main addition is `CrossPath`, which corresponds to `path-for-some-system?` at the Racket level. Every `Path` is also a `CrossPath`, and `Path` methods generally accept `PathString || CrossPath`; the documentation uses `Path.like` inn result annotations to clarify that `CrossPath`s will result only when non-`Path` `CrossPath`s are given. Filesystem operations require `PathString`s (i.e., they do not accept `CrossPath`s).

I first experimented with making `Path` operations not accept `CrossPath`s and adding separate `CrossPath` methods, instead, but I didn't think worked out as well.

Constructor summary:
```
   Path(s)
   CrossPath(s, convention)
   CrossPath.Unix(s) // = CrossPath(s,  #'unix)
   CrossPath.Windows(s) // = CrossPath(s,  #'windows)
   Path.Element(s)
   CrossPath.Element(s, convention)
```

Method summary:
```
   Path.name(p)
   Path.parent(p)
   Path.bytes(p)
   Path.string(p)
   Path.convention(p)
   Path.add(p, path, ...)
   Path.split(p)
   Path.to_absolute_path(p, ~relative_to: base)
   Path.to_directory_path(p)
   Path.directory_only(p)
   Path.suffix(p)
   Path.replace_suffix(p, sfx)
   Path.add_suffix(p, sfx, sep)
   Path.cleanse(p)
   Path.normal_case(p)
   Path.simplify(p) // not filesystem-sensitive
   Path.as_relative_to(p, base)
   Path.Element.bytes(p)
   Path.Element.string(p)
```
Annotation summary:
```
   Path
   Path.Absolute
   Path.Relative
   Path.DriveRelative // mutually exclusive with Absolute & Relative
   Path.Element
   Path.Directory // a directory syntactically
   Path.Dot // #'up or #'same
   Path.like(expr) // Path or CrossPath, depending on expr
   CrossPath
   CrossPath.Absolute
   CrossPath.Relative
   CrossPath.DriveRelative
   CrossPath.Element
   CrossPath.Directory
   CrossPath.Unix
   CrossPath.Windows
   CrossPath.Convention
```

### DIFF
--- a/rhombus-lib/rhombus/private/amalgam/boolean-annotation.rkt
+++ b/rhombus-lib/rhombus/private/amalgam/boolean-annotation.rkt
@@ -17,18 +17,9 @@
 ;; ----------------------------------------
 ;; &&
 
-(define (handle-combine/c sep-op-sep form)
-  (and (pair? (cdr form))
-       (list? (cdr form))
-       (or (get-primitive-subcontract (cdr form))
-           (apply string-append
-                  (get-or-format-primitive-contract (cadr form))
-                  (for/list ([contract/rkt (in-list (cddr form))])
-                    (string-append sep-op-sep (get-or-format-primitive-contract contract/rkt)))))))
-
-(define (handle-and/c form) (handle-combine/c " && " form))
-
-(void (set-primitive-contract-combinator! 'and/c handle-and/c))
+(void (set-primitive-contract-combinator! 'and/c
+                                          (lambda (form)
+                                            (get-combined-primitive-contract " && " form))))
 (define-annotation-syntax &&
   (annotation-infix-operator
    (lambda () `((,(annot-quote \|\|) . stronger)))
@@ -102,9 +93,9 @@
 ;; ----------------------------------------
 ;; ||
 
-(define (handle-or/c form) (handle-combine/c " || " form))
-
-(void (set-primitive-contract-combinator! 'or/c handle-or/c))
+(void (set-primitive-contract-combinator! 'or/c
+                                          (lambda (form)
+                                            (get-combined-primitive-contract " || " form))))
 (define-annotation-syntax \|\|
   (annotation-infix-operator
    null
--- a/rhombus-lib/rhombus/private/amalgam/boolean-annotation.rkt
+++ b/rhombus-lib/rhombus/private/amalgam/boolean-annotation.rkt
@@ -17,14 +17,16 @@
 ;; ----------------------------------------
 ;; &&
 
-(define (handle-and/c form)
+(define (handle-combine/c sep-op-sep form)
   (and (pair? (cdr form))
        (list? (cdr form))
        (or (get-primitive-subcontract (cdr form))
            (apply string-append
-                  (get-primitive-contract (cadr form))
+                  (get-or-format-primitive-contract (cadr form))
                   (for/list ([contract/rkt (in-list (cddr form))])
-                    (string-append " && " (get-primitive-contract contract/rkt)))))))
+                    (string-append sep-op-sep (get-or-format-primitive-contract contract/rkt)))))))
+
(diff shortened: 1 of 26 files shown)


rhombus/rhombus/scribblings/reference/filesystem.scrbl
```diff
@@ -0,0 +1,74 @@
...
+
+@doc(
+  fun filesystem.expand_user_path(path :: PathString) :: Path
+){
+
+ Uses @rhombus(Path.cleanse) on @rhombus(path) and then, on platforms
+ that use the @rhombus(#'unix) path convention, checks for a leading
+ @litchar{~} and replaces the path element starting from @litchar{~} with
+ the path to the specified user's home directory. A
+ @rhombus(Exn.Fail.Failesystem, ~annot) exception is thrown if the home
```
samdphillips, line 71: Typo.

mflatt: Fixed - thanks!

rhombus-lib/rhombus/private/amalgam/boolean-annotation.rkt
```diff
@@ -17,15 +17,17 @@
 ;; ----------------------------------------
 ;; &&
 
-(define (handle-and/c form)
+(define (handle-combine/c sep-op-sep form)
   (and (pair? (cdr form))
        (list? (cdr form))
        (or (get-primitive-subcontract (cdr form))
```
usaoc, line 23: This also needs to be factored out.

rhombus-lib/rhombus/private/amalgam/path-object.rhm
```diff
@@ -33,8 +33,8 @@ namespace PathString:
   | 'to_absolute_path':
       'converting(fun (p :: PathString) :~ Path: to_absolute_path(p))'
 
-operator ((l_path :: PathString)
-            +/ (r_path :: PathString || matching(#'up) || matching(#'same)))
+operator ((l_path :: PathString || CrossPath || Path.Dot)
+           +/ (r_path :: PathString || CrossPath || Path.Dot))
```
usaoc, line 37: The indentation is off.

rhombus-lib/rhombus/private/amalgam/boolean-annotation.rkt
usaoc: Sorry, I should’ve made this clearer: `get-primitive-subcontract` needs to be factored out, because this function specifically is only intended for `and/c` contracts, for rewriting contracts like `(and/c hash? (not/c immutable?))` into `mutable-hash?` (which then gets further rewritten).  Maybe we can have something like `get-primitive-supercontract`, if that makes sense for `or/c` contracts, but I don’t see any `or/c` contract that needs that treatment yet.

mflatt: Ah, I had not noticed or understood the "subcontract" part. Meanwhile, I made `get-primitive-contract` check for a complete match before trying the combinator table. Does it make sense to use the contract table that way, instead of having a subcontract table? If necessary in the future, a combination-handling function can try building and looking up combinations of subsets of components to handle something like `(and/c X? hash? (not/c immutable?))`.

mflatt: I added a commit to show what I mean.

PR #568, opened 2024-11-06
Add `filesystem.list_directory`

samdphillips:


rhombus/rhombus/scribblings/reference/filesystem.scrbl
```diff
@@ -72,3 +72,15 @@
  directory cannot be found.
 
 }
+
+@doc(
+  fun filesystem.list_directory(path :: PathString = Path.current_directory(),
+                                ~build_path: build = #false) :: List.of(Path)
```
sorawee, line 78: Would the keyword name `absolute` be better? The return signature is `List.of(Path)`, so it's building `Path`s whether `build_path` is true or false...

samdphillips: It doesn't build an absolute path though.  It extends the `path` with the entry from the directory.

samdphillips: `extend_path` may be better since the library doesn't use `extend` instead of `build`.

sorawee: I agree that `extend_path` sounds better.

The reason I proposed `absolute` is because the text says:

```
  If @rhombus(build) is @rhombus(#false) the resulting
  paths are all @rhombus(Path.Element)s; otherwise, the results are combined
  with @rhombus(path) to create @rhombus(Path.Absolute)s.
```

Is that not accurate?

samdphillips: You're right I miswrote that.  Will fix.

rhombus/rhombus/scribblings/reference/filesystem.scrbl
```diff
@@ -72,3 +72,15 @@
  directory cannot be found.
 
 }
+
+@doc(
+  fun filesystem.list_directory(path :: PathString
+                                  = Path.current_directory(),
+                                ~extend_path: extend = #false) :: List.of(Path)
```
usaoc, line 79: Nit: Can you move the result annotation to a new line?

rhombus/rhombus/scribblings/reference/filesystem.scrbl
```diff
@@ -72,3 +72,15 @@
...
+){
+
+  Returns a list of all files and directories in the directory specified
+  by @rhombus(path).  If @rhombus(extend) is @rhombus(#true) the resulting
+  paths are extended from @rhombus(path).
+
+}
```
usaoc, line 86: I think the diff shows a missing new line here?

rhombus/rhombus/scribblings/reference/filesystem.scrbl
samdphillips: No problem

usaoc:
Thanks!

PR #570, opened 2024-11-09
fill in `filesystem` API

mflatt:
This is a work in progress, especially because it's not yet tested, but the `filesystem` API here is ready for feedback.

The goal is to cover the best and most useful parts of `racket/base` and `racket/file`. In some cases, `racket/file` provides a more general version of a simple function provided from `racket/base`; only the general version is included in this `filesystem`, but with options that make the simple behavior accessible (if not always the default). The implementation uses only `racket/base`.

Many small changes compared to `racket/base` and `racket/file` are intended to make things clearer, more consistent, and read more nicely. But I'm unsure about some of the keyword choices, especially `~as`.

```
fun filesystem.type(path :: PathString,
                    ~must_exist: must_exist = #false)
  :: maybe(matching(#'file || #'directory || #'link || #'directory_link))

fun filesystem.file_exists(path :: PathString) :: Boolean
fun filesystem.directory_exists(path :: PathString) :: Boolean
fun filesystem.link_exists(path :: PathString) :: Boolean

// extends and replaces `list_directory`
fun filesystem.files(
  path :: PathString = Path.current_directory(),
  ~add_path: add_path = #false,
  ~recur: recur = #false,
  ~follow_links: follow_links = #false,
  ~skip: skip :: Path -> Any = fun (x): #false
) :: List.of(Path)

fun filesystem.roots() :: List.of(Path.Absolute)

fun filesystem.rename(path :: PathString,
                      to_path :: PathString,
                      ~exists_ok: exists_ok = #false)
  :: Void

fun filesystem.delete(
  path :: PathString,
  ~as: mode :: matching(#'any || #'file || #'directory) = #'any,
  ~recur: recur = #false,
  ~must_exist: must_exist = #false
) :: Void

fun filesystem.make_directory(
  path :: PathString,
  ~parents: parents = #false,
  ~permissions: permissions :: Int.in(0, 65535) = 0o777
) :: Void

fun filesystem.make_link(to_path :: PathString,
                         path :: PathString)
  :: Void

fun filesystem.make_temporary(
  ~in: dir :: PathString = system.path(#'temp_dir),
  ~as: as:: (PathString || matching(#'file) || matching(#'directory))
       = #'file,
  ~make_name: make_name :: String -> Path.Relative
                = fun (s): Path("tmp" +& s),
  ~permissions: permissions :: maybe(Int.in(0, 65535))
                  = #false,
  ~replace_permissions: replace_permissions = #false

) :: Path

fun filesystem.copy(
  src_path :: PathString,
  dest_path :: PathString,
  ~recur: recur = #false,
  ~exists_ok: exists_ok = #false,
  ~follow_links: follow_links = #true,
  ~permissions: permissions :: maybe(Int.in(0, 65535)) = #false,
  ~replace_permissions: replace_permissions = #true,
  ~keep_modify_seconds: keep_modify_seconds = #false
) :: Void

fun filesystem.size(path :: PathString) :: NonnegInt

fun filesystem.modify_seconds(
  path :: PathString,
  ~change_to: change_to :: maybe(Int) = #false
) :: Int || Void

fun filesystem.modify_seconds(
  path :: PathString,
  ~change_to: change_to :: maybe(Int) = #false,
  ~fail: fail_k :: Function.of_arity(0)
) :: ~any

fun filesystem.permissions(
  path :: PathString,
  ~bits: bits = #false,
  ~change_to: change_to :: maybe(Int.in(0, 65535)) = #false
) :: Int || List.of(matching(#'read || #'write || #'exceute)) || Void

fun filesystem.identity(
  path :: PathString,
  ~follow_links: follow_links = #true
) :: PosInt

fun filesystem.stat(path :: PathString,
                    ~follow_links: follow_links = #true)
  :: Map

fun filesystem.bytes(
  path :: PathString,
  ~mode: mode :: Port.Mode = #'binary
) :: Bytes

fun filesystem.string(
  path :: PathString,
  ~mode: mode :: Port.Mode = #'binary
) :: String

fun filesystem.bytes_lines(
  path :: PathString,
  ~mode: mode :: Port.ReadLineMode = #'any
) :: List.of(String)

fun filesystem.lines(
  path :: PathString,
  ~mode: mode :: Port.ReadLineMode = #'any
) :: List.of(String)

fun filesystem.write_string(
  path :: PathString,
  str :: String,
  ~mode: mode :: Port.Mode = #'binary
) :: Void

fun filesystem.write_bytes(
  path :: PathString,
  bstr :: Bytes,
  ~mode: mode :: Port.Mode = #'binary
) :: Void

fun filesystem.simplify_path(path :: PathString) :: Path
fun filesystem.normalize_path(path :: PathString) :: Path
fun filesystem.resolve_path(path :: PathString) :: Path
fun filesystem.expand_user_path(path :: PathString) :: Path

Parameter.def filesystem.current_force_delete_permissions
  :: Any.to_boolean
```

### DIFF
--- a/rhombus-lib/rhombus/private/amalgam/filesystem.rhm
+++ b/rhombus-lib/rhombus/private/amalgam/filesystem.rhm
@@ -4,6 +4,13 @@ import:
   lib("racket/base.rkt") as rkt
   lib("racket/path.rkt") as rkt_path
   "error-adjust.rkt" as rkt_error
+  "error.rhm" open
+  "maybe.rhm" open
+  "closeable.rhm" open
+  "path-object.rhm" open
+  "rename-parameter.rkt".#{rename-parameter}
+
+use_static
 
 export:
   filesystem
@@ -17,7 +24,28 @@ namespace filesystem:
     file_exists
     directory_exists
     link_exists
-    list_directory
+    type
+    files
+    roots
+    names:
+      rename
+    delete
+    make_directory
+    current_force_delete_permissions
+    make_temporary
+    make_link
+    copy
+    size
+    modify_seconds
+    permissions
+    identity
+    stat
+    read_bytes
+    read_string
+    read_lines
+    read_bytes_lines
+    write_bytes
+    write_string
 
   fun simplify_path(p :: PathString) :: Path:
     rkt.#{simplify-path}(p, #true)
@@ -25,12 +53,8 @@ namespace filesystem:
   fun normalize_path(p :: PathString) :: Path:
     // potential error from `simplify-path` as called by `simple-form-path`
     ~who: who
-    rkt_error.#{call-with-local-error-adjust}(
-      #'#{simple-form-path},
-      who,
-      fun ():
-        rkt_path.#{simple-form-path}(p)
-    )
+    rkt_error.local_error_adjust { #'#{simple-form-path}: who }:
+      rkt_path.#{simple-form-path}(p)
 
   fun resolve_path(p :: PathString) :: Path:
     rkt.#{resolve-path}(p)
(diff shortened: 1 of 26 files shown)


rhombus/rhombus/scribblings/reference/port.scrbl
```diff
@@ -24,8 +24,22 @@ output; it is possible for an object to be both an input and output port.
 
 @doc(
  def Port.eof :: Port.EOF
+ bind.macro 'Port.eof'
 ){
 
- A value (distinct from all other values) that represents an end-of-file.
+ The @rhombus(Port.eof) value represents an end-of-file (distinct from
+ all other values), and the @rhombus(Prof.eof) binding match matches that
```
samdphillips, line 31: `Prof.eof` -> `Port.eof`

rhombus-lib/rhombus/private/amalgam/filesystem.rhm
```diff
@@ -47,5 +71,344 @@ namespace filesystem:
...
+    fun files_here(accum :~ List, p, onto_path):
+      let files :~ PairList:
+        rkt_error.local_error_adjust { #'#{directory-list}: who }:
+          rkt.#{directory-list}(p, #{#:build?}: !onto_path && add_path)
+      for values (accum :~ List = accum) (file: files):
+        let file = (if onto_path | Path.add(onto_path, file) | file)
+        skip_when skip(file)
+        let new_accum = accum.add(file)
+        if recur && (if follow_links
+                     | directory_exists(file)
```
samdphillips, line 97: If `start_p` is an absolute path more than one directory deep, and `add_path` is `#false` then `recur` has no effect.  The `file` that `directory_exists` checks will be absolute, but not attached to the current directory.

Similarly if called `files("a/b", ~recur: #true)` and there is a directory under the current directory called `b` the wrong `b` will be traversed.

rhombus-lib/rhombus/private/amalgam/filesystem.rhm
```diff
@@ -47,5 +71,344 @@ namespace filesystem:
...
+      | (recur && mode != #'file) || mode == #'any:
+          match (rkt_error.local_error_adjust { #'#{file-or-directory-type}: who }:
+                   rkt.#{file-or-directory-type}(p))
+          | #'file || #'link:
+              delete_file()
+          | #'directory:
+              when recur
+              | let files :~ PairList:
+                  rkt_error.local_error_adjust { #'#{directory-list}: who }:
+                    List(& rkt.#{directory-list}(p, ~build: #true))
```
samdphillips, line 148: Should be the `build?` keyword

rhombus-lib/rhombus/private/amalgam/filesystem.rhm
```diff
@@ -47,5 +71,344 @@ namespace filesystem:
...
+              unless moved
+              | rkt.#{delete-file}(p)
+              rkt.#{delete-file}(tmp)
+        | ~else:
+            rkt_error.local_error_adjust { #'#{delete-file}: who }:
+              rkt.#{delete-file}(p)
+      cond
+      | (recur && mode != #'file) || mode == #'any:
+          match (rkt_error.local_error_adjust { #'#{file-or-directory-type}: who }:
+                   rkt.#{file-or-directory-type}(p))
```
samdphillips, line 141: Can this be simplified to use `filesystem.type`?

rhombus-lib/rhombus/private/amalgam/filesystem.rhm
```diff
@@ -47,5 +71,344 @@ namespace filesystem:
...
+          | #'file || #'link:
+              delete_file()
+          | #'directory:
+              when recur
+              | let files :~ PairList:
```
samdphillips, line 146: `PairList` -> `List`

rhombus-lib/rhombus/private/amalgam/filesystem.rhm
```diff
@@ -47,5 +71,344 @@ namespace filesystem:
...
+                  rkt_error.local_error_adjust { #'#{directory-list}: who }:
+                    List(& rkt.#{directory-list}(p, ~build: #true))
+                for (f: files):
+                  recur(f)
```
samdphillips, line 150: `recur` is a boolean here.

rhombus-lib/rhombus/private/amalgam/filesystem.rhm
mflatt: Thanks! I haven't written any tests, yet, so I'm pretty sure it mostly doesn't work. 😄

rhombus-lib/rhombus/private/amalgam/filesystem.rhm
mflatt: The idea was to have any error reported as `delete`, but maybe this is a case where no error is possible.

rhombus-lib/rhombus/private/amalgam/filesystem.rhm
```diff
@@ -47,5 +71,344 @@ namespace filesystem:
...
+    ~who: who
+    rkt_error.local_error_adjust { #'#{make-file-or-directory-link}: who }:
+      rkt.#{make-file-or-directory-link}(to_path,
+                                         path)
+
+  fun copy(src_path :: PathString,
+           dest_path :: PathString,
+           ~recur = #false,
+           ~exists_ok = #false,
+           ~permissions: permissions :: maybe(Int.in(0, 65535)) = #false,
```
samdphillips, line 246: Having a permission option that would duplicate the permissions from the source files and directories would be nice.

gus-massa:
Why does `filesystem.modify_seconds` have a `~fail` argument. It looks very odd, because none of the other functions have a similar argument. (It it only for compatibility with Racket?)

rhombus-lib/rhombus/private/amalgam/filesystem.rhm
```diff
@@ -47,5 +71,344 @@ namespace filesystem:
...
+           ~keep_modify_seconds = #false,
+           ~follow_links = #true):
+    ~who: who
+    fun copy_file(src_path, dest_path):
+      rkt_error.local_error_adjust { #'#{copy-file}: who }:
+        rkt.#{copy-file}(src_path,
+                         dest_path,
+                         #{#:exists-ok?}: exists_ok,
+                         ~permissions: permissions,
+                         #{#:replace-permissions}: replace_permissions)
```
samdphillips, line 257: `#:replace-permissions` -> `#:replace-permissions?`

rhombus-lib/rhombus/private/amalgam/filesystem.rhm
```diff
@@ -47,5 +71,344 @@ namespace filesystem:
...
+    rkt_error.local_error_adjust { #'#{file-or-directory-stat}: who }:
+      rkt.#{file-or-directory-stat}(path, !follow_links)
+
+  fun content(who, path, mode, get, append):
+    let size:
+      try:
+        rkt.#{file-size}(path)
+        ~catch _: 0
+    Closeable.let in = (rkt_error.local_error_adjust { #'#{open-input-file}: who }:
+                          rkt.#{open-input-file}(path, ~mode: mode))
```
samdphillips, line 350: Can this use `Port.Input.open_file`?

rhombus/rhombus/scribblings/reference/filesystem.scrbl
```diff
@@ -20,6 +36,373 @@
...
+ is not provided, then failure throws an exception.
+
+}
+
+@doc(
+  fun filesystem.permissions(
+    path :: PathString,
+    ~bits: bits = #false,
+    ~change_to: change_to :: maybe(Int.in(0, 65535)) = #false
+  ) :: Int || List.of(matching(#'read || #'write || #'exceute)) || Void
```
samdphillips, line 303: `#'exceute` -> `#'execute`

rhombus-lib/rhombus/private/amalgam/filesystem.rhm
samdphillips: Oh, that makes more sense.  This `local_error_adjust` pattern with `file-or-directory-type` happens a few times in the module, maybe it could be a private function that takes a `who` parameter?

samdphillips:
I have mixed feelings for the names `filesystem.bytes`, `filesystem.string`, etc.  They're short, simple, and easy (for me) to figure out and remember.  But they seem a little weird.

rhombus-lib/rhombus/private/amalgam/filesystem.rhm
samdphillips: It mostly works :D.

rhombus-lib/rhombus/private/amalgam/filesystem.rhm
```diff
@@ -47,5 +71,344 @@ namespace filesystem:
...
+  fun permissions(path :: PathString,
+                  ~bits = #false,
+                  ~change_to: change_to :: maybe(Int.in(0, 65535)) = #false):
+    ~who: who
+    rkt_error.local_error_adjust { #'#{file-or-directory-permissions}: who }:
+      cond
+      | change_to:
+          rkt.#{file-or-directory-permissions}(path, change_to)
+      | bits:
+          List(& rkt.#{file-or-directory-permissions}(path, #'bits_to))
```
samdphillips, line 328: `#'bits_to` -> `#'bits`

rhombus-lib/rhombus/private/amalgam/filesystem.rhm
```diff
@@ -47,5 +71,344 @@ namespace filesystem:
...
+  fun identity(path :: PathString,
+               ~follow_links = #true) :: PosInt:
+    ~who: who
+    rkt_error.local_error_adjust { #'#{file-or-directory-identity}: who }:
+      rkt.#{file-or-directory-identity}(path, #{#:as-link?}: !follow_links)
+
+  fun stat(path :: PathString,
+           ~follow_links = #true) :~ Map:
+    ~who: who
+    rkt_error.local_error_adjust { #'#{file-or-directory-stat}: who }:
```
samdphillips, line 341: The keys in the returned hash should probably not use hyphens.

mflatt:
> Why does `filesystem.modify_seconds` have a `~fail` argument. It looks very odd, because none of the other functions have a similar argument. (It it only for compatibility with Racket?)

I agree that this is awkward, but it does reflect what is available from Racket. I'll think more about this and maybe revisit the Racket interface.

rhombus-lib/rhombus/private/amalgam/filesystem.rhm
mflatt: It's currently documented as "low level, so using Racket symbol conventions". My thought here was that creating a new map to rename keys is a lot of work with little payoff, and using the function is already a much leakier abstraction than the others.

rhombus/rhombus/scribblings/reference/port.scrbl
mflatt: Hah! It's retention-and-tenure-review season, so maybe my fingers have become mistrained.

rhombus-lib/rhombus/private/amalgam/filesystem.rhm
mflatt: Agreed.

mflatt:
Other changes I'm trying:
* Use `~set_to` instead of `~change_to` for `filesystem.modify_seconds` and `filesystem.permissions`
* Make `filesystem.permissions` return a set instead of a list
* Adding `~exists` to `filesystem.write_bytes` and `filesystem.write_string`

mflatt:
New version pushed, now with tests.

samdphillips:
> > Why does `filesystem.modify_seconds` have a `~fail` argument. It looks very odd, because none of the other functions have a similar argument. (It it only for compatibility with Racket?)
> 
> I agree that this is awkward, but it does reflect what is available from Racket. I'll think more about this and maybe revisit the Racket interface.

Looking at the API (and not knowing the historical reason why) it seems that it is designed to do a cheap retry if setting the time on a file fails?

samdphillips:
> > I have mixed feelings for the names `filesystem.bytes`, `filesystem.string`, etc. They're short, simple, and easy (for me) to figure out and remember. But they seem a little weird.
> 
> Yes, I think I want too far with short names there. How about `filesystem.read_bytes`, `filesystem.read_string`, `filesystem.read_bytes_lines`, and `filesystem.read_lines`?

Yes, those seem better.

samdphillips:
I would like to write something like the following without having to filter out the directories.

```
fun scribl_only(path :~ Path):
  match filesystem.type(path)
  | #'directory || #'directory_link:
      let n = path.name()
      n == Path(".git") || n == Path("compiled")
  | _:
      path.suffix() != #".scrbl"

filesystem.files("/home/sam/projects/contrib/rhombus",
                 ~skip: scribl_only,
                 ~add_path: #true,
                 ~recur: #true)
```

I'm not sure how that can be achieved though without making the API for `filesystem.files` more complicated in some way.

mflatt:
> I would like to write something like the following without having to filter out the directories.

I wonder whether that might be better handled by a separate set of combinators for building filter functions, and then you could more conveniently generate a function to pass as `~skip`.

One way that `filesystem.files` could help is by having both `~skip` and `~keep`, though, so that it's easy to use either disposition of filter, and you don't have to remember which one to use.

That strategy could also resolve the problem with `List.filter` of remembering how it treats the results of the function. It would accept up to two functions with `~keep` and `~skip`.

samth:
This nicely removed many of my imports from `racket/file`, `racket/base`, and `racket/port`. Still missing:

- `with-{input-from,output-to}-{port,file,bytes}`
- `file-position`
- `write-{bytes,string}`

Also, it's a bigger discussion but it's not obvious what the right way to replace `read`/`write` serialization is in Rhombus.

PR #579, opened 2024-11-18
support `~nonempty` after `...` to match 1 or more in a repetition

mflatt:
Example:

```
> :
    match [1, 3]
    | [1, x, ..., 3]: [x, ...]
[]
> :
    match [1, 3]
    | [1, x, ... ~nonempty, 3]: [x, ...]
match: no matching case
> :
    match [1, 2, 3]
    | [1, x, ... ~nonempty, 3]: [x, ...]
[2]
```

### DIFF
--- a/rhombus/rhombus/tests/array.rhm
+++ b/rhombus/rhombus/tests/array.rhm
@@ -103,6 +103,10 @@ block:
     Array(v, ...).length()
     ~is 3
   check:
+    def [v, ... ~nonempty] = dynamic([1, 2, 3])
+    Array(v, ...).length()
+    ~is 3
+  check:
     class Posn(x, y)
     def arr :: Array.later_of(Posn) = Array(Posn(1, 2))
     arr[0].x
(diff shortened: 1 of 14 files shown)


mflatt:
Now with `... ~once`, which is useful as a form of "optional"  that's lightweight (in the sense that it's a conceptually and notationally small step from `...`).

rhombus/rhombus/tests/array.rhm
```diff
@@ -183,6 +187,15 @@ check:
...
   ~completes
 
+check:
+  def Array(0, _, ...) = Array(0)
+  ~completes
+
+check:
+  def Array(0, _, ... ~nonempty) = Array(0)
+  "ok"
+  ~raises error.annot_msg()
```
usaoc, line 197: `~raises` → `~throws`, here and elsewhere

mflatt: Ah — missed this before I merged. I'll push a repair.

PR #580, opened 2024-11-20
Add `~mode:` flag to `str.s`

samdphillips:


usaoc:
Can you also add some tests to `rhombus/tests/str.rhm` that use that new `~mode` flag?  Thanks!

mflatt:
Thanks!

PR #587, opened 2024-11-27
export namespace extension static-info

AlexKnauth:
If I have this in a `def.rhm`:
```
#lang rhombus/static
export: MyNamespace
namespace MyNamespace
def MyNamespace.l :: List: []
```
And this in a `use.rhm`:
```
#lang rhombus/static
import: "def.rhm".MyNamespace
for (e: MyNamespace.l):
  println(e)
```
Before this PR, it gives me an error about about not enough static information. After this PR, the error goes away.

- [x] Tests


mflatt:
Thank you for investigating enough to suggest a solution! I don't think this is the right direction, though, since it defeats the intended check on whether a definition is meant to extend a namespace. I'll add the right plumbing for `define-static-info-syntax`.

PR #589, opened 2024-11-30
Repair `List` reducer not producing `List.empty`

samdphillips:


samdphillips:
I just noticed `for/treelist` uses `(fx= 0 size)` as the test, so maybe this should do the same?

mflatt:
In this context, probably `(eqv? 0 size)` is simplest and best. I'll make that adjustment and merge. Thanks for the repair!

PR #591, opened 2024-11-30
add `definitely` epression and binding form

mflatt:
The `definitely` form is a destructor to go with the `maybe` constructor. As an expression, `definitely(expr)` produces the same result as `expr`, except that it errors if that result is `#false`. And if `expr` has static information from `maybe(annot)`, then `definitely(expr)` has the static information of `annot`. The `definitely(bind)` binding form matches only non-`#false` values that match `bind`. Using `definitely` is the same as writing `expr :: annot` or `bind :: annot`, except that you don't have to write out `annot`.

The commit also adjusts `match` to propagate static information from the target expression to binding patterns, which it should have done anyway, but is especially needed to make `definitely` work as intended.

### DIFF
--- a/rhombus-lib/rhombus/private/amalgam/annot-macro.rkt
+++ b/rhombus-lib/rhombus/private/amalgam/annot-macro.rkt
@@ -177,11 +177,11 @@
     (check-syntax who static-infos)
     (syntax-parse binding
       #:datum-literals (parsed)
-      [(parsed #:rhombus/bind _::binding-form)
+      [(parsed #:rhombus/bind b::binding-form)
        (no-srcloc
         #`(parsed #:rhombus/annot
                   #,(annotation-binding-form
-                     binding
+                     #'b
                      (wrap-expression body)
                      (pack-static-infos who (unpack-term static-infos who #f)))))]
       [_ (raise-arguments-error* who rhombus-realm
--- a/rhombus-lib/rhombus/private/amalgam/arithmetic.rkt
+++ b/rhombus-lib/rhombus/private/amalgam/arithmetic.rkt
@@ -28,7 +28,7 @@
                      mod
                      rem
 
-                     !
+                     ! ;; replaced by `!?` in "maybe.rhm"
                      &&
                      \|\|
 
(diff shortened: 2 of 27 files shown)


sorawee:
In TypeScript this is the `!` operator. Does it make sense to adopt the notation as well? `definitely` is kinda long and not ergonomic to type.

usaoc:
This reminded me that I wanted to add a few pipeline operators, among which was the `&>` (“and pipe-forward”) operator that would work with `maybe` values (it would pipe when the left-hand side is true).  I didn’t have the time to look into how to make static info cooperates with it, but I think part of this PR could make it work.

> In TypeScript this is the ! operator.

I wonder if we could make the postfix `?` operator work.

mflatt:
Postfix `!` and `?` seem good to me. It's a little awkward that `!` in prefix mode is unrelated to `!` in suffix mode, and both have to be the same binding (because Rhombus doesn't support binding infix and prefix separately), but it can work.

I'm currently leaving `definitely` in place, and postfix `!` is an alternative. I think that will make it easier to explain and organize the documentation, but we'll see.

mflatt:
Maybe it's better to use `!!` as in Kotlin instead of overloading `!` in an unusual way. Also, avoiding overloading makes it feel less useful to keep `definitely`.

mflatt:
The latest commit uses postfix `!!` instead of `definitely`.

It also has a `?` that manages to work with any operator, such as the combination `?.` or `? |>`, as in these examples (where `Syntax.srcloc`'s result is `maybe(Srcloc)`):

```
'stx'.srcloc()?.line
'stx'.srcloc() ? |> Srcloc.line
```

This `?` is a kind of operator modifier that exploits the flexibility of infix-operator macros, but I think it's too weird. It seems better to have specific operators for `.` and `|>` variants. Since `?.` is not a valid shrubbery identifier, would `.?` be a good choice? And `|>?` to go with that?

usaoc:
It seems very unfortunate that an operator cannot end with `.` just to support the less used decimal notation.  Does it make sense that we relax the restriction (for example, allow an operator to end with `.`, maybe `+` and `-` in addition, unless it’s followed by a non-negative integer)?  (I would’ve liked to use `<..` as one of the range operators, as well.)

I would also prefer `?>` over `|>?`, because it makes it clearer that the `?` part is about the left-hand side, not the right-hand side.  Same for `?.` over `.?`.

mflatt:
Continue the discussion on relaxed operator syntax in #592.

mflatt:
Closing in favor of #593

PR #592, opened 2024-12-02
allow `+`, `-`, and `.` at the end of a multi-character operator

mflatt:
Remove the constraint against `+`, `-`, and `.` at the end of a multi-character operator. This means that `?.`, `=.`, and `...+` all are operators, for example.

It also means that `1+-2` is now `1 +- 2`, not `1 + -2`. Similarly, `pi*-1/2` is `pi *- 1/2`, not `pi * -1/2`. Expressions like `pi*-1/2` appeared in some existing Rhombus code, but only because that was written before `-1/2*pi` was allowed.

It would make sense to treat `-` specially when followed by a digit so that `1+-2` would be `1 + -2`, but it seems simpler and workable to not do that.


mflatt:
The commit also shows that `.+` in the regexp language is no longer `. +`, but adding a `.+` shorthand is consistent with the existing `.*` and `.?` shorthands.

Besides `pi*-1/2`, the other code I found in existing code is also in this commit, changing `frac*-100` to `frace*(-100)`, and that change seems good to me.

PR #595, opened 2024-12-04
Repair `RX.match_range` with missing captures

samdphillips:
If a capture is missing when performing one of the range matches the result will be `#false`.  This repairs that.


usaoc:
Thanks!  Just a stylistic suggestion that you may ignore: I would’ve written a `maybe_range` function and use it in the repetition, so it would look like

```
fun maybe_range(maybe_pr):
  MATCH maybe_pr
  | Pair(s, e): s .. e
  | ~else: #false
```

and to use it

```
MATCH ....
| PairList[Pair(s0, e0), maybe_pr, ...]:
    RXMatch(s0 .. e0, [maybe_range(maybe_pr), ...], _vars)
| ~else: #false
```

samdphillips:
That's a good suggestion.  I'll make the change.

PR #597, opened 2024-12-05
change `<..<` range operator to `<..`

usaoc:
This is the form that I originally wanted to use, but couldn’t use until c3ca21e.


mflatt:
LGTM

PR #599, opened 2024-12-07
fill out `Port` API

mflatt:
Add everything from the Racket reference sections 13.1.1 through 13.1.17, except for `byte-ready?` and `char-ready?`.

Add `stdin` as a shorthand for `Port.Input.current()`, and similarly add `stdout` and `stderr`.

Add `Port.Input.using` and `Port.Output.using` forms as a shorthand for `Closeable.let` plus `parameterize` plus sometimes `open_file`:

```
Port.Input.using ~file "/tmp/data.txt":
  stdin.read_line()

Port.Input.using Port.Input.open_string("data"):
  stdin.read_line()

Port.Output.using ~file "/tmp/data.txt":
  ~exists: #'replace
  println("data")
```

Also generalize dot-provider handling so that an annotation like `Port.Input && Port.FileStream` can work and provide static access to all the methods of both `Port.Input` and `Port.FileStream`.

### DIFF
--- a/rhombus-lib/rhombus/private/amalgam/builtin-dot.rkt
+++ b/rhombus-lib/rhombus/private/amalgam/builtin-dot.rkt
@@ -2,6 +2,7 @@
 (require "mutability.rkt"
          "treelist.rkt"
          "mutable-treelist.rkt"
+         "port-pipe.rkt"
          (submod "dot.rkt" for-builtin)
          (submod "map.rkt" for-builtin)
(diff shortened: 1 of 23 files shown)


samth:
The `using` approach works nicely for almost all the cases I've encountered except `with-output-to-{string,bytes}`. Maybe allowing a `~string` or `~bytes` keyword by itself would work, although that changes the result value in a way that may be surprising.

PR #610, opened 2025-01-07
add operator orders for relative precedence

mflatt:
Previously, the only way to specify precedence between two operators is for one of the operators to know specifically about the other one. For example, `+` needed to mention `*` specifically, or vice versa.

This commit adds support for *operator orders* as a alternative. An operator order is a category like `addition` or `comparison` that has a precedence relation to other operator orders. For example, the `+`, `-`, `flonum.(+)`, `flonum.(-)`, `fixnum.(+)`, `fixnum.(-)`, `fixnum.wraparound.(+)`, and `fixnum.wraparound(-)` operators all declare the `addition` order, and the `*`, `flonum.(*)`, etc., operators have the `multiplication` order; all those operators then relate to each other as expected, because the `addition` declares itself weaker than `multiplcation`.

An operator declares itself to have up to one operator order. Any precedence relation defined relative to that order then applies releative to the operator, too. The operator inherits precedence declarations (with respect to other orders) from the order's definition, but it can optionally override those. An operator order is also defined with an associativity that serves as a default for operations that have the order. Precedence relationships declared in an operator can metion either operator orders or specific operators.

An operator order is not associated with an operator binding space, such as the expression space or annotation, so multiple spaces can all used the same operator orders. This flexibility is ensured in part by having operator orders refer only to other operator orders, never to specific operators. (That is, operators can refer to operator orders, but not the other way around.)

Predefined orders in this commit:

```
   addition
   multiplication
   exponentiation
   integer_division
   arithmetic ; shorthand for the above four
   comparison
   assignment
   enumeration
   concatenation
   logical_negation
   logical_conjunction
   logical_disjunction
   bitwise_negation
   bitwise_shift
   bitwise_conjunction
   bitwise_disjunction
   bitwise_test
```

The `arithmetic` operator order is more precisely an operator order set. It can only be used for writing precedence relationships, while an operator can only declare itself to have one of the specific operator orders.

When an operator's precedence is `~weaker` or `~stronger` with `~other`, then there's usually no need for an operator order, because there's nothing left to say about precedence. That's why there's no predefined operator order for `.`, for example, at least not so far. An operator set with `~weaker` or `~stronger` as `~other` can be useful if some precedence declaration needs to make an exception, however, such as `with` with respect to `assignment`.

### DIFF
--- a/enforest-lib/enforest/operator.rkt
+++ b/enforest-lib/enforest/operator.rkt
@@ -17,7 +17,13 @@
 ;; precedence than the one referenced by the identifier. An operator
 ;; is implicitly the 'same as itself (i.e., not covered by 'default).
 
-(provide operator
+(provide order
+         order?
+         order-precedences
+         order-assoc
+         order-ref
+
+         operator
          operator?
          operator-precedences
          operator-protocol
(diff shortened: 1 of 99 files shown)


mflatt:
I remember now why `.` needs an operator order: so operators like `!!.` and `?.` can be declared in the same order. So, I'll add it.

PR #613, opened 2025-01-15
change `[]` on a `map` to return a `maybe` of a value

mflatt:
Until very recently, it seemed clear to me that `map[key]` should throw an exception if `key` is not mapped to a value within `map`. When writing Rhombus programs, though, I have wanted `#false` for unmapped keys about half of the time that I write `map[key]`. I can always write `map.get(key, #false)`, but that departs from the `[]` syntax that's intended for map lookups (to avoid having everything look like a function/method call).

After trying out different directions and defaults (listed below), it seems like my Rhombus programs are nicest when `map[key]` returns `#false` for an unmapped key, after all.

After the commit here:

 * `map[key]` returns `#false` for unmapped keys. If `map` has annotation `Map(Key, Val)`, then `map[key]` has annotation `maybe(Val)`.

 * `map[key]!!` is a way of saying that `key` needs to be mapped (but also that it needs to be mapped to a non-`#false` value). This is using the existing `!!` operation, so `map[key]!!` has annotation `Val` when `map` has `Map(Key, Val)`.

 * `map.get(key)` throws an exception if `key` is not mapped (and this is not a change). Nothing about the annotation of `map` is propagated in `map.get(val)`, however, since method result annotations are not parameteric (at least not currently).

 * `map.get(key, default_val)` returns `default_val` when `key` is not mapped (and this is not a change).

A post-hoc rationalization for why this works out:

 * The new `map[key]` is exactly what I want about half the time.

 * Most of the time that I want an error, `map[key]!!` catches errors early, it's easy to write, and it's reasonably clear to read.

 * Much of the time that I write `map[key]` and want an error for an unmapped key, I'm in a static-mode context where I get a type-like reminder to use `!!`.

 * I can still write `map.get(key)` to opt into checking, but the cases where that's needed are rarer than places where I want the new `map[key]` behavior.

Some other things I tried:

 * Using a veneer to change the static resolution of `map[key]`. This worked ok for bindings that I control completely, but it was awkward to combine with `cmdline.state`.

 * Adding a new operator like `?` to change from an exception-throwing `map[key]` to a `#false`-returning `map?[key]`. The drawback here is using up another operator and introducing a new concept, while the other direction of writing `map[key]!!` uses an existing `!!` operator.

 * Adding a property like `maybe` so that `map.maybe[key]` corresponds to `map.get(key, #false)`. This sort of works, but static-information propagation is fragile without method result annotations that are parameteric in argument annotations, and I'm doubtful of that direction for the annotation layer.

It's certainly possible, of course, that there's a better solution that I missed!

A reminder of what some other languages do:

 * JavaScript: Returns `undefined` for `map[key]` when `key` is not mapped. This is particularly a problem in JavaScript, because `map.field` corresponds to `map[key]` (where `key` is the string form of `field`). Of course, `map.field` in Rhombus remains an error if `field` does not exist. JavaScript also has a `Map` class; if `map` is a `Map` then `map.get(key)` produces `undefined` when `key` is not in `map`.

 * Lua: Like JavaScript.

 * Python: Errors for `map[key]` when `key` is unmapped. The `get` method for maps, meanwhile, returns `None` for unmapped keys by default and accepts an optional value to return insteda of `None`.

 * Groovy: Returns false for `map[key]` when `key` is unmapped. That convention is meant to be combined with `?:` (which is `||` in Rhombus) and `?.` (which Rhombus inherits as `?.` by way of Kotlin and Rust).

 * Kotlin: Returns `null` for `map[key]` when `key` is unmapped, reflected in the type system by a nullable result type. In addition to `?:` and `?.`, Kotlin has `!!` (which Rhombus inherits) to work with nullable types. A `map.get(key)` method call is like `map[key]`, while `map.get_value(key)` throws an exception if `key` is not mapped.

 * Java: Returns `null` for `map.get(key)` when `key` is not mapped, and does not support `[]` notation for maps. There seems to be no method that reports an error, instead; when a key is not mapped, but `map.getOrDefault(key, default_val)` can replace the `null` result compated to `map.get(key)`.

 * Rust: The `get` method of `HashMap` returns an `OptionOf`, the `index` method errors when a key is not found, and `[]` notation uses `index`.

So, the commit proposed here makes Rhombus less Python/Rust-like and more Kotlin/Groovy-like.

### DIFF
--- a/rhombus-lib/rhombus/private/amalgam/composite.rkt
+++ b/rhombus-lib/rhombus/private/amalgam/composite.rkt
@@ -38,6 +38,7 @@
                                                   #:bounds-key [bounds-key #f]
                                                   #:accessor->info? [accessor->info? #f] ; extend composite info?
                                                   #:index-result-info? [index-result-info? #f]
+                                                  #:index-result-maybe? [index-result-maybe? #f]
                                                   #:sequence-element-info? [sequence-element-info? #f]
                                                   #:rest-accessor [rest-accessor #f] ; for a list-like "rest"
                                                   #:rest-to-repetition [rest-to-repetition #'in-list] ; to convert "rest" to a sequence
@@ -73,7 +74,7 @@
           #,steppers #,accessors #,static-infoss
           (a-parsed.infoer-id ... post-a-parsed.infoer-id ...) (a-parsed.data ... post-a-parsed.data ...)
           #,(length post-args)
-          #,accessor->info? #,index-result-info? #,sequence-element-info?
+          #,accessor->info? #,index-result-info? #,index-result-maybe? #,sequence-element-info?
           #,(and rest-arg
                  #`(#,rest-accessor
                     #,rest-to-repetition
@@ -94,7 +95,7 @@
                       steppers accessors ((static-info ...) ...)
                       (infoer-id ...) (data ...)
                       num-post
-                      accessor->info? index-result-info? sequence-element-info?
+                      accessor->info? index-result-info? index-result-maybe? sequence-element-info?
                       rest-data))
      #:with (arg-static-infos ...) (cond
                                      [(syntax-e #'accessor->info?)
@@ -128,16 +129,16 @@
                     (syntax-parse (normalize-static-infos/values 2 maybe-infos)
                       [(() ()) #f]
                       [(car-infos cdr-infos) #'((car car-infos) (cdr cdr-infos))])))
-             (or (and (syntax-e #'index-result-info?)
-                      (static-info-lookup/pair #'static-infos #'#%index-result))
-                 (and (syntax-e #'sequence-element-info?)
+             (or (and (syntax-e #'sequence-element-info?)
                       (static-info-lookup/pair #'static-infos #'#%sequence-element))
+                 (and (syntax-e #'index-result-info?)
+                      (static-info-lookup #'static-infos #'#%index-result))
                  #'())]
             [(#t)
-             (or (and (syntax-e #'index-result-info?)
-                      (static-info-lookup #'static-infos #'#%index-result))
-                 (and (syntax-e #'sequence-element-info?)
+             (or (and (syntax-e #'sequence-element-info?)
                       (static-info-lookup #'static-infos #'#%sequence-element))
+                 (and (syntax-e #'index-result-info?)
+                      (static-info-lookup #'static-infos #'#%index-result))
                  #'())]
             [(#f)
              (define (static-info-lookup/wrap infos key)
@@ -203,7 +204,9 @@
                                                          #'()))
                                                    (append
                                                     (if (syntax-e #'index-result-info?)
-                                                        (list #`(#%index-result #,cdr-infos))
+                                                        (list #`(#%index-result #,(if (syntax-e #'index-result-maybe?)
+                                                                                      #`((#%maybe #,cdr-infos))
+                                                                                      cdr-infos)))
                                                         '())
                                                     (if (syntax-e #'sequence-element-info?)
                                                         (list #`(#%sequence-element
@@ -212,7 +215,9 @@
                                                   [(#t)
                                                    (append
                                                     (if (syntax-e #'index-result-info?)
-                                                        (list #`(#%index-result #,rest-static-infos))
+                                                        (list #`(#%index-result #,(if (syntax-e #'index-result-maybe?)
+                                                                                      #`((#%maybe #,rest-static-infos))
+                                                                                      rest-static-infos)))
                                                         '())
                                                     (if (syntax-e #'sequence-element-info?)
                                                         (list #`(#%sequence-element #,rest-static-infos))
--- a/rhombus-lib/rhombus/private/amalgam/indexable.rkt
+++ b/rhombus-lib/rhombus/private/amalgam/indexable.rkt
@@ -207,7 +207,7 @@
     [(treelist? indexable) (treelist-ref indexable index)]
     [(list? indexable) (list-ref indexable index)]
     [(vector? indexable) (vector-ref indexable index)]
-    [(hash? indexable) (hash-ref indexable index)]
+    [(hash? indexable) (hash-ref indexable index #f)]
     [(set? indexable) (set-ref indexable index)]
     [(string? indexable) (string-ref indexable index)]
     [(bytes? indexable) (bytes-ref indexable index)]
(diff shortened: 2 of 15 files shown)


jackfirth:
Would `list[i]` also change to return false instead of throwing when accessing an out-of-bounds index? Or would `a[b]` behave differently for missing keys in map-like values and list-like values?

mflatt:
> Would `list[i]` also change to return false instead of throwing when accessing an out-of-bounds index?

No, the change here is only for maps, and trying to access a list with a wrong index still throws an exception.

usaoc:
I think making `[]` behave differently than the `.get` method isn’t ideal, because they are expected to behave the same on normal `Indexable` values.  Also, @jackfirth raises a good point that other `.get` methods can also throw, and special treatment for maps *only* isn’t ideal either, but probably accessing missing keys is way more frequent than accessing out-of-bounds indices.

Preliminary idea: we probably want a way for user-defined classes to implement `maybe_get` methods (`MaybeIndexable`?), but then we should have a separate syntax for that, in case a class implements both `maybe_get` and `get`.

mflatt:
It would be relatively easy to have `MaybeIndexable` where `[]` tries that interface first, then falls back to `Indexable`. That would resolve the difference between `[]` and `Map.get`, assuming we add a `Map.maybe_get` that is like `Map.get` with a `#false` second argument.

As for a separate syntax — that turns out to be a more difficult direction, as far as I can tell. See "Some other things I tried".

jackfirth:
Matthew, can you create a pull request or commit or other diffable thing that shows how different the code you're writing looks with and without this feature? I'm personally not sure the added complexity of a `MaybeIndexable` interface and a difference between how missing indices are handled by `[]` notation on the two most common kinds of collections is worth being able to write `map[key]` instead of `map.get(key, #false)`. But it's hard to understand that tradeoff without looking at some real code.

mflatt:
> Matthew, can you create a pull request or commit or other diffable thing that shows how different the code you're writing looks with and without this feature?

In this PR, see "rhombus/run.rhm". That has changes not directly related to `[]`, but you'll see places where `.get(key, #false)` changed to `[key]`. In the case of `#'evals`, it was better to start the map off with an initial value, but in most other cases it's nicer to get `#false` for unmapped keys.

mflatt:
For lots more examples of using `[]` with `#false` as the default, see https://github.com/mflatt/netsec-slides. That code doesn't depend on these changes here, but it uses `OptMap` in many places (grep for uses), and so that could be simplified somewhat after this change. Meanwhile, there's one module that would need changes to accomodate `[]` having `maybe` static info, so I pushed that as a branch: https://github.com/mflatt/netsec-slides/tree/ref-maybe

jackfirth:
Looking at `rhombus/run.rhm`, the cases I see where `map.get(key, #false)` are rewritten to `map[key]` are also cases there the only value for that key that could be in the map is `#true`. Those could also be written as `map.has_key(key)`, correct?

After taking a cursory look through netsec-slides, the uses of falsey indexing I've seen so far seem to fall into two categories:

- The value mapped by the key is ignored, e.g. `if options[#'reinfect]`. These can be `map.has_key(key)` instead.
- The value mapped by the key is immediately combined with `||` to get a default, e.g. `options[#'compiler_src] || 0`. These can be `map.get(#'compiler_src, default)` instead.

I'm not actually seeing many cases where one would write `map.get(key, #false)`. So perhaps we shouldn't be talking about this change as an alternative to `map.get(key, #false)`, since a lot of the cases in practice would be more clearly written in other ways.

I think we also should consider the cost this change imposes on error messages. Consider this code:

```
let foo = map[unintentionally_missing_key]
do_thing_that_doesnt_expect_false(foo)
```

Previously, this throws an error from `map[unintentionally_missing_key]` with a message that includes `unintentionally_missing_key`. Now however, `do_thing_that_doesnt_expect_false(foo)` would fail with a message saying it got `#false` unexpectedly. That's harder to debug since we don't know which key was missing. Given this, I think `map[key]` returning false would need to be the user-desired behavior significantly _more_ than 50% of the time to justify the choice, since it comes at the cost of error message quality.

mflatt:
> Those could also be written as `map.has_key(key)`, correct?

Yes. But the point is to write an expression with `[]`, not something the shape of a method call. I get that using `[]` won't seem important to everyone, but in the context of Rhombus, not being able to use a common-domain specific notation and falling back to function/method notation feels like something going wrong.

There is definitely a trade-off in error reporting. I would have said that it's obvious which trade-off to make, but then I wrote some Rhombus programs, and now I'm not so sure. Errors from unexpectedly missing keys — particularly errors that didn't show up write after writing the reference in question — have felt like less trouble and given me less pause in writing code than having to avoid `[]` when I know the key might not be there.

jackfirth:
> Yes. But the point is to write an expression with `[]`, not something the shape of a method call. I get that using `[]` won't seem important to everyone, but in the context of Rhombus, not being able to use a common-domain specific notation and falling back to function/method notation feels like something going wrong.

What if we tried to come up with some domain-specific notation for `map.has_key(key)`? That might let us have our cake and eat it too, if we can imagine a good notation.

jackfirth:
I believe python uses `key in map` notation for this, which seems reasonable. It could work with sets too.

sorawee:
> The value mapped by the key is ignored, e.g. if options[#'reinfect] (where the map never has #'reinfect explicitly assigned to #false). These can be map.has_key(key) instead.

I totally agree with this. It could be even more ergonomic by doing the Python-style `key in map`.

> The value mapped by the key is immediately combined with || to get a default, e.g. options[#'compiler_src] || 0. These can be map.get(#'compiler_src, default) instead.

Not exactly. If `default` has side-effect or is computationally intensive, doing the `#false` + `||` option is more appropriate than `.get(..., default_value)`. 

But there are also other ways to accomplish this. In Racket, it's giving a thunk to `hash-ref`. In Rust, it's doing `unwrap_or_else` on the option type.

mflatt:
> What if we tried to come up with some domain-specific notation for map.has_key(key)?

Ah, that's a idea that I haven't tried or considered before! I'll give `in` a try. I also like that `in` could replace `[]` for checking membership in a set.

usaoc:
An `in`-like notation makes sense for sets, and probably for other set-like operations like `List.has_element`, `Range.has_element`, etc.  I’m not sure if there’s a similar notation that would make sense for *mutable* sets, specifically the `mset[elem] := in` syntax (add `elem` if `in` is true, delete `elem` otherwise), but probably we should encourage people to use (the missing) `.add` and `.delete` methods instead?

mflatt:
Yes, that was my thinking. I don't think I'll be too sad to lose a notation for adding to or deleting from a mutable set, but we'll see.

mflatt:
Adding `in` is a clear win for Rhombus oveall. Using `in` in place of a `[]` for a map seems a little awkward to me, still. I still suspect I'd be happier in the long run with `maybe` behavior from `[]`, but I'll yield to consensus here and leave it as-is. Closing the PR.

PR #616, opened 2025-01-26
add `shrubbery/text` meta-language

### LINKED ISSUE #612: Add text mode to `#lang shrubbery`
The [shrubbery language](https://docs.racket-lang.org/shrubbery/language.html) currently parses the source invariably in `top` mode, which is the default one of `parse-all`:

https://github.com/racket/rhombus/blob/master/shrubbery-lib/shrubbery/main.rkt#L24

I would welcome an option (or a second language) such that parsing happens in `text` mode.

usaoc:
Like `shrubbery` meta-language, but parses the module in `'text` mode.

Resolves #612.


mflatt:
LGTM - thanks!

PR #618, opened 2025-01-31
Change `List.index` and `List.find_index`

mflatt:
The old `List.index` was like `List.find`, taking a predicate as an argument. The new `List.index` that takes a value to locate (like `List.contains`), which makes it more like `treelist-index-of` and Python's `index` method on list.

The old `List.index` is renamed here to `List.find_index`. I'm not sure whether it should be kept.

Ditto for methods in `PairList` and `MutableList`.


mflatt:
Not sure why I added `List.index` that way in the first place. I think I was confused.

PR #619, opened 2025-02-02
change regexp `#%comp` to recognize `min ..= max`

usaoc:
In `pat {min .. max}`, `pat` would be matched from `min` (inclusive) to `max` (inclusive) times, but expression `min .. max` doesn't include `max`.  Instead, the new `pat {min ..= max}` form corresponds to expression `min ..= max`, which includes `max`.  The `pat {min ..}` form is still kept as-is, corresponding to expression `min ..`.

Also add `..`/`..=` to regexp pattern space, and change `#%comp` to recognize them by bindings.

### DIFF
--- a/rhombus-lib/rhombus/private/amalgam/rx.rhm
+++ b/rhombus-lib/rhombus/private/amalgam/rx.rhm
@@ -93,6 +93,9 @@ export:
     names:
       #%literal #%comp #%parens #%juxtapose #%call #%brackets #%index
 
+    // recognized by `#%comp`
+    .. ..=
+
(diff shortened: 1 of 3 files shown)


mflatt:
Aligning to `..` and `..=` as in expressions seems like a clear improvement.

I'm unsure about matching `..` and `..=` by expression-space binding, though. Would be it better to bind `..` and `..=` in the pattern space just for recognition by `#%comp`?

usaoc:
Updated to add and recognize `..`/`..=` in regexp pattern space.

mflatt:
LGTM - thanks!

PR #620, opened 2025-02-05
support range arguments in `substring`-like operations

usaoc:
The main change is described in the third commit’s message, copied below:

Support range arguments in `substring`-like operations, including `String.substring`, `Bytes.subbytes`, and `(Mutable)List.sublist`.  When given one argument, that argument should be a range, and `start` and `end` are derived akin to Python slice notation or Rust `SliceIndex` impls for range(-like) objects.  (It's actually more similar to specifically Rust `SliceIndex` impls for `(Bound<usize>, Bound<usize>)`, because we have 3 * 3 = 9 types of ranges.)

This is implemented on `substring`-like operations, instead of Python-like `[]` slice forms, because we don't have (ad-hoc) parametric result annotations, and overloading `[]` (i.e., `.get` methods) this way makes static info less useful, in addition to performance considerations.

In principle, more operations that take `start` and `end` arguments can be implemented this way, as opposed to making `start` and `end` optional.


mflatt:
This looks good to me, and I have been trying to write `substring` and `sublist` with two arguments always, so far, in anticipation of this change. Thanks!

PR #623, opened 2025-02-10
more consistently name `maybe(....)` conversion functions

usaoc:
Change `ModulePath.try` to `ModulePath.maybe`; add `String.maybe_to_{int,number}` and change `String.to_{int,number}` to throw exceptions instead of returning `#false`.  In general, `maybe(....)` conversion functions should include `maybe` in their names, and variants without `maybe` should throw an exception instead of returning `#false`.

Also fix behavior of `String.to_{int,number}` annotations on mutable strings, that is, they should not match.


mflatt:
LGTM

PR #636, opened 2025-03-20
Improve repetition syntax error message

jackfirth:
The [Repetitions](https://docs.racket-lang.org/rhombus/repetition.html) documentation includes this example code:

```
> def five = 5
> [five, ...]
five: used with wrong repetition depth
  expected: 0
  actual: 1
```

In cases where the bound variable isn't a repetition at all, like above, I find the error message confusing. So this pull request adds a special case for that which changes the message to this:

```
> [five, ...]
five: not a repetition binding
  used repetition depth: 1
```

This new message is only shown when the variable isn't a repetition and is used in any repetition position. In all other cases, the original error message is used.


mflatt:
Thanks! I adjusted a little to make it clearer in the source that a choice is made between two different error formats, but I expect that's uncontroversial (and we can readjust if not).

PR #637, opened 2025-03-21
Add `guard` and `guard.let`

jackfirth:
This pull request adds _guard statements_ (as seen in Swift and Rust) in the form of two new control flow utility macros, `guard` and `guard.let`. They're definition sequence macros that give straight-line code the ability to escape early if a condition isn't met. Their semantics are the same as in my [`guard`](https://docs.racket-lang.org/guard/) package for Racket, with the exception that `define/guard` and `guarded-block` aren't needed. As an example, here are two versions of a `list_equals(xs, ys)` function, one with guard statements and one without:

```haskell
fun list_equals(xs :: List, ys :: List):
  cond
  | xs == []: ys == []
  | ys == []: #false
  | ~else:
      let [x, & rest_xs] = xs
      let [y, & rest_ys] = ys
      x == y && list_equals(rest_xs, rest_ys)
```

```haskell
fun list_equals(xs :: List, ys :: List):
  guard.let [x, & rest_xs] = xs | ys == []
  guard.let [y, & rest_ys] = ys | #false
  x == y && list_equals(rest_xs, rest_ys)
```

The `guard` form checks that an expression is true; otherwise, it short-circuits the enclosing block and evaluates the given alternative. The `guard.let` form checks that a pattern matches an expression or else short-circuits. They're equivalent to putting the remainder of the block in an enclosing `if` or a `match` expression. Here is a longer piece of example code that uses a mix of `guard` and `guard.let`statements:

```haskell
fun foo(x, y):
  guard.let x :: Int = x | #false
  guard.let y :: Int = y | #false

  println("looks like the types are right")
  println("so far so good")

  guard x > 0 | #false
  guard y > 0 | #false

  println("both values positive")

  guard x > y
  | println("x must be greater than y")
    #false

  guard.let delta :: Int = x - y
  | println("this is supposed to be impossible")
    println("compiler bug?")
    #false

  delta
```

Tests and documentation are not yet included, as I wanted to get feedback on the proposal first.

### DIFF
--- a/rhombus-lib/rhombus/private/amalgam/guard.rhm
+++ b/rhombus-lib/rhombus/private/amalgam/guard.rhm
@@ -6,36 +6,92 @@ import:
 export:
   guard
 
+meta syntax_class GuardAndSuccess(stx):
+  kind: ~multi
+  fields: [condition, ...]
+          failure_body
+          success_body
+| '$condition ... | $failure_body
+   $(success_body :: Multi)':
+    when '$condition ...' matches ''
+    | syntax_meta.error("empty condition for guard",
+                        stx)
+    when failure_body matches ''
+    | syntax_meta.error("empty failure body for guard",
+                        stx)
+    when success_body matches ''
+    | syntax_meta.error("empty success sequence after guard",
+                        stx)
+| '$(seq :: Sequence)
+   $(_ :: Multi)':
+    match seq
+    | '$_ ... | $__body | ...':
+        syntax_meta.error("expected a single `|` alternative for failure body",
+                          stx,
+                          seq)
+    | ~else:
+        syntax_meta.error("missing `|` alternative for failure body",
+                          stx,
+                          seq)
+    field [condition, ...] = []
+    field failure_body = #false
+    field success_body = #false
 
-defn.sequence_macro 'guard $condition ... | $failure_body
-                     $success_body':
-  values(
-    'if $condition ...
-     | $success_body
-     | $failure_body',
-    '')
+defn.sequence_macro 'guard $seq ... ~nonempty
+                     $rest':
+  ~all_stx: stx
+  match '$seq ...; $rest':
+  | '$(g :: GuardAndSuccess(stx))':
+      values(
+        'if $g.condition ...
+         | $g.success_body
+         | $g.failure_body',
+        ''
+      )
 
-
-meta syntax_class GuardRightHandSide:
-  fields: as_expression
-  root_swap: as_expression raw_sequence
-| ': $e ...':
-    field as_expression = '$e ...'
-| ': $(b :: Block)':
-    field as_expression = 'block $b'
-| '$(bound_as expr_meta.space: '=') $e ...':
-    field as_expression = '$e ...'
-
+meta syntax_class GuardBindAndRightHandSide(who):
+  fields: [bind, ...]
+          rhs_expression
+| '$bind ... ~nonempty: $e ...':
+    field rhs_expression = '$e ...'
+    when rhs_expression matches ''
+    | syntax_meta.error(~who: who,
+                        "empty expression for binding",
+                        '$bind ...')
+| '$bind ... ~nonempty: $(b :: Block)':
+    field rhs_expression = 'block $b'
+| '$bind ... ~nonempty $(eql && bound_as expr_meta.space: '=') $e ...':
+    field rhs_expression = '$e ...'
+    when '$bind ... $e ...' matches '$_ ... $(bound_as expr_meta.space: '=') $_ ...':
+    | syntax_meta.error(~who: who,
+                        "multiple immediate equals not allowed in this group;\n"
+                          ++ " use parentheses to disambiguate",
+                        eql)
+    when '$e ...' matches ''
+    | syntax_meta.error(~who: who,
+                        "empty expression for binding",
+                        '$bind ...')
+| '$bind ...':
+    syntax_meta.error(~who: who,
+                      "expected binding followed by expression or block",
+                      '$bind ...')
+    field rhs_expression = #false
 
 namespace guard:
   export:
     rename guard_let as let
 
-  defn.sequence_macro 'guard_let $test_pattern ... $(rhs :: GuardRightHandSide)
-                       | $failure_body
-                       $success_body':
-    values(
-      'match $rhs
-       | $test_pattern ...: $success_body
-       | ~else: $failure_body',
-      '')
+  defn.sequence_macro 'guard_let $seq ... ~nonempty
+                       $rest':
+    ~op_stx: who
+    ~all_stx: stx
+    match '$seq ...; $rest':
+    | '$(g :: GuardAndSuccess(stx))':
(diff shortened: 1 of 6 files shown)


rhombus-lib/rhombus/private/amalgam/guard.rhm
```diff
@@ -0,0 +1,32 @@
...
+
+import:
+  "core-meta.rkt" open
+
+export:
+  guard
+
+
+defn.sequence_macro 'guard $condition ... | $failure_body
+                     $success_body':
```
distractedlambda, line 11: Does `$success_body` need a `...`, or does `defn.sequence_macro` do something where a plan trailing `$`-binding implicitly matches the remainder of input?

distractedlambda:
I _really_ like the look of this (especially as someone who's used Rust and Swift a fair bit). It also hadn't even occurred to me to use `defn.sequence_macro` for this (I had imagined some sort of escape-continuation-based thing, but this seems much better).

One thing I'm still slightly wrestling with is that it _does_ use the purely syntactic notion of a defn-or-expr sequence for what constitutes the code to short-circuit. This is almost certainly for the best, since it keeps it easy to understand how `guard` will behave in combination with other macros... I guess it just feels a little strange when I try to analogize it to early-return in Swift/Rust/Zig, which cares only about function boundaries (but then for that sorta thing, explicit escape continuations are probably the right answer for Rhombus).

samth:
One thing that would be really nice would be to have a way of producing a failure result automatically based on static information for the enclosing context, and then having `?` do that as well, similar to how it works in Rust. I'm not sure if that's possible currently though.

rfindler:
This is a really nice idea! 

> One thing that would be really nice would be to have a way of producing a failure result automatically based on static information for the enclosing context, and then having `?` do that as well, similar to how it works in Rust. I'm not sure if that's possible currently though.

Ah, that would be great. I get hung up on the syntax of the examples with this extra "guard" stuff. But if there was a specific sentinel value that these bindings worked with somehow and you could write patterns that explicitly didn't include that sentinel and then that would mean "abort to the enclosing block" but doing it all statically, not via (escaping) continuations, I think that would make the notation look very nice. And DrRacket/Emacs/VSCode could offer you some kind of feedback when you've done that so you could easily see the shortcircuiting potential.

rhombus-lib/rhombus/private/amalgam/guard.rhm
jackfirth: The latter. `$success_body` here matches the remainder of the enclosing block.

distractedlambda:
> One thing that would be really nice would be to have a way of producing a failure result automatically based on static information for the enclosing context, and then having `?` do that as well, similar to how it works in Rust. I'm not sure if that's possible currently though.

I like the idea of having a syntactically-lightweight error-handling mechanism that "synthesizes" the aborting path, but my feeling is that that makes more sense as a separate mechanism / construct. Though I might just be biased by analogizing to Rust/Swift/Zig (where constructs like `guard` or `if let` are distinct from `try`).

I also think there may be a question there of "how far to escape". In Rust/Swift/Zig, the `try` construct (which is postfix-`?` in Rust, much to my annoyance) is hard-wired to escape out of the enclosing _function_, and I think that's probably the right behavior for the majority of uses. I could imagine that working semi-statically by having a "current-escape-continuation-that-`try`-should-call" tracked as a syntax parameter, potentially with every `fun` form automatically introducing one of those at the top. Though, we'd want to be sure that the compiler removes the overhead of these when they're not actually used.

mflatt:
> I also think there may be a question there of "how far to escape". In Rust/Swift/Zig, the try construct (which is postfix-? in Rust, much to my annoyance) is hard-wired to escape out of the enclosing function, and I think that's probably the right behavior for the majority of uses.

Unless I misunderstand, this doesn't sound like an approach that would work well for Rhombus. Things that look like nested-block forms may actually be implemented with `fun` via macro expansion. For example, the body of a `check` form is under `fun`. We could have different kinds of `fun`s, but the ease of converting `<expr>` to `(fun (): <expr>))()` without thinking too hard about it is probably something we don't want to lose from Scheme. There's also a question of tail calls. The body of `let/cc` is not in tail position, so it wold be bad for every function to wrap one around its body; maybe there's a variant of `let/ec` that would do the right thing, but I'm not immediately sure.

Along the lines of functional-versus-imperative style, I think I wouldn't want to include a `guard` form that tries to jump out to a function boundary, even if macros and tail calls were not a problem. Continuations/gotos are difficult to reason about, and I doubt that I'd want an implicit `goto` point at the start of every function.

A block-based `guard` form makes sense to me, and I'm ok with `guard` in Rhombus being inspired by Swift and Rust without being the same thing. If there's a different name than `guard` with a closer precedent, that would be good, but I don't happen to know one.

distractedlambda:
> Unless I misunderstand, this doesn't sound like an approach that would work well for Rhombus.

I hadn't considered implicit introduction of `fun`s when I wrote that, but I agree fully with your reasoning. My original reasoning for why function boundaries seemed "more right" was that it would avoid having to wrap nested block expressions in more `guard`s or `try`s just to fully bubble up an error... though it's not like function boundaries are always the right granularity for that, either.

FWIW, limited forms of `goto` _are_ regaining popularity in mainstream languages. Both Rust and Zig let you attach lexically-scoped labels to block expressions, and to `break` out of such an expression (supplying a value for the expression result) regardless of nesting depth, _as long as you don't cross a function boundary_. Kotlin goes further by allowing you also cross function boundaries in certain situations (specifically, you can cross the boundary of a lambda when you have already forced it to be `inline`d).

mflatt:
> Rust and Zig let you attach lexically-scoped labels to block expressions, and to break out of such an expression

Something like `let/cc` (with a better name) would be a good idea to add to Rhombus. Something that explicitly declares a named point that might be jumped to — and where I can easy find all uses of the name — is certainly ok.

jackfirth:
This is now ready for review. I'm in agreement with Matthew that a more `return`-shaped or `goto`-shaped form of block-based control is much trickier to achieve in Rhombus due to macros and implicit lambdas. Something static there seems achievable, but I would consider it beyond the scope of `guard`. Future work.

Along similar lines, a Rust-like `?` operator for escaping would also be difficult. I like the idea, but I don't know how to achieve it in Rhombus yet. This too, I think we can consider to be future work. For now, simple block-based short-circuiting seems to be the way to go.

mflatt:
Thanks for this very thorough PR!

I pushed two candidate commit. The first make minor adjustments to docs, and I it's probably not controversial.

For the second added commit, I started by trying to adjust the pattern in small ways to improve syntax errors, but that didn't work out. I ended up trying much bigger changes to tailor the error messages. The result is uglier and more complex, but it provides much better error messages. This is maybe an old story, in that `syntax-parse` can only sometimes provide good errors. Still, the gap between the automatic and tailored messages is really big, though, and I'm not seeing a clear way to reduce that gap. I think it's probably worthwhile to customize errors like this for built-in forms, but I'm not certain.

PR #638, opened 2025-03-22
add `net/url` and `net/http`

mflatt:
The `net/url` library builds on `net/url-string` and `net/uri-code`. It's in its own package, `rhombus-url-lib` plus `rhombus-url`.

The `net/http` library builds on `http-easy`. It's also in its own package, `rhombus-http-lib` plus `rhombus-http`.

So, to try out the change, use
```
raco pkg install rhombus-url-lib/ rhombus-url/ rhombus-http-lib/ rhombus-http/
```
in a Rhombus repo clone and accept dependency installs.

I think the general approach here is probably uncontroversial, so if there are no objections in the next day or so, I'll merge and register the packages, and then we can refine from there. (In other words, I don't think we're committed to "rare breaking changes" right away for new modules outside of `#lang rhombus`.)

### DIFF
--- a/rhombus-url/rhombus/net/scribblings/rhombus-url.scrbl
+++ b/rhombus-url/rhombus/net/scribblings/rhombus-url.scrbl
@@ -1,0 +1,212 @@
+#lang rhombus/scribble/manual
+@(import:
+    meta_label:
+      rhombus open
+      net/url open)
+
+@title(~style: #'toc){Rhombus URL and Form Parsing}
+
+@docmodule(net/url)
+
+The @rhombusmodname(net/url) library provides functions for parsing and
+manipulating URLs, as well as functions for HTML form and URI component
+encoding and decoding.
+
+@table_of_contents()
+
+@section(~tag: "url"){URL Parsing}
+
+@doc(
+  class URL(
+    ~scheme: scheme :: maybe(String) = #false,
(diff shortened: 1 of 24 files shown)


rhombus-url/rhombus/net/scribblings/rhombus-url.scrbl
```diff
@@ -0,0 +1,212 @@
...
+    params :: List.of(String)
+  )
+  class KeyValue(
+    key :: Symbol,
+    value :: maybe(String)
+  )
+){
+
+ A class to represent URLs as defined by
+ @hyperlink("http://www.ietf.org/rfc/rfc1939.txt"){RFC 3986}. The
```
LiberalArtist, line 41: Seems like the wrong link: maybe <https://www.rfc-editor.org/rfc/rfc3986.html>? (Racket links to <https://www.ietf.org/rfc/rfc2396.txt>: I'm not sure if there are relevant changes.)

mflatt: Thanks! It seems that the Racket docs link to RFC 2396 at the top and RFC 3986 at the `url` struct. I think the library is as consistent with RFC 3986 as RFC 2396 (though it's difficult to be sure), so I'll change to always refer to the newer one.

PR #642, opened 2025-03-24
JSON

mflatt:
This is @samdphillips's JSON wrapper in a "rhombus-json" and "rhombus-json-lib" package:

* `json.JSON` - the annotation
* `json.read`, `json.from_string`, `json.from_bytes` - parsing
* `json.write`, `json.to_string`, `json.to_bytes` - printing

Some notable choices in this implementation:

* Keys in a map `json.JSON` are strings, not symbols.
* `json.JSON` recognizes lists, maps with string keys, etc., without wrapping. The result of a successful `json.JSON` check via `is_a` or `::` is cached (similar to `list?` in Racket), so it's amortized constant-time for satisfying values.
* `json.from_string` and `json.from_bytes` complain if the given string or byte string doesn't contain a single JSON representation. That's different from the Racket library, where `(string->jsexpr "")` produces `eof` and `(string->jsexpr "1 2")` produces `1`.
* `json.read`, `json.from_string`, and `json.from_bytes` have a `~replace_malformed_surrogate` argument that defaults to `#false`, the same as the Racket library.

### DIFF
--- a/rhombus-json-lib/rhombus/json.rhm
+++ b/rhombus-json-lib/rhombus/json.rhm
@@ -1,0 +1,124 @@
+#lang rhombus/static/and_meta
+
+import:
+  lib("racket/base.rkt").#{make-immutable-hashalw}
+  lib("json/main.rkt") ! #{for-extension} as rkt_json:
+    rename #{read-json*} as read
+    rename #{write-json*} as write
+
+def known_json = WeakMutableMap.by(===)()
+
+// Using a `json` namespace adds a prefix to names
+namespace json:
+  export:
+    JSON
+    write
+    to_string
+    to_bytes
+    read
+    from_string
+    from_bytes
+
+  fun is_json(v):
+    match v
+    | _ :: (Int || (Flonum && Rational)): #true
+    | _ :: String: #true
+    | _ :: Boolean: #true
+    | #'null: #true
+    | vs :: List:
+        cond
+        | vs in known_json: #true
+        | (for all (v in vs): is_json(v)):
+            known_json[vs] := #true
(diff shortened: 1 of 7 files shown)


rhombus-json-lib/rhombus/json.rhm
```diff
@@ -0,0 +1,124 @@
...
+    write
+    to_string
+    to_bytes
+    read
+    from_string
+    from_bytes
+
+  fun is_json(v):
+    match v
+    | _ :: (Int && Exact) || (Real && Inexact && Rational): #true
```
sorawee, line 24: From Discord:

> @usaoc:
> Yeah, maybe Int || (Flonum && Rational) instead

samth:
I will say that I have repeatedly found the ability to parse multiple JSON entries from a stream very useful in Racket, and so I think maintaining that (at least as an option) would be better.

samth:
Ah, I see that `read()` maintains the Racket behavior; it's just the `from_string()` method that checks.

rhombus-json-lib/rhombus/json.rhm
mflatt: Adjusted - thanks.

usaoc: Note that `Flonum` would not match single-precision flonums, so it’s not totally like `Inexact && Real`.

mflatt: I think `Flonum` is probably the right choice to ensure that information isn't lost when writing and reading, so I've adjusted the docs to not suggest `Inexact`.

Issue #129, opened 2019-11-02
Redesign xexpr


sorawee:
Currently, xexpr doesn't support HTML5 [boolean attributes](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes#Boolean_Attributes). While HTML5 is not directly related to XML, xexpr has been widely used to represent HTML, so having this support would be great.

racket/racket#2885 could also potentially be fixed elegantly with a change of grammar. The problem is that attributes only support string, but in fact `valid-char?` could be in attributes too. The conservative solution is to convert `valid-char?` to string to make it representable, but if we allow grammar change, allowing `valid-char?` in the attributes one be one feasible solution. Another solution would be to get rid of `valid-char?` entirely from the whole grammar.

Lazerbeak12345:
One possible idea for the boolean thing is just to allow for this as a boolean attribute:

```
'(span (@ (boolean)) "Text content")
```

which would become

```
<span boolean>Text content</span>
```

```
'(span (@ (boolean "")) "Text content")
```

would become

```
<span boolean="">Text content</span>
```

```
'(span (@ (boolean "foobar")) "Text content")
```

would become

```
<span boolean="foobar">Text content</span>
```

This would mean the least amount of change, and wouldn't be hard to detect at all.

Much better than `'(span (@ (boolean #t)) "Text content")` and `<span boolean="boolean">Text content</span>` (and conversely `'(span (@ (boolean #f)) "Text content")` and `<span>Text content</span>`) as it wouldn't maintain the html conventions anymore.

PR #644, opened 2025-03-25
XML library

mflatt:
Includes a "XML syntax object" representation, which is analogous the to the Racket layer's "xexprs". The `xml.syntax` form is essentially an alias of `''`, but it checks that the result syntax object (after escapes are replaced) conforms to constraints for an XML syntax object.

```
    def my_doc:
      xml.syntax:
        words:
          greeting:
            ~weight: "bold"
            "hello"
          parting:
            ~weight: "normal"
            ~size: $(to_string(2 * 12))
            "bye"

  > my_doc
  'words:
     greeting:
       ~weight: "bold"
       "hello"
     parting:
       ~weight: "normal"
       ~size: "24"
       "bye"'

    > xml.write_syntax(my_doc, ~indentation: #'peek)
    <words>
      <greeting weight="bold">hello</greeting>
      <parting weight="normal" size="24">bye</parting>
    </words>
```

See also #643, which was motivated in part by the use of keywords for attributes in this PR's XML syntax objects. But this `xml` library also supports swapping `_` and `-` on conversion to/from XML syntax objects and XML or `xml.Document`, which seems likely useful in applications that tend to use `-` and not `_`.

The non-syntax-object representation mostly follows the Racket library, except that `cdata` is split into `xml.CData` and `xml.Verbatim`, where `xml.CData` adds `<![CDATA[` and `]]>` when rendering to XML, while `Verbatim` lets you subvert the encoding and write raw text (which I think is a common use of `cdata` in the Racket `xml` library).

"Permissive" mode for XML syntax objects is supported in the same way as in xexprs. I'm not sure that's a good idea, but I find several places where permissive mode is used with Racket `xml`, so I took that as a sign that it's useful.

Currently, this `xml` library doesn't try to support XHTML or other legacy features of the Racket `xml` library.

### DIFF
--- a/rhombus-xml-lib/info.rkt
+++ b/rhombus-xml-lib/info.rkt
@@ -1,0 +1,13 @@
+#lang info
+
+(define collection 'multi)
+
+(define deps
+  '("base"
+    "rhombus-lib"))
+
+(define pkg-desc "implementation (no documentation) part of \"rhombus-xml\"")
+
+(define license '(Apache-2.0 OR MIT))
+
+(define version "0.1")
--- a/rhombus-xml-lib/rhombus/xml.rhm
+++ b/rhombus-xml-lib/rhombus/xml.rhm
@@ -1,0 +1,181 @@
+#lang rhombus/static/and_meta
+import:
+  lib("xml/main.rkt") as rkt
+  "xml/private/class.rhm":
+    expose:
+      Document
+      Element
+  "xml/private/to_rhm.rhm" open
+  "xml/private/to_rkt.rhm" open
+  "xml/private/to_xstx.rhm" open
+  "xml/private/from_xstx.rhm" open
+  "xml/private/swap.rhm" open
+
+export:
+  all_from(.class)
+  to_syntax
+  from_syntax
+
+namespace xml:
+  export:
+    read
+    from_string
+    from_bytes
+    read_syntax
+    string_to_syntax
+    bytes_to_syntax
+
+    write
+    to_string
+    to_bytes
+    write_syntax
+    syntax_to_string
+    syntax_to_bytes
+
+    Indentation
+
+    current_read_normalize
+    current_read_comments
+    current_read_processing_instructions
+    current_read_count_bytes
+    current_read_collapse_whitespace
+    current_write_empty_shorthand
+
+    doc
+    syntax
+
+  fun read(~in: inp :: Port.Input = Port.Input.current(),
+           ~read_post_misc: read_post_misc = #true)
+    :~ Document:
+      let x:
+        parameterize { rkt.#{read-comments}: current_read_comments(),
+                       rkt.#{xml-count-bytes}: current_read_count_bytes(),
+                       rkt.#{collapse-whitespace}: current_read_collapse_whitespace() }:
+          if read_post_misc
+          | rkt.#{read-xml}(inp)
+          | rkt.#{read-xml/document}(inp)      
+      to_rhm(x,
+             ~keep_pi: current_read_processing_instructions(),
+             ~normalize: current_read_normalize())
+
+  fun read_syntax(~in: inp :: Port.Input = Port.Input.current())
+    :~ Syntax:
+      to_syntax(read(~in: inp))
+
+  fun from_string(str :: String)
+    :~ Document:
+      read(~in: Port.Input.open_string(str))
+
+  fun string_to_syntax(str :: String)
+    :~ Document:
+      to_syntax(read(~in: Port.Input.open_string(str)))
+
+  fun from_bytes(bstr :: Bytes)
+    :~ Document:
+      read(~in: Port.Input.open_bytes(bstr))
+
+  fun bytes_to_syntax(bstr :: Bytes)
+    :~ Document:
+      to_syntax(read(~in: Port.Input.open_bytes(bstr)))
+
+  enum Indentation:
+    none
+    peek
+    scan
+
+  fun write(doc :: Document || Element,
+            ~out: outp :: Port.Output = Port.Output.current(),
+            ~indentation: indentation :: Indentation = #'none):
+    parameterize { rkt.#{permissive-xexprs}: class.current_permissive(),
+                   rkt.#{empty-tag-shorthand}: (if current_write_empty_shorthand()
+                                                | #'always
+                                                | #'never) }:
+      if doc is_a Document
+      | rkt.#{display-xml}(to_rkt(doc), outp,
+                           ~indentation: indentation)
+      | rkt.#{display-xml/content}(element_to_rkt(doc), outp,
+                                   ~indentation: indentation)
+
+  fun write_syntax(xstx :: Syntax,
+                   ~swap_underscore_dash: swap = default_swap,
+                   ~out: outp :: Port.Output = Port.Output.current(),
+                   ~indentation: indentation :: Indentation = #'none):
+    ~who: who
+    let doc = _from_syntax(~who: who,
+                           xstx,
+                           ~swap_underscore_dash: swap)
+    write(doc,
+          ~out: outp,
+          ~indentation: indentation)
+
+  fun to_string(doc :: Document || Element,
+                ~indentation: indentation :: Indentation = #'none)
+    :~ String:
+      let outp = Port.Output.open_string()
+      write(doc,
+            ~out: outp,
+            ~indentation: indentation)
+      outp.get_string()
+
+  fun to_bytes(doc :: Document || Element,
+               ~indentation: indentation :: Indentation = #'none)
+    :~ Bytes:
+      let outp = Port.Output.open_string()
+      write(doc,
+            ~out: outp,
+            ~indentation: indentation)
+      outp.get_bytes()
+
+  fun syntax_to_string(xstx :: Syntax,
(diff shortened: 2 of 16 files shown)


LiberalArtist:
It might be worth considering whether to stick so closely to the representation choices of Racket&rsquo;s `xml` library or not.

In particular, it looks like `EntityInt` represents numeric character references in the same way as the Racket `valid-char?`. This can be useful in contexts when you care about the concrete syntax, e.g. to read, modify, and write and XML document with minimal diff.

However, at the level of the [XML infoset](https://www.w3.org/TR/xml-infoset/), numeric character references and (well-formed) `<![CDATA[...]]>` sections are just interchangeable concrete syntaxes for [character data](https://www.w3.org/TR/REC-xml/#syntax)/[character information items](https://www.w3.org/TR/xml-infoset/#infoitem.character). In many contexts, it would be much more convenient to have them all parsed as strings.

(Symbolic entity references really are different, since their meaning is application-dependent, except for the [predefined entities](https://www.w3.org/TR/REC-xml/#sec-predefined-ent) `amp`, `lt`, `gt`, `apos`, and `quot`.)

samdphillips:
I haven't had a chance to fully look at the PR yet.  Does it address that element and attribute names can contain `:` ?

mflatt:
> It might be worth considering whether to stick so closely to the representation choices of Racket’s xml library or not.
> [...] This can be useful in contexts when you care about the concrete syntax, e.g. to read, modify, and write and XML document with minimal diff.

I don't use XML enough to have strong opinions here. In this first cut, I stuck with the Racket `xml` library's choices, except where I had run into consistent problems with the use of `cdata` as verbatim`. But I think I see what you're saying.

The Racket `xml` library's representation effectively tells a conversion to XML what format it should use. That makes sense for an application where some part of the pipeline has unfortunately become tied to a particular representation, such as using `CDATA` or not. Preserving the representation choices (mostly) when reading fits with that scenario, but also perpetuates and encourages the problem. Normalizing when reading XML would make it easier to process and discourage dependence on an encoding choice.

So, I think it would make sense for the default to be normalizing when reading XML, but also to include an option to turn off normalization. Normalization includes converting numeric and predefined entities to strings, always using `PCData` instead of `CData`, and merging consecutive `PCData`s. Anything else?

Is it better to keep the name `PCData` or change it to something that doesn't name an encoding choice, such as `Text`? I imagine it would still mean the same things as `PCData` for XML output — and, generally, I imagine there would be no change to the output strategy.

LiberalArtist:
> So, I think it would make sense for the default to be normalizing when reading XML, but also to include an option to turn off normalization.

This approach makes sense to me.

> Is it better to keep the name `PCData` or change it to something that doesn't name an encoding choice, such as `Text`?

I think it would be **much** better to avoid the term &ldquo;pcdata&rdquo;, which I think is a misnomer in the Racket `xml` library. The term is a legacy from SGML: in XML, it only occurs [as a keyword in DTDs](https://www.w3.org/TR/REC-xml/#sec-mixed-content) to declare that an element contains &ldquo;mixed content&rdquo;, i.e. &ldquo;character data&rdquo; potentially interspersed with child elements. But the Racket [`pcdata`](https://docs.racket-lang.org/xml/index.html#%28def._%28%28lib._xml%2Fmain..rkt%29._pcdata%29%29) structure actually contains only &ldquo;character data&rdquo; that has already been fully parsed. (IIRC, the term made more sense in SGML, where there were more content models were available that could turn off some parsing, as is the case with certain HTML elements.)

One drawback of `Text` would be that the [XML spec defines &ldquo;text&rdquo;](https://www.w3.org/TR/REC-xml/#dt-text) as &ldquo;a sequence of characters, which may represent markup or character data.&rdquo; The spec uses the term &ldquo;character data&rdquo; for the logical structure we&rsquo;re trying to talk about, but that seems to invite confusion with the use of `CData` for a particular physical representation.

> Normalization includes converting numeric and predefined entities to strings &hellip;

For normalization, yes, this is the right behavior.

When not normalizing, while talking about &ldquo;numeric entities&rdquo; is common, technically I think it is incorrect: [the XML specification](https://docs.racket-lang.org/xml/index.html#%28def._%28%28lib._xml%2Fmain..rkt%29._pcdata%29%29) talks about (numeric) &ldquo;character references&rdquo; vs. (symbolic) &ldquo;entity references&rdquo;.

Also, when not normalizing, it might be more useful to represent a character reference as a `Char` instead of an `Int`.

> &hellip; always using `PCData` instead of `CData`, and merging consecutive `PCData`s. Anything else?

I think that&rsquo;s all as far as normalizing alternate concrete syntax to strings.

Maybe it's worth noting that many applications will still want to do more application-specific normalization that is probably beyond the scope of a general-purpose library. Having application-specific semantics for collapsing whitespace seems particularly common, as does having child elements where order doesn't matter. So the proposed normalization on reading wouldn't eliminate the need for application-specific post-processing, but it would let that post-processing be implemented at a higher level. I think it's particularly worthwhile in a library because, as an application writer, it is easy to overlook the need to test on some whimsical choices of concrete syntax that nonetheless may come up in the wild.

A few less-closely-related thoughts:

1. It might be useful to have an option analogous to [`read-comments`](https://docs.racket-lang.org/xml/index.html#%28def._%28%28lib._xml%2Fmain..rkt%29._read-comments%29%29) for skipping processing instructions.
2. The representation as `List.of(xml.Attribute)`/`(listof attribute?)` allows duplicate attributes, which violate a [well-formedness constraint](https://www.w3.org/TR/REC-xml/#sec-starttags). However, I don't know of this having caused much inconvenience in practice, and the ability to parse &ldquo;XML documents&rdquo; that are ill-formed in this way can be useful.
3. For the feature corresponding to `Verbatim`, [the `sql` library uses `INJECT`](https://docs.racket-lang.org/sql/index.html#%28part._escapes%29), which I think communicates the peril as well as the power of the feature clearly.

sorawee:
> Currently, this xml library doesn't try to support XHTML

Just to be clear, there will be a separate mechanism to support HTML, correct? And that mechanism will also primarily target modern HTML, rather than XHTML?

samth:
I think the key things to support HTML are (1) a parser from HTML, which should probably target `xml.syntax` from this library and (2) a writer to HTML, which should take `xml.syntax` as input. The best examples I know of for Racket are: (1) https://docs.racket-lang.org/html-parsing/index.html and (2) https://docs.racket-lang.org/html-printer/index.html which both use x-expressions. 

From an initial read, both seem like they would be perfectly doable with this `xml.syntax` representation, and that with conversions to/from Racket xexprs those existing libraries could be used to get started.

sorawee:
X-expression and the current implementation of `xml.syntax` doesn't seem to support the preferred form of [Boolean attribute](https://developer.mozilla.org/en-US/docs/Glossary/Boolean/HTML), for example. 

See also #129 which talks about inconsistent handling of `valid-char?`. I haven't taken a look closely how this is handled in `xml.syntax`.

samth:
For boolean attributes, there are two different questions. One is how to print them so that the results are in the preferred form. I think the html-printer library does that correctly now, using xexprs as input, and so there isn't a need to change from that perspective. The other is whether there should be a shorthand syntax for writing them using `xml.syntax`. That seems less clear, and less important, and would make the library more specialized towards HTML.

For #129 isn't the right thing just to parse that example correctly?

LiberalArtist:
> The best examples I know of for Racket are: (1) https://docs.racket-lang.org/html-parsing/index.html

This is an extremely useful library, but it predates adoption of HTML 5. I've found it most useful at getting some structure out of old &ldquo;tag soup&rdquo;. It does not attempt to implement [the WHATWG algorithm for parsing HTML](https://html.spec.whatwg.org/multipage/parsing.html#parsing), which explicitly is not based on XML or SGML.

LiberalArtist:
> See also #129 which talks about inconsistent handling of `valid-char?`. I haven't taken a look closely how this is handled in `xml.syntax`.

> For #129 isn't the right thing just to parse that example correctly?

The linked https://github.com/racket/racket/issues/2885 was useful for context.

IIUC, attribute values are allowed to contain entity references to internal entities (only, i.e. external entities are forbidden), as well as character references, though I don't think I've ever seen an intentional use of an entity reference to a non-predefined entity in an attribute value.

It looks like (but don't take my word for it!) SXML avoids exposing that complexity by fully expanding attribute values to Scheme strings, which I think works out only [because Kiselyov says](https://okmij.org/ftp/Scheme/SXML.html), &ldquo;Internal parsed entities are always expanded by the XML processor at the point of their reference in the body of the document.&rdquo; In contrast, the Racket `xml` library:

> does not provide Document Type Declaration (DTD) processing, including preservation of DTDs in read documents, or validation. It also does not expand user-defined entities or read user-defined entities in attributes.

Maybe a practical compromise would be to expand only character references and predefined entity references in attribute values? Adding any kind of DTD handling would greatly complicate the underlying library (though I've toyed with doing something based on Kiselyov's SSAX parser from time to time).

LiberalArtist:
> Although `xml.Entity` is still present, in normalizing read mode, an numeric character reference is normalized to `xml.Text` with `#'entity` mode.

Belated note: `xml.Entity` doesn't bother me, because it is common casually to refer to numeric character references as "entites", but it occurred to me that `xml.Reference` would be a technically-correct name to encompass both.

PR #649, opened 2025-03-29
disallow 3-D syntax except as initiated via `Syntax.inject`

mflatt:
In a syntax `$` escape or other explicit and implicit uses of `Syntax.make()`, allow only numbers, booleans, strings, byte strings, symbols, keywords, path, srclocs, and `#void`. "3-D" syntax objects can still be constructed, but only by injecting some other kind of value into the syntax-object world with `Syntax.inject`.

The intent is to prevent acciental construction of 3-D syntax, which can be confusing and lead to delayed errors. For example `'$gen()'` would create a syntax object containing the closure `gen` followed by `()`, while `$(gen())` is more likely the intent. Without early detection of 3-D syntax, `'$gen()'` may seem to work until marshaling a compiled use fails.

Meanwhile, when intended, 3-D syntax objects can be useful, particularly since syntax objects in Rhombus take over the role that plain S-expression sometimes serve for Racket. As a concrete example, the `rhombusblock` form in Rhombus Scribble supports `#,` escapes that generate Scribble elements, and those element objects need to sit inside a syntax object that otherwise preserves the source of an argument to `rhombusblock`.

The `Syntax.inject` function is the same as `Syntax.make` on values that `Syntax.make` would accept and treat as syntax atoms. Lists and listables other than pair lists are injected as themselves by `Syntax.inject`, not traversed to find atoms. Pairs and pair lists are simply disallowed.

### DIFF
--- a/rhombus-lib/rhombus/private/amalgam/implicit.rkt
+++ b/rhombus-lib/rhombus/private/amalgam/implicit.rkt
@@ -2,7 +2,8 @@
 (require (for-syntax racket/base
                      syntax/parse/pre
                      shrubbery/print
-                     "srcloc.rkt")
+                     "srcloc.rkt"
+                     "injected.rkt")
          "expression.rkt"
          "binding.rkt"
          "repetition.rkt"
@@ -107,17 +108,18 @@
   (expression-transformer
    (lambda (stxes)
      (syntax-parse stxes
-       [(form-id datum . tail)
-        (check-literal-term #'form-id #'datum)
+       [(form-id raw-datum . tail)
+        (define datum (uninject #'raw-datum))
+        (check-literal-term #'form-id datum)
         (define quoted-datum
           ;; but reraw here
-          (reraw (list #'form-id #'datum)
+          (reraw (list #'form-id datum)
                  ;; copies all props, including originalness
-                 (relocate #'datum
-                           #`(quote #,(no-srcloc #'datum))
-                           #'datum)))
+                 (relocate datum
+                           #`(quote #,(no-srcloc datum))
+                           datum)))
         (values (cond
-                  [(literal-static-infos #'datum)
+                  [(literal-static-infos datum)
                    => (lambda (si)
                         (wrap-static-info* quoted-datum si))]
                   [else quoted-datum])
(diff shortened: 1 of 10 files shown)


usaoc:
I think maps need to be better specified here.  Maps can be serialized just fine, because they are Racket hashes, but there’s an issue with chaperones (chaperone wrappers would be lost, which means custom key comps would be lost, from what I understand).  Moreover, values in maps are recursively converted, so they are not quite “atomic”, and a simple `.unwrap` won’t recover the original maps.

Same thing with arrays, boxes, and `prefab` classes—they can be serialized, and they are not quite “atomic”.

mflatt:
@usaoc Are you saying that certain maps and other values should be explicitly supported by `Syntax.make`? Or just that the behavior of those kinds of values as coerced via `Syntax.inject` should be more specified?

usaoc:
> @usaoc Are you saying that certain maps and other values should be explicitly supported by `Syntax.make`? Or just that the behavior of those kinds of values as coerced via `Syntax.inject` should be more specified?

I’m more leaned toward the latter, I think.  Ideally, these values wouldn’t be recursively converted at all, so they should be actually atomic with regards to syntax wrap, but I’m not sure how that could be achieved.

mflatt:
I think I see what you're getting at, now. The original implementation of `Syntax.inject` didn't have sensible behavior on a map (that uses a built-in key comparison like  `==`) or array, because it would convert values to syntax, and then `Syntax.unwrap` would not recover the same map, array, etc.

I've adjusted the implementation so that `Syntax.inject` internally wraps values like maps, arrays, and pairs in an `injected` struct (prefab, but mutable), and so it now works as the documentation would suggest on those kinds of values. Also, there doesn't have to be a special prohibition against pairs or the empty pair list. Originaly, I was worried about finding all the places that syntax objects are implicitly unwrapped, but I think it turns out to be only `#%literal`, so adding an `injected` wrapper was straightforward.

There's still room for specifying what happens when you try to inject a map into an expression form, but I think we could save specifying that for later, assuming the change resolves your key concern.

We could further change `Syntax.inject` to treat any kind of value the same, instead of treating values that `Syntax.wrap` accepts the same as `Syntax.wrap` does — which includes, for example, converting mutable strings to immutable strings. I'm currently inclined to stick with behavior that make `Sytax.inject` mean "`Synatx.wrap`, but more permissive".

PR #650, opened 2025-03-30
add `alias`

mflatt:
Although `namespace ~open: export: rename: is as new_id` can be used to alias `new_id` to `id`, it's a lot to type for a relatively simple operation. Also, opening a namespace does not work for extending an existing namespace.

The `alias` form is similar to `macro`, but it binds in all spaces where the aliased name is bound, not just the expression space, at least by default. The `alias` form supports optional `~only_space` and `~except_space` clauses to constrain the set of bindings. Like `macro`, `alias` is provided by `rhombus` and not `rhombus/meta`.

Examples from the documentation:

```
> alias 'Roster': 'List'
> Roster[1, 2, 3]
[1, 2, 3]
> [1, 2, 3] is_a Roster
#true

> alias '%': 'mod'
> 5 % 2
1
```


rhombus/rhombus/scribblings/reference/alias.scrbl
```diff
@@ -0,0 +1,52 @@
...
+      $space ...
+      ...
+    ~except_space $space ...
+    ~except_space:
+      $space ...
+      ...
+){
+
+ Defines @rhombus(new_op_or_id_name) to be an alias for
+ @rhombus(new_op_or_id_name). By default, the new name is bound in all
```
shhyou, line 31: > Define `@rhombus(new_op_or_id_name)` to be an alias for `@rhombus(new_op_or_id_name)`
Two `@rhombus(new_op_or_id_name)`?

mflatt: Fixed - thanks!

distractedlambda:
I don't think `alias` is quite working correctly with namespaces; this code doesn't compile:
```rhombus
#lang rhombus

namespace Foo
namespace Bar

def Bar.baz = 42

alias 'Foo.Bar': 'Bar'

println(Foo.Bar.baz)
```
It complains that `baz` is not provided by `Foo.Bar`; the aliasing of `Foo.Bar` to `Bar` doesn't seem to also carry over `Bar`'s namespace extensions.

mflatt:
The problem with namespace extensions makes sense. Racket doesn't support namespaces directly, so it's a battle to be refought at every level. I'll work on that.

PR #657, opened 2025-04-04
optional binding of an escape identifier by `try`

mflatt:
This PR adds `let/ec`-like support to `try`, where an identifier written after `try` and before the body block is bound as an escape continuation. In other words, `try escape: <body>` is like `(let/ec escape <body>)` in Racket. An escape identifier can be provided while also using forms like `~finally` or `~catch` in the `try` body.

The idea behind overloading `try` this way is to continue bundling control-flow capabilities into `try`, instead of having different control forms like `let/ec`, `with-handlers`, and `dynamic-wind`.

### DIFF
--- a/rhombus-lib/rhombus/private/amalgam/control.rkt
+++ b/rhombus-lib/rhombus/private/amalgam/control.rkt
@@ -70,85 +70,105 @@
 (define/arity (Continuation.Marks.current)
   (current-continuation-marks))
 
+(define-for-syntax (parse-try stx now-stx)
+  (syntax-parse now-stx
+    [(head #:escape_as escape:identifier (~and b (tag::block . _)))
+     (values #`(let/ec esc
+                 (define escape esc)
+                 (define-static-info-syntax escape . #,(get-continuation-static-infos))
+                 (finish-esc-try #,stx (head b)))
+             #'())]
+    [(_ (tag::block g ...))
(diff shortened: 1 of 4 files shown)


sorawee:
`try` makes sense with `catch`/`finally`. It's about how there's failure possibility, so we "try" it. What is being "tried" in this code?

```
    try escape:
      1 + escape(0) + 2
      println("doesn't get here")
```

I think generalizing the construct makes sense, but perhaps names like `do` will make more sense?

eutro:
Putting the name right after reads very poorly I think. `try esc:` reads as if it were `esc` which is being "tried", which makes very little sense. Something like `try ~as: esc; ...` at the very least reads nicer. `let/ec` doesn't have this problem because `let` already suggests that something is being bound, which isn't the case with `try`.

shhyou:
Or maybe `try ~break escape:`?

mflatt:
Latest variant uses `~escape_as`.

The keyword `~as` seemed a little mismatched to the use of `as` in imports and exports. The keyword `~break` seemed mismatched to the use of `break` as a clause that breaks out of a block, as opposed to binding a break-like function. I could see `~break_as`; then again, escape continuations are also more general that breaks, since uses don't have to be syntactically apparent in the body of the relevant block. So that's my rationale for `~escape_as`, but I'm not set on it.

PR #659, opened 2025-04-06
rename `Map.add` to `Map.set`

usaoc:
Commit a01383c added `Map.add`, but I think the name was a mistake.  Rename it to correspond to `MutableMap.set`, and document it.

Also fix some bugs related to `Map.set`, `Set.add`, and the `++` special case.

### DIFF
--- a/rhombus-lib/rhombus/private/amalgam/map.rkt
+++ b/rhombus-lib/rhombus/private/amalgam/map.rkt
@@ -196,8 +196,9 @@
   ([maybe Map.maybe extract-maybe-statinfo])
   #:methods
   (append
-   remove
-   add))
+   set
+   [add Map.set] ; TEMP to be removed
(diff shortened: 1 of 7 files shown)


mflatt:
Changing to `Map.set` makes sense to me. I think the choice of `add` was probably a leftover from trying to keep mutation operations named differently than functional-update operations, which was an approach that I eventually abandoned. Thanks for these improvements.

For a while, I think we should have `Map.add` as an alias for `Map.set` to smooth the transition. I don't know the best way to track this so that we don't leave transition bindings in place indefinitely, but we should create some mechanism. Maybe a rename transformer that prints a "deprecated warning" whenever it's used before returning an alias, and with an expiration date built into the syntax?

usaoc:
Updated to include a backward compatibility alias.

PR #664, opened 2025-04-20
add argument-dependent function/method result annotations

mflatt:
Argument dependency in a function-result annotation enables the propagation of static information. It's reminiscent of parametric polymorphism. Along with this change, constructors infer more static information about constructed values from subexpressions.

For example, the expression

```
 ["x", "y", "z"]
```

now has the static information `List.of(String)`, instead of just `List`, and

```
 use_static
 ["x", "y", "z"].sort((_ :: String < _))[0].length()
```

now works, because the `sort` method indicates that its result will have the same static information as the elements of its argument.

Similarly, `List.first(lst)` now propagates static information the same way as `lst.first`, where the latter had been handled specially by the dot provider.

New annotation forms such as `Any.like` and `Any.like_element` implement dependency. For example, the `List.map` method is defined/documented as

```
  method (lst :: List).sort(is_less :: Function.of_arity(2) = (_ < _))
   :: List.of(Any.like_element(lst))
```

and `List.add` is

```
  method (lst :: List).add(elem :: Any)
    :: List.of(Any.like_element(lst) || Any.like(elem))
```

So, `["a", "b"].add("c")` is a `List.of(String)`, but `["a", "b"].add(#"c")` is just a `List`, since `String || Bytes` is the same as `Any`.

The `Any.like` annotation constructors are for static-information only; they have no associated predicate. That's unlike a polymorphic contract in Racket.

The `Any.like_field` annotation constructor generalizes from built-in datatypes to classes. Even more general combinations can be implemented by dropping down to the lower-level protocol.

To enable dependent result annotations, the annotation protocol is extend to have a `annot_meta.Context` objects that is optionally passed to annotation macros via a `~context` declaration. The `annot_meta.Parsed` syntax class accepts an optional context argument. (Also, syntax classes that have only optional arguments can now be used without parentheses, which provides a transition path, the same as in `syntax-parse`.) The implementation of of a dependent annotation typically uses `annot_meta.Context` to find arguments, and it generates static information that includes
`statinfo_meta.dependent_result_key` within
`statinfo_meta.call_result_key`; that is used, in turn, by the function-call form to provide argument information to a compile-time function specified in `statinfo_meta.dependent_result_key`.


mflatt:
This was originally motivated by adding `Port.Input.lines`, where I wanted to write `Port.Input.lines().to_list()` and end up with a `List.of(String)` instead of just a `List`. The generalization was predictably complicated, but I think it's going to be a worthwhile and far-reaching improvement to static-information propagation.

I'm sure there are still many places in where existing function result annotations can be refined to take advantage of this, since I focused on the container datatypes. I haven't yet checked how much improved static-information propagation lets me remove annotations in my existing code.

distractedlambda:
Will this also work for object protocols like `Indexable`? I was recently trying to capture a particular "heterogeneous map" pattern I've seen in statically-typed languages, where the type of a key tells you the type of value it can be associated with.

mflatt:
I expected this to work, but I see that it doesn't, yet:

```
#lang rhombus/static

class Identity():
  implements Indexable
  override get(x) :: Any.like(x):
    x

let i = Identity()
i["hello"].length()
```

Something goes wrong in overloading and/or the redirection of `[]` via `Indexable`. I will look at it more to find out what's wrong, but this kind of thing is intended to work.

mflatt:
The problem affecting the `Indexable` example above is now fixed.

mflatt:
Also, thanks to attendees at the April 17 meeting — @jackfirth and @rfindler — for help refining the design here.

PR #677, opened 2025-05-22
Update editor.scrbl

ShalokShalom:
Updates of Magic Racket made highlighting in VSCode better, it can now considered to be rudimentary and this commit reflects that

### DIFF
--- a/rhombus/rhombus/scribblings/guide/editor.scrbl
+++ b/rhombus/rhombus/scribblings/guide/editor.scrbl
@@ -28,5 +28,4 @@ specified by the initial @hash_lang() line.
 The
 @hyperlink("https://marketplace.visualstudio.com/items?itemName=evzen-wybitul.magic-racket"){Magic
  Racket} extension for Visual Studio Code provide support for running
-Rhombus programs, but its editor support is currently tied to Racket's
-parenthesis-oriented syntax.
+Rhombus programs, and adds rudimentary support for syntax highlighting as well.


rhombus/rhombus/scribblings/guide/editor.scrbl
```diff
@@ -28,5 +28,4 @@ specified by the initial @hash_lang() line.
 The
 @hyperlink("https://marketplace.visualstudio.com/items?itemName=evzen-wybitul.magic-racket"){Magic
  Racket} extension for Visual Studio Code provide support for running
-Rhombus programs, but its editor support is currently tied to Racket's
-parenthesis-oriented syntax.
+Rhombus programs, and adds rudimentary support for syntax highlighting as well.
```
jryans, line 31: This document doesn't really go into detail comparing editor features, so I don't really see a need to include modifiers like "rudimentary" here. I would suggest removing it. The highlighting can easily be improved through contributions, and we'd likely forget to go back and update this document. (I also don't really know what makes it "rudimentary", so please file issues on Magic Racket!).

Let's also go for something a bit more feature-inclusive similar to the [Racket](https://docs.racket-lang.org/guide/Visual_Studio_Code.html) version of this documentation. (They both could mention even more features like code navigation, but perhaps that's best left to the extension for future maintenance.) One way might be to say "includes many features such as...".

jryans: I can see that the highlight is a bit more limited that the Racket variant for the moment.

Still though, I'd personally rather sort that out in the issues of the extension, rather than having to remember to come back here and remove the word "rudimentary" in the future.

ShalokShalom: But that would be lying. 

Cause now, it does not support syntax highlighting, as that would suggest feature parity with Racket. 

Also Racket is described as "providing syntax highlighting."

jryans: I'll try one more time...

This editor documentation does not currently try to describe the quality or completeness of features, and I do not think it should start doing so now. In my opinion, it's best for this documentation to just point out the existence of editor integrations without applying status levels or judgements.

Those finer details can then be found in the integration's own repo and documentation, which is also more likely to be kept up to date by its maintainers.

Anyway, I'll stop here for now. We should probably wait for someone else to give their take on this.

LiberalArtist: From experience with the Racket Guide section on editors, I strongly agree that this document should not get into detail about the current features of extensions. It is very easy for these kind of documents to get it if date. More fundamentally, this documentation conceptually applies to a given released version of Rhombus (or snapshot), forever, but editor extensions have their own release cycles: a given version of Rhombus is not pinned to a specific version of any particular extension, nor visa versa.

The statement “Magic Racket supports syntax highlighting” appears to be literally true. Whether it does more, less, or a different kind of highlighting than a particular user would like is a different question, and IMO out of scope for this document. People have vastly different tastes about how they want their syntax to be highlighted.

ShalokShalom: I know this kind of coding standard is not welcome usually, but if we would adopt a document-first approach, then no documentation could ever go out of date. 

Before any code would get merged, it would require the update of the documents associated to it first. 

Personally, I find it misleading at best, to claim something is ready, when its really in beta, or something similar. 

That will just anger the people running into it, a quite human reaction. 
Sad, that we have such low standards here. 

> The statement “Magic Racket supports syntax highlighting” appears to be literally true. 

That means, we could just advertise syntax highlighting the moment strings are highlighted. That nobody is concerned with that scammy behavior, is sad.

LiberalArtist: > if we would adopt a document-first approach, then no documentation could ever go out of date.
> 
> Before any code would get merged, it would require the update of the documents associated to it first.

For many things in Racket, we do require documentation updates together with code changes. The challenge with this particular kind of document is that there isn't any code directly associated with it.

I'd also suggest that it would be more constructive to avoid words like “lying” or “scammy” to describe a community-run volunteer project that many people find works well for their needs. I am sure we all want to this documentation to be accurate. Personally, I notice only minor differences between the screenshot above and the syntax highlighting used on <https://rhombus-lang.org>, though I understand some people prefer much more colorful syntax highlighting than I'm used to.

Rhombus is described “ready for early adopters” and “complete and stable enough to be useful, but not all done.” If there are ways that Magic Racket's Rhombus support falls short of that standard, they are not obvious to me.

ShalokShalom: > The challenge with this particular kind of document is that there isn't any code directly associated with it.

That's a lie again. The code associated with this document is the syntax highlighting code related to Magic Racket.

I consider this behavior as devious and dishonest and don't want to have anything to do with it.

If you wanna change the current inaccurate statement (does not work at all) to be replaced with an overly enthusiastic statement that the app can't fulfill, then so be it.

I prefer that I don't have my name associated with that PR.

Thanks

PR #684, opened 2025-05-31
Adding Synchronizable interface and Evt.wrap

samdphillips:


rhombus/rhombus/scribblings/reference/evt.scrbl
```diff
@@ -73,3 +90,21 @@ count.
...
+ @rhombus(Synchronizable) is used with @rhombus(Evt.sync) the
+ @rhombus(Synchronizable.to_evt) method is called, and the result is used in
+ the synchronization.
+
+ The interface has a single abstract method:
+
+@itemlist(
+  @item{@rhombus(#,(@rhombus(to_evt, ~datum))()) --- produces an
+  @tech{synchronizable event} that can be used in @rhombus(Evt.sync).}
+)
```
LiberalArtist, line 108: Isn't the method `as_evt`?

I also wonder if the fact that the method is called inside a `guard-evt` has any implications that should be documented. (It does definitely seem to me like the right choice!)

samdphillips: Good catch on the method name.

I think that saying it gets called during an `Evt.sync` should cover it, but improvements welcome.

mflatt:
Thanks for this addition!

In general, I have tried avoiding boolean mode arguments, because symbol arguments generalize better and are usually easier to read. So, in the case of `evt.wrap`, maybe the `~tail_call` argument could instead be `~return` as either `#'tail` or `#'no_break` (in a `evt.wrap.Return` enum)?

mflatt:
Thanks!

PR #687, opened 2025-06-12
Repair http auth and payload wraps and unwraps

samdphillips:
This repairs using params with authentication functions and also using http.payload.form.


mflatt:
Thanks for this repair! Is there something we could add to `rhombus/rhombus-http/rhombus/net/tests/http.rhm` as a regression test, or would it be too complicated to make the mock server there comsume/produce something relevant?

PR #689, opened 2025-06-13
better propagate static infos in `::`/`:~`

usaoc:
This addresses an issue raised on Discord that static infos aren’t property `&&`ed in `::`/`:~` expression.  Now, static infos from the LHS expression are extracted and propagated to RHS annotation.  Specfically, when RHS annotation is predicate annotation, the resulting static infos is the `&&`, while when it’s converter annotation, the static infos is propagated to the inner binding (such as in the argument of `converter`).


mflatt:
Looks good. Thanks for working on this!!

usaoc:
Merged at 666be16.

PR #690, opened 2025-06-22
support multiple-value patterns in `match`

usaoc:
The `match` form is extended to support multiple-value patterns, similar to Racket `match/values`.  ~~The syntactic constraints are like in `match/values`: all clauses must match the same number of values, and at least one pattern clause must be provided.~~  Like `match/values`, the target expression is expected to produce a fixed, known number of values.  Unlike `match/values`, though, we support `~else` clause in multiple-value `match`~~, given that at least one pattern clause is provided~~.  Multiple-value patterns take the forms `(<bind>, ...)` or `values(<bind>, ...)`, as in `def`/`let`.

~~Compatibility note: most existing `match` forms will continue to work.  Some `match` forms with *only* `~else` clause or no clause at all (`match <expr>:«»`) will stop working, but I think such forms only really occur in macro expansion (`match <expr> | $clause | ...` with an empty `clause` sequence).  The other option is to still allow such forms, but this would make it that the target expression could produce any number of values when there isn't any pattern clause, and that seems less desirable.~~

The expected arity is inferred in a `match` form:

- when there is at least one pattern clause, infer it from the first pattern clause;
- when no clause or only `~else` clause is provided, take it to be 1.

It can also be explicitly given with an `~arity` declaration.  All clauses must match the same number of values and, if present, be consistent with the `~arity` declaration, otherwise a syntax error is reported.

The `guard.let` form now also supports multiple-value patterns, because it expands into `match`.

### DIFF
--- a/rhombus-lib/rhombus/private/amalgam/charset.rhm
+++ b/rhombus-lib/rhombus/private/amalgam/charset.rhm
@@ -90,11 +90,11 @@ class Charset(private _ranges):
     _Charset(
       recur loop(a = _ranges,
                  b = other._ranges):
-        match [a, b]
-        | [[], b]: b
-        | [a, []]: a
-        | [[[as, ae], ar, ...],
-           [[bs, be], br, ...]]:
+        match values(a, b)
+        | ([], b): b
+        | (a, []): a
+        | ([[as, ae], ar, ...],
+           [[bs, be], br, ...]):
             cond
             | bs .< as:
                 loop(b, a)
@@ -115,11 +115,11 @@ class Charset(private _ranges):
     _Charset(
       recur loop(a = _ranges,
                  b = other._ranges):
-        match [a, b]
-        | [[], b]: []
-        | [a, []]: []
-        | [[[as, ae], ar, ...],
-           [[bs, be], br, ...]]:
+        match values(a, b)
+        | ([], b): []
+        | (a, []): []
+        | ([[as, ae], ar, ...],
+           [[bs, be], br, ...]):
             cond
             | bs .< as:
                 loop(b, a)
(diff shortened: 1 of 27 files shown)


jackfirth:
My general view on multiple values can summarized in a few points:

- Multiple values should be avoided when possible. Simple aggregating structs/classes are easy to write and far easier to read, since they give the different values meaningful names.
- If you can't avoid them, it's probably for performance reasons. In that case, you probably want a _statically-known_ number of values, since otherwise you won't be gaining much performance from them.
- If you want multiple values _and_ a dynamic quantity of values, you _almost certainly_ would be better off using a normal list.

This means that I don't think _any_ forms should accept a dynamic number of values, with the sole exception of `call_with_values`. You should almost always know how many values an expression can evaluate to, and if it evaluates to a different number of values that's probably a bug and should result in a quick and loud failure.

So I don't think `match e | ~else:` should accept any number of values just because the number of values doesn't matter. We don't let you do that with `let` or `def`, and I don't think we should let you do it with `match` either.

mflatt:
FWIW, I remain in favor of having the expected arity determined by pattern clauses and defaulting to a single value when no pattern clauses are present. Using a parenthesized sequence of patterns seems like a more succinct and discoverable way of supporting multiple values than having a `match` variant. More generally, I'm ok with some irregularity (i.e., single-value mode as the default) when a regular variant is also available (such as through an `~arity` declaration).

usaoc:
Let’s try the alternative design, then.  In the updated commit, `match` recognizes an explicit `~arity` declaration, and default to single-value mode otherwise.  Multiple-value mode is still inferred automatically from the patterns.

mflatt:
LGTM!

PR #707, opened 2025-09-05
add `statinfo_meta.none_key` and specialize `||` and `statinfo_meta.or`

mflatt:
Change the `None` annotation to imply `statinfo_meta.none_key`, which is specially recognized by the `||` annotation so that `None || ann` is the same as `ann`.

The `&&` turns out not to treat `None` specially, and so something like `(error("oops") :~ List).length()` is allowed in static mode. In principle, `(x :~ None).length()` also should be allowed, but `None` here doesn't attempt to provide all static information for all annotations; it just provides the static information with key `statinfo_meta.none_key`.

### DIFF
--- a/rhombus-lib/rhombus/private/amalgam/annotation.rkt
+++ b/rhombus-lib/rhombus/private/amalgam/annotation.rkt
@@ -1026,7 +1026,7 @@
 (void (set-primitive-contract! 'flonum? "Flonum"))
 (void (set-primitive-contract! 'fixnum? "Fixnum"))
 (define-annotation-syntax Any (identifier-annotation (lambda (x) #t) ()))
-(define-annotation-syntax None (identifier-annotation (lambda (x) #f) ()))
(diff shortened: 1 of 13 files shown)


rhombus/rhombus/scribblings/reference/continuation.scrbl
```diff
@@ -17,6 +17,10 @@ or by using it with @rhombus(Continuation.in).
  @rhombus(Continuation.capture). Continuations are callable as
  functions and satisfy @rhombus(Function, ~annot).
 
+ The static information of @rhombus(Continuation, ~annot) indicates that
+ its result when called as a function has static information
+ @rhombus(None, ~annot), since the call never returns.
```
usaoc, line 22: I think this is only true for *non*-composable continuations?

mflatt: Oops - right!

PR #720, opened 2025-11-09
use `ssl.Client.Context` for `http.Session`

### LINKED ISSUE #712: Fresh default http.Session throws an error
Using a newly constructed `http.Session` always throws an error when used.  I've observed the same behavior on Linux and MacOS.

```
// using the default Session
> http.Session.current().request("https://www.google.com").status_line
Bytes.copy(#"HTTP/1.1 200 OK")

// using a new Session with default settings
> http.Session().request("https://www.google.com").status_line
racket/racket/collects/net/http-client.rkt:37:0: error reading from stream port
  system error: Connection reset by peer; errno=54
```

mflatt:
this repair would close #712, but it's a backward-incompatible change for anything that passes a Racket-level `ssl-client-context?` as `~ssl_context`. Ok to merge, or is more work needed to create a compatibility path?

### DIFF
--- a/rhombus-http-lib/info.rkt
+++ b/rhombus-http-lib/info.rkt
@@ -7,6 +7,7 @@
     "rhombus-lib"
     ["rhombus-json-lib" #:version "0.2"]
     "rhombus-url-lib"
+    "rhombus-ssl-lib"
     "http-easy-lib"))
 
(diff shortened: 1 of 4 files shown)


samdphillips (commented):
For my uses this is an improvement and compatibility is not needed.

PR #722, opened 2025-11-15
add `rhombus/date`

mflatt:
This attempt to add time and date functions in a `rhombus/date` library is based on the `gregor` library for Racket and Java's `date.LocalTime`, etc., classes. It doesn't try to do everything that those do, but instead provides basic functionality that is available from `racket/base` and `racket/date`, but in a form that hopefully does not preclude improvements and compatible generalization by new libraries (unlike the way Gregor has to avoid the Racket `date` structure type).

I think I get why Java puts `Local` in the names of classes, but I have opted not to do that, but the name `ZonedDateTime` does come from Java.

API sumary:

```
export:
  Time
  Date
  DateTime
  ZonedDateTime
  Format  

class Time(~hour: hour :: Int.in(0, 23) = 0,
           ~minute: minute :: Int.in(0, 59) = 0,
           ~second: second :: Int.in(0, 60) = 0,
           ~nanosecond: nanosecond :: Int.in(0, 999_999_999) = 0):
  private implements Comparable

  fun now(~local = #true) :: Time

  fun from_seconds(secs :: Real,
                   ~local = #true) :: Time             

  method to_string(~show_nanosecond = #false) :: String

class Date(~year: year :: Int,
           ~month: month :: Int.in(1, 12 ~inclusive),
           ~day: day :: Int.in(1, 31 ~inclusive)):
  private implements Comparable

  fun now(~local = #true) :: Date

  fun from_seconds(secs :: Real,
                   ~local = #true) :: Date

  method to_datetime(~time: time :: date.Time = Time()) :: DateTime
  
  method to_seconds() :: Real

  method to_string(~format: format :: date.Format = default_format) :: String

class DateTime(~year: year :: Int,
               ~month: month :: Int.in(1, 12),
               ~day: day :: Int.in(1, 31),
               ~hour: hour :: Int.in(0, 23) = 0,
               ~minute: minute :: Int.in(0, 59) = 0,
               ~second: second :: Int.in(0, 60) = 0,
               ~nanosecond: nanosecond :: Int.in(0, 999_999_999) = 0):
  private implements Comparable

  fun now(~local = #true) :: DateTime

  fun from_seconds(secs :: Real,
                   ~local = #true) :: DateTime

  method to_seconds() :: Real

  method to_zoned() :: ZonedDateTime

  method to_string(~format: format :: date.Format = default_format,
                   ~show_time = #true,
                   ~show_nanosecond = #false) :: String

class ZonedDateTime(~year: year :: Int,
                    ~month: month :: Int.in(1, 12),
                    ~day: day :: Int.in(1, 31),
                    ~hour: hour :: Int.in(0, 23) = 0,
                    ~minute: minute :: Int.in(0, 59) = 0,
                    ~second: second :: Int.in(0, 60) = 0,
                    ~nanosecond: nanosecond :: Int.in(0, 999_999_999) = 0,
                    ~week_day: week_day :: Int.in(0, 6) = 0,
                    ~year_day: year_day :: Int.in(0, 365) = 0,
                    ~is_dst: is_dst :: Boolean = #false,
                    ~time_zone_offset: time_zone_offset :: Int = 0,
                    ~time_zone_name: time_zone_name :: String = "UTC"):
  private implements Comparable

  fun now() :: ZonedDateTime

  fun from_seconds(secs :: Real,
                   ~local: local = #true)

  fun find_seconds(~second: second :: Int.in(0, 60) = 0,
                   ~minute: minute :: Int.in(0, 59) = 0,
                   ~hour: hour :: Int.in(0, 23) = 0,
                   ~day: day :: Int.in(1, 31),
                   ~month: month :: Int.in(1, 12),
                   ~year: year :: Int,
                   ~local: local = #true) :: Int

  method to_string(~format: format :: date.Format = default_format,
                   ~show_time = #true,
                   ~show_nanosecond = #false,
                   ~show_time_zone = #true) :: String

  method to_seconds() :: Real

enum Format:
  rfc2822
  rfc3339
  iso8601
  american
  european
  julian

```

### DIFF
--- a/rhombus-lib/rhombus/date.rhm
+++ b/rhombus-lib/rhombus/date.rhm
@@ -1,0 +1,426 @@
+#lang rhombus/static
+import:
+  lib("racket/base.rkt") as rkt
+  lib("racket/date.rkt") as rkt_date
+  "private/date.rhm" as util
+
+export:
+  Time
+  Date
+  DateTime
+  ZonedDateTime
+  Format
+  TimeFormat
+  TimeZoneFormat
+  util.is_leap_year
+  util.days_in_year
+  util.days_in_month
+  util.year_day
+  util.week_day
+
+// for referencing in annotatons with a `date.` prefix:
+namespace date:
+  export:
+    Time
+    Date
+    DateTime
+    ZonedDateTime
+    Format
+    TimeFormat
+    TimeZoneFormat
+    util.days_in_month
+    util.year_day
+    util.week_day
+
+fun rhm_to_string(s): to_string(s)
+alias 'rhombus_compare_to': 'compare_to'
+
+fun str_2d(n) :~ String:
+  str.d(n, ~width: 2, ~pad: Char"0", ~align: #'right)
+
+fun date_string(year, month, day, ~iso = #false) :~ String:
+  (if iso
+   | str.d(year, ~min_width: 4, ~pad: Char"0", ~align: #'right, ~sign_align: #'left)
+   | str.d(year))
+    ++ "-" ++ str_2d(month) ++ "-" ++ str_2d(day)
+
+fun time_string(hour, minute, second, nanosecond, format):
+  str_2d(hour)
+    ++ ":" ++ str_2d(minute)
+    ++ (if (format == #'minutes
+              || (format == #'auto_subminutes && second == 0 && nanosecond == 0))
+        | ""
+        | ":" ++ str_2d(second))
+    ++ (if (format == #'seconds
+              || format == #'minutes
+              || ((format == #'auto_subseconds || format == #'auto_subminutes)
+                    && nanosecond == 0))
+        | ""
+        | let s = str.d(nanosecond, ~width: 9, ~pad: Char"0", ~align: #'right)
+          "." ++ (match format
+                  | #'milliseconds: s.substring(0, 3)
+                  | #'microseconds: s.substring(0, 6)
+                  | #'nanoseconds: s
+                  | ~else:
+                      recur loop (i = s.length()):
+                        if s[i-1] == Char"0" && s[i-2] == Char"0" && s[i-3] == Char"0"
+                        | loop(i - 3)
+                        | s.substring(0, i)))
+
+fun time_zone_string(time_zone_offset, sep :~ String):
+  let offset = math.abs(time_zone_offset)
+  let h = offset div 3600
+  let m = (offset mod 3600) div 60
+  (if time_zone_offset < 0 | "-" | "+")
+    ++ str_2d(h)
+    ++ sep ++ str_2d(m)
+
+enum Format:
+  rfc2822
+  rfc3339
+  iso8601
+  american
+  european
+  julian
+
+enum TimeFormat:
+  minutes
+  seconds
+  milliseconds
+  microseconds
+  nanoseconds
+  auto_subminutes
+  auto_subseconds
+
+enum TimeZoneFormat:
+  offset
+
+def default_format = #'rfc3339
+
+class ZonedDateTime(~year: year :: Int,
+                    ~month: month :: Int.in(1 ..= 12) = 1,
+                    ~day: day :: Int.in(1 ..= date.days_in_month(year, month)) = 1,
+                    ~hour: hour :: Int.in(0 ..= 23) = 0,
+                    ~minute: minute :: Int.in(0 ..= 59) = 0,
+                    ~second: second :: Int.in(0 ..= 60) = 0,
+                    ~nanosecond: nanosecond :: Int.in(0 ..= 999_999_999) = 0,
+                    ~time_zone_offset: time_zone_offset :: Int = 0,
+                    ~time_zone_name: time_zone_name :: String = "UTC",
+                    ~is_dst: is_dst :: Boolean = #false,
+                    private _week_day = date.week_day(year, month, day),
+                    private _year_day = date.year_day(year, month, day),
+                    private mutable _seconds = #false,
+                    private _handle = rkt.#{date*}(second, minute, hour, day, month, year,
+                                                   _week_day, _year_day,
+                                                   is_dst, time_zone_offset,
+                                                   nanosecond, time_zone_name)):
+  internal _ZonedDateTime
+
+  serializable:
+    ~serialize: fun (): Array(_handle)
+    ~deserialize: fun (handle): from_handle(handle)
+
+  property handle: _handle
+
+  private implements Printable
+  private override method describe(mode, recur):
+    if mode == #'text
+    | to_string()
+    | recur(this, ~as: #'super)
+
+  private implements Comparable
+  private override method compare_to(other :~ ZonedDateTime):
+    to_seconds() rhombus_compare_to other.to_seconds()
+
+  private implements Equatable
+  private override equals(other :~ _ZonedDateTime, recur):
+    recur(_handle, other._handle)
+  private override hash_code(recur):
+    recur(_handle)
+
+  property week_day :~ Int: _week_day
+  property year_day :~ Int: _year_day
+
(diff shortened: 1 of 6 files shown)


samth:
Can you say more about why you decided not to use gregor directly? Also, gregor provides a bunch of functionality that requires various tables (eg the tz database). Do you think that should eventually be a part of this library, some other standard library, or something else?

LiberalArtist:
Along similar lines, re:

> in a form that hopefully does not preclude improvements and compatible generalization by new libraries (unlike the way Gregor has to avoid the Racket `date` structure type).

Can you say more about which problems with Racket's `date` this tries to avoid?

I noticed that there are separate fields for `is_dst`, `time_zone_offset`, `time_zone_name`, making it possible to represent nonsensical states, which I think of as one of the problems with Racket's `date`.

I looked back at https://github.com/racket/racket/issues/1220 for some context. In https://github.com/racket/racket/issues/1220#issuecomment-173806277, @97jaz wrote:

> Gregor's date struct really just represents a date. Racket's represents a date, a time, a UTC offset, and a few other things like the day-of-week, which Gregor treats as a function of the data rather than part of it. Racket's date is similar to what Gregor calls a "moment," but the two still aren't quite compatible. Racket will let me construct, for example, a date representing Sunday, January 21, 2016 (at some time or other, etc.), but Gregor keeps to the Gregorian calendar in which January 21, 2016 is a Thursday. Also, Gregor is picky about what will count as a time zone.

Re @samth:

> Also, gregor provides a bunch of functionality that requires various tables (eg the tz database).

In discussion around https://github.com/racket/racket/issues/1220#issuecomment-181915686, the size of data files was mentioned as an obstacle to including Gregor in the main Racket distribution, but I hope that can be overcome.

I need to refresh my memory about what needs the `cldr-core` data as opposed to just the tz database. For the tz database specifically, though, it is provided by the operating system except on Windows, where it's sourced from a platform-specific dependency on the [`tzdata`](https://pkgs.racket-lang.org/package/tzdata) Racket package. Today I discovered [PEP 615](https://peps.python.org/pep-0615/), which adopted the same approach for the tz database in Python (but I don't see discussion of `cldr-core` data there).

Another development since previous discussion is that currently-supported Windows versions provide ICU's C API as a system library (see also https://github.com/racket/racket/pull/4844), and ICU integrates the tz database and CLDR data. Potentially, if we could limit ourselves to functionality exposed by the ICU C API, we could avoid distributing data files for Windows altogether. PEP 615 has this to say about that rejected alternative:

> Providing bindings for this would allow us to support Windows “out of the box” without the need to install the `tzdata` package, but unfortunately the C headers provided by Windows do not provide any access to the underlying time zone data — only an API to query the system for transition and offset information is available. This would constrain the semantics of any ICU-based implementation in ways that may not be compatible with a non-ICU-based implementation — particularly around the behavior of the cache.

mflatt:
> Can you say more about which problems with Racket's `date` this tries to avoid?

The main one I know is using the word "date" for "date and time". I believe the `Time`, `Date`, and `DateTime` classes here line up with same-named structure types in `gregor`. 

As you note, there's no attempt here to ensure that a `ZonedDateTime` has a consistent set of fields (because that's not easy), and that's partly why the class is not called `Moment`. I view `ZonedDateTime` primary as a way to report back the OS facility for converting from sections. Maybe there could be some bridge between `Moment` and `ZonedDateTime` in an eventual Rhombus variant of `gregor`, but the intent is to have a clearer point of departure where they do not have to be the same.

LiberalArtist:
> On the other hand, I still wouldn't have allowed the construction of nonsense dates and times

I had forgotten that Racket is happy to represent Sunday, February 31, the 365th day of 2025, as `(date 0 0 0 31 2 2025 0 365 #f 0)`.

> That's mostly due to the CLDR i18n/l10n data and code, which in hindsight I would have omitted from the library

If it's reasonable to split complicated formatting and parsing off from the issue of correct time zones, so that we don't need `cldr-core`, it seems even more reasonable implement the timezone part, either by depending on [the `tzdata` package](https://pkgs.racket-lang.org/package/tzdata) on Windows or using ICU. In fact, I see that [the `gen:tzinfo-source` interface](https://docs.racket-lang.org/tzinfo/index.html#%28part._.Data_.Source_.Generics%29) already contemplates alternative backends like ICU, requiring just the methods:
```
tzinfo->all-tzids
tzinfo-has-tzid?
tzinfo-tzid->country-codes
tzinfo-country-code->tzids
seconds->tzoffset/utc
seconds->tzoffset/local
detect-system-tzid
```
That seems very feasible to implement on Windows using [the ICU `ucal.h` API](https://unicode-org.github.io/icu-docs/apidoc/released/icu4c/ucal_8h.html).

I do recognize that "seems very feasible" is a long way from just using either `racket/date` or `gregor` as-is. But Racket having an attractive-seeming but fundamentally wrong `date` in the standard library seems worth avoiding.

mflatt:
Changes in the latest version:

 * Leap seconds are disallowed in `date.Time` and `date.DateTime`, although they're still allowed in `date.ZonedDateTime`.

 * A supplied day for `date.Date`, `date.DateTime`, or `date.ZonedDateTime` must be within `date.days_of_month(year, month)`.

 * `week_day` and `year_day` are now properties for `date.Date`, `date.DateTime`, and `date.ZonedDateTime` (derived on demand, except that they're precomputed for `date.ZonedDateTime` as part of creating a Racket `date` struct).

 * All classes are serializable.

 * Functions `date.is_leap_year`, `date.days_in_year`, `date.days_in_month`, `date.year_day`, and `date.week_day` are exported.

LiberalArtist:
I can see how `Time`, `Date`, and `DateTime` are generally useful, and the new version of `ZonedDateTime` seems more closely tailored to the niche of reporting `localtime_r`/`SystemTimeToTzSpecificLocalTime`&rsquo;s opinion of conversion from seconds.


Re @97jaz&rsquo;s comment:

> On the other hand, I still wouldn't have allowed the construction of nonsense dates and times. That's pretty easy to do with two exceptions:
> 
>     * Leap seconds, which gregor completely ignores; it does not allow 60 as a valid second of the hour and so _cannot_ represent all times that your OS might report). Note that java.time also ignores leap seconds.
> 

Relegating leap seconds to the system&rsquo;s-opinion variant seems extremely defensible. Many systems de facto use something like TAI or UT while claiming to use UTC (and a few use other non-leap-second time systems), some systems [smear](https://docs.ntpsec.org/latest/leapsmear.html) away leap seconds in practice, some applications need to [calculate ΔT](https://eclipse.gsfc.nasa.gov/SEcat5/deltat.html) (to some acceptable precision) using historical and physical models, and the [BIPM and ITU-R](https://www.bipm.org/en/-/2023-12-12-wrc-dubai) have made the future of leap seconds less than certain. It seems better to treat leap seconds as a feature of some time systems, not as a feature of time per se.

>     * Local timelines, where hours are skipped or repeated due to daylight saving time adjustments. (Well, not necessarily just _hours_, but you get the idea.)

These discrepancies (especially now that the more egregious `week_day` and `year_day` are gone) seem like a slight variation on the problem solved by [Gregor&rsquo;s offset/gap/overlap resolvers](https://docs.racket-lang.org/gregor/timezone.html#%28part._offset-resolvers%29), which could be used to interoperate with a more sophisticated library.

----

> I view `ZonedDateTime` primary as a way to report back the OS facility for converting from sections.

I still think that the tz database and/or ICU would be a more useful notion of &ldquo;the OS facility&rdquo; than `localtime_r`/`SystemTimeToTzSpecificLocalTime`. But I can see the case for sticking with the status quo, especially when weighing hypothetical code against actual, existing code.

samth:
It might be also worth looking at the design of the `Temporal` library in JavaScript, which is the most recently-designed widely-used date/time library I know of: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Temporal

PR #739, opened 2025-12-12
change `enum` syntax to use `|` alternatives

mflatt:
The `enum` form declares an annotation that is satisfied by a set of alternatives, so it seems like something that should use `|`.

```
enum Fruit
| apple
| banana
| cherry
| ~is_a String
```

The existing syntax uses `:`, which is not terrible, but it stands out less well compared to classes and functions.

```
enum Fruit:
  apple
  banana
  cherry
  ~is_a String
```

I'm not sure was I was thinking before, but maybe I went with `:` because `enum` isn't a branching *expression* form. As Rhombus has evolved, using `|` for alternatives more generally has worked well.

In this PR, the old syntax is still supported, but it triggers a deprecation message.


gus-massa:
How does this handle the single line case? Something like

    enum YN: yes | no

mflatt:
You can write `enum YN | yes | no`.

gus-massa:
I feel like I'm arriving a few years late to the party, and I hope it was not discussed before, but ... 

... How many things would it break to be able to use this instead?

  enum YN: yes | no

PR #747, opened 2026-01-06
Add a form of “descending ranges”

usaoc:
This PR addresses the lack of range-like sequences in *descending* order, as previously discussed on Discord.  In summary:

- `<..` and `<..=` ranges with a starting point are now `{Sequence,List}Range`;
- “Descending ranges” (that are *not* ranges) are added, and `ListRange.descending` produces (listable) descending ranges.

Methods that produce sequences are as follows:

```
(rge :: SequenceRange).to_sequence()             // `rge` by itself is specially inlined
(rge :: SequenceRange).step_by(step :: PosInt)   // specially inlined

(rge :: DescendingRange).to_sequence()           // `rge` by itself is specially inlined
(rge :: DescendingRange).step_by(step :: NegInt) // specially inlined
```

Descending ranges are of the following forms, each corresponding to a sequence range form:

```
start >=.. end  // .from_to
start >=..= end // .from_to_inclusive
start >=..      // .from
start >.. end   // .from_exclusive_to
start >..= end  // .from_exclusive_to_inclusive
start >..       // .from_exclusive
```

Descending ranges satisfy `DescendingRange`, which is *not* a subclass of `Range`.  They only serve as expressions, not as bindings.  Among them, listable ones also satisfy `DescendingListRange`, which offers a `.to_list` implementation.

Moreover, `(rge :: ListRange).descending()` also produces a descending range (more precisely `DescendingListRange`), and optimization in implicit `.to_sequence` and dot `.step_by` cooperate with it.  This means that `(0..5).descending().step_by(-2)` as a sequence is specially inlined.

### DIFF
--- a/rhombus-lib/rhombus/private/amalgam/range-struct.rkt
+++ b/rhombus-lib/rhombus/private/amalgam/range-struct.rkt
@@ -5,10 +5,12 @@
          (struct-out list-range))
 
 (module+ descending
-  (provide (struct-out descending-range)))
+  (provide (struct-out descending-range)
+           (struct-out descending-list-range)))
 
 (struct range () #:authentic)
 (struct sequence-range range () #:authentic)
 (struct list-range sequence-range () #:authentic)
 
 (struct descending-range () #:authentic)
+(struct descending-list-range descending-range () #:authentic)
--- a/rhombus-lib/rhombus/private/amalgam/range.rkt
+++ b/rhombus-lib/rhombus/private/amalgam/range.rkt
@@ -47,7 +47,8 @@
                      Range
                      SequenceRange
                      ListRange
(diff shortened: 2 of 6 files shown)


mflatt:
This looks great!

In your summary above, you annotated `SequenceRange.step_by` as specially inlined, and the improved documentation says the same thing. But aren't `SequenceRange.step_by` and `SequenceRange.to_sequence` both specially inlined (including prior to this commit)? Or am I missing something about they way they interact?

Having all of`SequenceRange.step_by`, `SequenceRange.to_sequence`, and `SequenceRange.ascending` to do the same thing seems like a lot, but I guess the reason for so many aliases is connected to this part that I'm missing about `for` optimization.

PR #782, opened 2026-04-15
add `while`

mflatt:
The `while` form is a shorthand for `for`, which means that `for` clause forms can be used in the body of `while` --- including forms like `each`, but the intent is to allow `break_when` and `skip_when` to play the roles traditionally served by `break` and `continue`.

Related to #771


vinipsmaker:
LGTM

PR #801, opened 2026-07-14
update with the support for typst

capfredf:


mflatt:
Thanks, but I'm not sure it's time to merge this change, yet. It will be a while before Typst support is available in a Rhombus release that's built on a Racket release, and so it seems better to wait before changing the docs.

capfredf:
Got you. I am closing this for the time being.

PR #806, opened 2026-07-24
docs: add missing "or" between the two risks of importing with "open"

das-g:
This way, it's more clear that importing with the "open" modifier has two separate risks:

 * creating import conflicts
 * obscuring the origin of a binding

Without the "or", these two risks seem linked to each other in an unclear (or maybe just nonsensical) way. (At least to me it's unclear what "import conflicts of obscuring *&lt;something>*" would be supposed to mean.)


mflatt:
Thanks!

PR #814, opened 2026-08-08
`explain_anim` add a new `:mode` argument to show extents instead of epochs in timeline

oflatt:


mflatt:
Some initial thoughts:

* The `~mode` keyword is relatively generic for an argument that is specific to labels on the timeline. Maybe `~label_as`, or something like that?
* The interpretation of `s=N` for "seconds" doesn't seem as natural as `t=N` for "time". What what do you think of something like `t=Ns` (i.e., keep `t` and add `s` at the end to stand for `seconds` but is still fairly compact)?

oflatt:
Both sound good to me! Pushed a commit implementing them

PR #816, opened 2026-08-19
pict: add `as_rebuildable`

mflatt:
The `as_rebuildable` form is a shorthand for the most common kind of use of `rebuildable`.


oflatt (approved):
Nice!

PR #819, opened 2026-09-03
Fix 2 instead of @ typo in draw.scrbl

greghendershott:
Looks like shift key not down while typing `@` so got `2`.


usaoc:
Thanks!

PR #820, opened 2026-09-05
edits to the low-level binding-macro protocol docs

rfindler:
Not completely sure that `@deftech`is right for the bullets content but it doesn't seem to cause problems, at least.


usaoc:
Thanks!

PR #821, opened 2026-09-11
rhombus repl should tell the user how to get help

sidkshatriya:
rhombus repl should tell the user how to get help

Before:
```
$ rhombus
Welcome to Rhombus v1.1.0.1
>
```

With this PR:
```
$ rhombus
Welcome to Rhombus v1.1.0.1
Type ,? followed by enter key for help
>
```

(This was not obvious to me. In fact having seen error messages encouraging me to do `,bt` made me realize there might be `,help` and others. This small change will be useful to people new to the language)


usaoc:
I think we should only show this when [XREPL](https://docs.racket-lang.org/xrepl/index.html) is available (which provides the `,` meta commands).  Looking at [`racket/interactive`](https://github.com/racket/racket/blob/master/racket/collects/racket/interactive.rkt), we can tell its availability with `(collection-file-path "main.rkt" "xrepl" #:fail (lambda (msg) #f))`, which is `collect.file_path(~collect: "xrepl", ~file: "main.rkt")` (requiring [`rhombus/collect`](https://docs.racket-lang.org/rhombus-reference/Module_Collection_Paths.html)) in Rhombus.  (The same change may also be useful in the `racket` executable, if you’re willing to submit a patch to [`racket/racket`](https://github.com/racket/racket) too.)

sidkshatriya:
@usaoc Thanks for your helpful comment. Closed in favour of a new PR on racket https://github.com/racket/racket/pull/5584

PR #823, opened 2026-09-14
Minor fixes

rfindler:
I believe these are all just typo fixes and that they are correct.


rhombus-lib/rhombus/private/amalgam/arrow-annotation.rkt
```diff
@@ -279,8 +279,6 @@
      #:with (lhs-i::binding-form ...) #'(l.binding ...)
      #:with (lhs-impl::binding-impl ...) #'((lhs-i.infoer-id () lhs-i.data) ...)
      #:with (lhs::binding-info ...) #'(lhs-impl.info ...)
-     #:with (arg-id ...) (for/list ([name-id (in-list (syntax->list #'(lhs.name-id ...)))])
```
usaoc, line 282: I think you intended to include this change in #824.

rfindler: Oh, right. My mistake.

rfindler:
I've updated both prs, thanks @usaoc .

usaoc:
Thanks!

PR #824, opened 2026-09-14
Delay conversion annotation to propagate name

rfindler:
I didn't work out how the generalized `annot_meta.pack_converter` that works with the delayed names should go, nor am I sure that I really picked the right new name (`annotation-binding-name-form`).

Let me know if either of those seem worth pursuing now. I'm happy to do if that seems important.


rhombus-lib/rhombus/private/amalgam/annotation.rkt
```diff
@@ -682,26 +698,9 @@
...
-             (binding-form
-              #'annotation-predicate-infoer
-              #`(#,(shrubbery-syntax->string #'t.parsed)
-                 #,(and checked? #'c-parsed.predicate)
-                 c-parsed.static-infos
-                 left.infoer-id
-                 left.data))]
-            [c-parsed::annotation-binding-form
-             #:do [(unless checked?
-                     (raise-unchecked-disallowed #'op.name #'t))]
```
usaoc: Minor detail, but I think we should keep the check here.

usaoc:
Since the only purpose of the new packed form is to receive a name from the enclosing context, does it make sense to add a new optional name argument to the `annot_meta.Parsed` (`:annotation`) syntax class instead, so that the enclosing context can pass the name while parsing the annotation?

mflatt:
That was our first idea. The problem is that with a binding form like `<bind> :: <annot>`, the `<annot>` part needs to be parsed right away to produce a parsed version of the `::` form. (The `<annot>` cannot be parsed later, because parsing that determines the end of the `::` form.) That parsed binding form then receives RHS static info, at which point `<bind>`'s infoer can be applied with the same static info, and the relevant name is in the infoer result— but much after `<annot>` had to be parsed.

